PR #94, opened 2016-05-15
Close session without waiting for <Enter> if process terminates with SIGHUP or exit status 129

### LINKED ISSUE #56: Option to disable the [Process completed - press Enter to close] message
Is that possible to have access to an option menu to have several sessions and set the font size, I would be happy to be able to set a smaller font to show more information or have possibility to remove the end message.
Thanks.

rustybird:
See https://github.com/termux/termux-app/issues/56#issuecomment-218978463

I added the exit status 129 check because Marshmallow /system/bin/sh seems to violate  http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html#tag_20_117_09 (by catching the signal and acting as if the last child process had received it).

Fixes #56

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -73,6 +73,7 @@ import java.io.InputStream;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Properties;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -860,19 +861,14 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
     }
 
     public void removeFinishedSession(TerminalSession finishedSession) {
-        TermuxService service = mTermService;
+        int index = mTermService.removeTermSession(finishedSession);
+        List<TerminalSession> remainingSessions = mTermService.getSessions();
 
-        int index = service.removeTermSession(finishedSession);
         mListViewAdapter.notifyDataSetChanged();
-        if (mTermService.getSessions().isEmpty()) {
-            // There are no sessions to show, so finish the activity.
+        if (remainingSessions.isEmpty())
             finish();
-        } else {
-            if (index >= service.getSessions().size()) {
-                index = service.getSessions().size() - 1;
-            }
-            switchToSession(service.getSessions().get(index));
-        }
+        else
+            switchToSession(remainingSessions.get(Math.min(index, remainingSessions.size() - 1)));
     }
 
 }
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -393,6 +393,11 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
             }
 
             @Override
+            public void onSessionClosingItself(TerminalSession session) {
+                removeFinishedSession(session);
+            }
+
+            @Override
             public void onClipboardText(TerminalSession session, String text) {
                 if (!mIsVisible) return;
                 showToast("Clipboard:\n\"" + text + "\"", false);
@@ -855,7 +860,6 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
     }
 
     public void removeFinishedSession(TerminalSession finishedSession) {
-        // Return pressed with finished session - remove it.
         TermuxService service = mTermService;
 
         int index = service.removeTermSession(finishedSession);
--- a/app/src/main/java/com/termux/app/TermuxService.java
+++ b/app/src/main/java/com/termux/app/TermuxService.java
@@ -323,6 +323,12 @@ public final class TermuxService extends Service implements SessionChangedCallba
     }
 
     @Override
+    public void onSessionClosingItself(TerminalSession session) {
+        if (mSessionChangeCallback != null)
+            mSessionChangeCallback.onSessionClosingItself(session);
+    }
+
+    @Override
     public void onTextChanged(TerminalSession changedSession) {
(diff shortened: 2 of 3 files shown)


fornwall:
Thanks! Will look into this and try it out.

rustybird:
rebased against current master

rustybird:
Should I improve something in this pull request? For example, I could remove the `Simplify removeFinishedSession()` commit if you think it does not belong here.

(FWIW, no bugs so far in my daily usage.)

fornwall:
@rustybird Sorry for the delay in feedback!

I'm not sure what the best course of action is here. After talking with some people using the app, perhaps it's best just to close sessions automatically on exit code 0, and let everything else (other exit codes and signals) cause the current behaviour where Enter is needed to proceed.

The rare case where someone messes up the startup (as in `exit 0` in `~/.profile`) could be handled by having a separate shortcut to a failsafe session.

What do you (and others) think about that?

atomi:
To be honest, tapping enter just isn't that big of a deal.

It might be worth skipping enter for ~/.shortcuts (perhaps a termux specific commandfor scripts) but again, I just don't find it that troublesome..

rustybird:
@fornwall:
> After talking with some people using the app, perhaps it's best just to close sessions automatically on exit code 0, and let everything else (other exit codes and signals) cause the current behaviour where Enter is needed to proceed.

Sounds good to me, closing on exit status 0 is more intuitive than SIGHUP for sure. Should I push an updated commit?

> The rare case where someone messes up the startup (as in `exit 0` in `~/.profile`) could be handled by having a separate shortcut to a failsafe session.

Maybe a double tap (or some other gesture) on Termux widgets?

ejaquay:
Love termux but find having to press Enter after pressing ^D annoying.  On any unix like system users can break their accounts with buggy init files and a dozen other ways,  With termux there is no admin to save users from their mistakes but pretty much anything a user can do to screw up his environment (buggy .profile, .bashrc, etc) can be fixed by either clearing termux data or re-installing it.  I would add an option in termux.properties to disable the prompt for Enter and advise users to periodically backup their work, maybe even provide a pair of scripts to do backup/restore to/from shared storage.  The restore script could also rename termux.properties as additional protection.

delt01:
I think the best approach would be an option in tasker that can be switched on/off by the user. Maybe a check box in the long-press menu? If the user's profile is somehow corrupted or messed up, it can always be fixed by editing the files using another app or adb.

cgarz:
I agree that if your just casually using termux pressing enter is no big deal. 

I am not however. I have it as an ssh server itself and a remote task starter for my server at home. Using the .shortcut folder and widgets. 

I have a large list of shortcuts and I often have to press a few.

Since it can't be closed programmatically I either have to:
Activate the shortcuts one by one. Which is painfully slow.
or:
tab out and press the next while the others are running. Creating a situation where I have to go back later and press enter sometimes as many as 10 times in a row.

Sheltering noobs is fine. But not at the expense of encumbering those who could easily fix simple problems like a bad .bashrc config.

I think any solution to this would make this already versatile and fantastic app much better.

IMO The seperate failsafe shortcut idea sounds the most noob friendly.

luisdavim:
Maybe this should be configurable in the `~/.termux/termux.properties` file.

kai11:
@fornwall any resolution on this?

alexlance:
Any movement on this? 

I would have thought the ideal behaviour from a widget is just to fire off a command and get out of your way. Like a tv remote control.  Hanging around waiting for you to press enter and clear the window out of recents is pretty annoying.

My understanding is that the whole point of the widget is to be a button. Users can pop open the full shell if they want, using the main app of course, but if you're using a termux widget then you really just want to push the button and forget about it. (Well imho anyway).

Btw termux is a fantastically useful app and cheers times a million for making it :+1:

Auxilus:
With a little searching I found that Exit status 129 (128+1) means you got
HUP kill signal for your session, that means either you got low ram or it
maybe your android os (I'd like to know whos os version are you on)

PS:This maybe incorrect....

alexlance:
Just found out you can put items in ~/.shortcuts/tasks/ and then they won't open the terminal at all. Awesome!

cgarz:
@alexlance is right. Just checked. No more enter spamming! yey! The shortcut widget prefixes the shortcut with task/ so some of my shortcut names are a little truncated. And the Termux window doesn't show at all so any info you might want to display while the shortcut is running will need to be changed to toasts. Really happy with this regardless. Would still like to have the window show with a no enter needed option as well though. For long messages toasts aren't ideal.

cgarz:
Further tested the .shortcut/tasks method - since termux doesn't show at all any script that needs user input via read command or similar won't work. Just realized when testing my home light control script which runs a script on my server via ssh and has a menu using the read command. So still stuck with press enter in this case.

lotheac:
>The rare case where someone messes up the startup (as in exit 0 in ~/.profile) could be handled by having a separate shortcut to a failsafe session.

There's always ways to break things. `kill $(cut -d' ' -f4 /proc/$$/stat)` (kill parent of current process; written this way because this busybox's ps does not seem to do -p) is a way to just kill the entire termux app on startup, and I don't see a way to recover from having that in shell initialization, aside from a separate failsafe shortcut.

delt01:
Very good point. I was also going to suggest a separate failsafe shortcut, or some way to start the app in "failsafe" mode.

Grimler91:
@delt01 the waiting for enter enables you to open the left drawer and open a failsafe session by long-pressing "new session", so it at least enables you to fix the problem (without root).
But I agree that a separate failsafe shortcut would be nice instead.

As a side note, I think termux has lots of uses even without root. I seldom use root even though my devices are rooted.

rustybird:
I'm closing this. Anyone still interested: Feel free to take the code, rebase it on master again, and open a new pull request.

PR #111, opened 2016-07-02
enable default shortcuts configuration from config file

aminehmida:
I use an AZERTY virtual keyboard, so I needed to be able to configure default shortcuts from config file.
I didn't add F1 - F10.
This will also keep the default bindings.

termux.properties:

```
left-key=a
right-key=d
up-key=w
down-key=s

page-up-key=p
page-down-key=n

tab-key=t
insert-key=i
home-key=h

underscore-key=u
pipe-key=p

escape-key=e
hat-key=.

jump-backward-key=b
jump-forward-key=f
emacs-x-key=x

show-volume-key=v

show-extra-keys-key=q
```

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxKeyListener.java
+++ b/app/src/main/java/com/termux/app/TermuxKeyListener.java
@@ -49,6 +49,96 @@ public final class TermuxKeyListener implements TerminalKeyListener {
     }
 
     @Override
+    public int getLeftKey() {
+        return mActivity.mSettings.mLeftKey;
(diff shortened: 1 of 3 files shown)


aminehmida:
Any chance you can integrate this upstream ?

fornwall:
I'm on a vacation with limited computer access, will look at this later (in about two weeks)!

TheDiamondYT1:
Once java is working on termux then that wont be a problem @fornwall

PR #132, opened 2016-08-08
Make ExtraKeysView work on Android 5

michalbednarski:
On Android 5 ExtraKeysView shows empty (though can be swiped for TextView).

This is fixed by explicitly specifying alignment as GridLayout.FILL

I have figured out that this was fixed in Android 6 in commit
https://android.googlesource.com/platform/frameworks/base/+/6dafd87fb4174447018b044bc67818d54fab57d8%5E%21/#F0
which set default alignment to FILL if weight is nonzero


fornwall:
Awesome find, really nice work!

The change is in the `v0.39` update of the app which should be available as an update shortly.

PR #194, opened 2016-11-18
Add italian translation

anthologist:

PR #217, opened 2016-12-25
German translation

mueller-ma:


mueller-ma:
Just saw #194

fornwall:
Right - I'm holding of with translations for now, but thanks for the offer!

PR #242, opened 2017-01-23
Super-quick hack for offline installation

xloem:
This is a quick change to the code so that the application does not require a network connection in order to be installed.

This implements a fix to issue #1 .

I needed to locally install this app offline, so implemented this, and am sharing my work for others.

This code could be generalized so as to provide one codebase that could be used for both offline and online installations, depending on whether the appropriate asset file is bundled.  This would be easy to do, but I have not done it yet.


F0RIS:
Thanks! It helped to me with this error, device Xperia ZR
device-2017-02-22-063301

lygstate:
ping for upgrade

lygstate:
@xloem we may create a termux-app-offline fork constantly sync with termux-app

xeffyr:
There no plans for implementing offline app currently.

Current bootstrap zips are large, about 15+ MB. In order to have proper offline installable Termux, the 8 (!!!) bootstrap zips has to be included into APK: aarch64, arm, i686, x86_64 for API 21 and 24. This will make Termux app unnecessarily large with about 80% unused asset payload.

Maximum allowed APK size in Google Play is 100 MB.

If you still need that, go with creating fork. You can use following patch as base: [0004-termux-v0.68-offline-bootstraps.git.patch.txt](https://github.com/termux/termux-app/files/3522234/0004-termux-v0.68-offline-bootstraps.git.patch.txt)

F0RIS:
@xeffyr May be App Bundle can resolve that?
https://developer.android.com/platform/technology/app-bundle

PR #243, opened 2017-01-25
Move more parts to external storage

Wetitpig:
Allow Termux to be installed on external storage, and extend permission to read_external_storage, as advised here:
https://developer.android.com/training/basics/data-storage/files.html


fornwall:
Hello and thanks! External storage does not allow executable permissions, so this would unfortunately not work for the installed binaries.

For the `READ_EXTERNAL_STORAGE` permission, this is already granted implicitly by `Any app that declares the WRITE_EXTERNAL_STORAGE permission is implicitly granted this permission`.

PR #330, opened 2017-06-02
Make cursor save/restore affect color attributes

whydoubt:
SGR attributes are stored in three variables: mEffect, mForeColor, and
mBackColor.  Saving/restoring the cursor only preserves mEffect.

Change the cursor save/restore methods to additionally preserve
mForeColor and mBackColor.  This affects both 'explit' saving/restoring
the cursor and switching to/from the alternate screen buffer.


fornwall:
Thanks a lot - this is included in the v0.51 update that is now rolling out as an update!

PR #331, opened 2017-06-03
Fix FC from too many control code parameters

whydoubt:
When the number of parameters in a CSI control code exceeds the size of
the mArgs array, the code may attempt to read past the end of the array,
resulting in a force close of the app.
    
Stop the index from moving past the last element of the mArgs array.


fornwall:
Thanks a lot - this is included in the v0.51 update that is now rolling out as an update!

PR #187, opened 2016-11-05
Android N multi-window support

pp3345:
This PR adds full multi-window support for Android 7.0+.

As soon as the framework enters multi-window mode a new button "New window" (keyboard shortcut Ctrl+Alt+W) becomes visible in the drawer menu. Clicking the button will create a new window. Terminal sessions are shared between windows, however, only one terminal window may attach to a specific session at a time. When creating a new window, the first unattached session will be attached to the new window. If no unattached sessions are available, a new session will be created.

Some screenshots of Termux running in multi-window mode on Android 7.1.1:
screenshot_1478367704
(split-screen)

screenshot_1478368051
(freeform mode)

In the second screenshot, for example, the left window may not be attached to session 2 and 3, since they are already attached to the other windows. Any window can switch to session 4.

Issue #29, opened 2015-12-18
Add support for Samsung's multi-window


joakim-noah:
Would be nice to have a web page or video open next to my Termux shell in splitscreen, [docs available here](http://developer.samsung.com/resources/multiwindow).

sarimak:
Extract from the docs is also available here: https://f-droid.org/forums/topic/request-add-support-for-samsung-multi-window/

Holzhaus:
Bump, this would be absolutely awesome.

makripodis:
That would be mandatory for me as well. I have the same need as joakim-noah, I am using an ssh client to my server to do this instead ov termux, so it would be realy great if you made the app split-view compatible!

fornwall:
Thanks to @bumper314 this will be available in a v0.52 update available later today!

joakim-noah:
Just downloaded the update and tried it out, works great!  Thanks, @bumper314.

PR #339, opened 2017-06-12
Add Multi Window support for Samsung devices

bumper314:
A single meta-data entry is enough to make Termux "Multi Window" aware on supported Samsung devices.
Tested and confirmed working on Samsung Note Pro 12.2 running Android 5.1.1.
screenshot_2017-06-12-01-22-28

This addition should not interfere with Android 6+'s split screen feature.
Tested and confirmed split screen functions as normal on a Nexus 6 running Android 7.0.

This addition should go unnoticed on non-Smasung Android 5 devices.
Tested and confirmed no functionality visible on Nexus 5 running Android 5.1.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -99,6 +99,7 @@
                   android:grantUriPermissions="true"
                   android:name="com.termux.app.TermuxOpenReceiver$ContentProvider" />
 
+        <meta-data android:name="com.sec.android.support.multiwindow" android:value="true" />
     </application>
 
 </manifest>


fornwall:
Looks great!

> This addition should not interfere with Android 6+'s split screen feature.

I guess this should be Android 7, as that was the version that introduced split screen AFAIK (https://developer.android.com/guide/topics/ui/multi-window.html)?

For my understanding: Does this affect also Samsung devices running Android 7.0 or later (does it change/improve the multiwindow handling on that version, or does Samsung just use the built-in Android multiwindow support as of 7.0+ regardless of the `com.sec.android.support.multiwindow` meta-data)?

bumper314:
> I guess this should be Android 7, as that was the version that introduced split screen AFAIK

Correct, my mistake.

> For my understanding: Does this affect also Samsung devices running Android 7.0 or later (does it change/improve the multiwindow handling on that version, or does Samsung just use the built-in Android multiwindow support as of 7.0+ regardless of the com.sec.android.support.multiwindow meta-data)?

I'm confident that Samsung will choose Android 7's multi-window over their own if both are present: http://developer.samsung.com/galaxy/multiwindow

Does anyone have a Samsung device on Android 7 who can confirm?
Here's an APK http://d.pr/f/w2he6z

bumper314:
Just tested on two Samsung tablets.
- Android 6 works as expected using Samsung's multi-window support.
- Android 7 works as expected, and appears to use Android 7's native split-screen with some additional feature (like being able to pop up a floating window).

Seems like a fairly safe merge.  After that, I'll look into supporting multiple instances (which Samsung's multi window API supports) leveraging this merge request:
  https://github.com/termux/termux-app/pull/187

bumper314:
Just noticed this addresses Issue #29

fornwall:
@bumper314 Thanks a lot - this will be available in a v0.52 update available later today!

Yajo:
Amazing how simple this was! I don't understand why don't everybody add it...
Thanks!

PR #344, opened 2017-06-20
Basic accessibility support for the terminal view

pvagner:
Terminal view as implemented in the termux app is a view resembling terminal windows known from computers. It's designed to be efficient however for me and other people who are using accessibility services such as Google Talkback to access their android devices it's not very usable. By adding contentDescription property to the view dynamically as it repaints, we can get at least very basic accessibility support.
When using termux on Android with Talkback running it is possible to review currently visible text on the screen by characters, by words and by lines.
It is possible to use two finger scrolling (android built-in feature) to scroll the view and access other parts of the text.
I'm using this for a few months already and I have got some friends interested in this functionality so I'm submitting this as a pull request.
I know it's very minimalistic, but even in its current limited form it's a life changing feature for tech savvy blind people so hopefully you can accept it and / or point out how I might be able to enhance it so everyone can benefit.

### DIFF
--- a/terminal-view/src/main/java/com/termux/view/TerminalView.java
+++ b/terminal-view/src/main/java/com/termux/view/TerminalView.java
@@ -15,6 +15,7 @@ import android.text.InputType;
 import android.text.TextUtils;
 import android.util.AttributeSet;
 import android.util.Log;
+import android.view.accessibility.AccessibilityManager;
 import android.view.ActionMode;
 import android.view.HapticFeedbackConstants;
 import android.view.InputDevice;
@@ -74,6 +75,8 @@ public final class TerminalView extends View {
 
     /** If non-zero, this is the last unicode code point received if that was a combining character. */
     int mCombiningAccent;
+
+    private boolean mAccessibilityEnabled;
 
     public TerminalView(Context context, AttributeSet attributes) { // NO_UCD (unused code)
         super(context, attributes);
@@ -197,6 +200,8 @@ public final class TerminalView extends View {
             }
         });
         mScroller = new Scroller(context);
+        AccessibilityManager am = (AccessibilityManager) context.getSystemService(context.ACCESSIBILITY_SERVICE);
+        mAccessibilityEnabled = am.isEnabled();
     }
 
     /**
@@ -384,7 +389,7 @@ public final class TerminalView extends View {
         mEmulator.clearScrollCounter();
 
         invalidate();
-        setContentDescription(getText());
+        if (mAccessibilityEnabled) setContentDescription(getText());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


fornwall:
Thanks a lot!

Is it necessary to call `setContentDescription(getText())` on each render call, or would it be enough to only call it when updating the screen in `TerminalView.onScreenUpdated()`?

pvagner:
Hello,
I just wanted to make sure we will get all the updates.
Now when thinking more about it I assume
TerminalView.onScreenUpdated()
would be enough.
I'll change and test it today.

pvagner:
Changed, tested and working fine.

pvagner:
@fornwall Please can you take a look one more time?
I hope it's changed as you have suggested.

stormdragon2976:
Did this ever get added to Termux? I installed it today and wasn't able to read the terminal's output. Is there something I need to do to get it working with Talkback?

stormdragon2976:
I'm having lots of issues with the jre and android-studio playing nicely together. Is there an apk somewhere that I can just install until this is merged?

Grimler91:
@stormdragon2976 you can try this one: https://grimler.se/apps/termux-app-access.apk
I have the other apps available from there as well, signed with the same key. 

I haven't tested it at all so it's great if you are willing to!

stormdragon2976:
Thanks so much for this. It is very helpful. The accessibility is very basic indeed, but it does work. Definitely a good start.

fornwall:
@pvagner I'm really sorry for forgetting about this one, and I would really like to see it included!

Could you (or anyone else interested) have a look at enabling this only if accessibility services are enabled (or something similar), as this may slow down the app needlessly if not?

Perhaps with something like this in the TerminalView constructor

```java
private boolean mAccessibilityEnabled;
[...]
public TerminalView(Context context, AttributeSet attributes) {
[...]
    AccessibilityManager am = (AccessibilityManager) context.getSystemService(ACCESSIBILITY_SERVICE);
    mAccessibilityEnabled = am.isEnabled();
```

and then guarding the call with `if (mAccessibilityEnabled) setContentDescription(getText());`. Would this work?

pvagner:
I'll be able to use and integrate this code snipped in a week or so. I'm sure something like this is doable.

fornwall:
Thanks a lot!

It's beyond the scope of this immediate change, but could you give some pointers about possible next steps to improve accessibility support?

fornwall:
Thanks a lot, both for this change and the suggestions for further improvements!

vortex1024:
any idea about what would be required to make new text be read automatically in the terminal? I guess using an accessibility live region for the whole terminal view won't work, as the whole terminal would be read out, not only the new text. A hidden accessibility region containing only the new text would work, but sounds like an ugly hack. Other ideas? Maybe something linked to the accessibility node info @pvagner talked about?

PR #360, opened 2017-07-13
Add Dual Window support for LG devices

bumper314:
Similar to pull request #339.

A single meta-data entry is enough to make Termux "Dual Window" aware on supported LG devices.

screenshot_2017-07-13-15-14-11

Tested and confirmed working on LG VK815 tablet running Android 6.0.

Source: http://mobile.developer.lge.com/develop/dev-guides/lg-dual-window-guide/development/enabling-dual-window/

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -112,6 +112,7 @@
                   android:name="com.termux.app.TermuxOpenReceiver$ContentProvider" />
 
         <meta-data android:name="com.sec.android.support.multiwindow" android:value="true" />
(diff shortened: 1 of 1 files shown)


fornwall:
I'm going through open PR:s, sorry for the huge delay here. I don't think this is relevant anymore due to the native Android multi-window support now - let me know if I'm wrong!

bumper314:
I still think this is relevant because native Android multi-window support didn't come until Android 7, while this PR gives dual window support to LG devices running Android 6.  My LG tablet will never get Android 7, so this one-line PR is both helpful and safe.

PR #399, opened 2017-08-24
Make statusbar transparent

jmi2k:
This PR makes the statusbar transparent so it can blend with the current theme. Some notes:

- It only affects API >= 23, so it can turn the statusbar icons dark for light themes.
- It requires an additional property in all themes. I addressed it in this PR *TODO*.


Indranil012:
Could i know why this is not implemented yet? Is it too hard to fix?

PR #417, opened 2017-09-09
Added SHIFT to extraKeys

Auxilus:
I might have missed something here ( like backend code to handle shift keypress event ), but hey! You can do that right?


nymsa:
Will shift+F5 be possible in mcedit?

trygveaa:
@nymsa: This PR is completely useless, it just adds the button without any functionality. So with this PR, no it won't be possible. However, if someone implements it properly, it should work with shift+F5 too.

PR #428, opened 2017-09-17
Updated tr (Turkish) translation

mission712:
Commit made via [Stringlate](https://lonamiwebs.github.io/stringlate/)


fornwall:
Sorry, but the app is currently not trying to be localized!

PR #433, opened 2017-09-23
Updating manifest to support resizeableActivity

twoboxen:
Updating manifest to support resizeableActivity for main app and help screen. 

[This is also used by Samsung's DeX mode.](http://developer.samsung.com/samsung-dex/modify-optimizing)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -27,7 +27,9 @@
             android:name="com.termux.app.TermuxActivity"
             android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
             android:launchMode="singleTask"
-            android:windowSoftInputMode="adjustResize|stateAlwaysVisible" >
+            android:windowSoftInputMode="adjustResize|stateAlwaysVisible"
+            android:resizeableActivity="true"
+            >
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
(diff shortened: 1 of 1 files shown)


pp3345:
This should not make any difference as resizeableActivity defaults to true already.

twoboxen:
Why doesn't it work on dex already then?

fornwall:
As @pp3345 says this attribute should already default to true:

    If your app targets API level 24, but you do not specify a value for this attribute, the attribute's value defaults to true.

From https://developer.android.com/guide/topics/ui/multi-window.html.

So dex problems should come from something else.

zhunsaker:
question 3 samsung developer guide


Your app might have targeting the wrong targetSDKVersion. Android:resizeableActivity is supported when the targetSDK version is greater than or equal to 24 (Android N). IF targetSDKVesion is under 24, activities are only resizable when the screenOrientation is not set as fixed. Check to make sure you do not have this code in your manifest: 

<activity android:screenOrientation="portrait | landscape"> </activity>

fornwall:
@twoboxen Sorry about this - you were right in that it was resizeableActivity="true" which was needed. I still don't understand why it is needed as it should be the default value when the target API is 24+ (could it be a Samsung bug perhaps), but anyway it works now.

twoboxen:
Hey no worries.  I'm just glad you got it working!

Thanks!!

PR #434, opened 2017-09-25
Update TerminalSession.java

Auxilus:
Endprocess string modified to check for `~/hushlogout` if exists then exit the terminal without echoing exitDescription

### DIFF
--- a/terminal-emulator/src/main/java/com/termux/terminal/TerminalSession.java
+++ b/terminal-emulator/src/main/java/com/termux/terminal/TerminalSession.java
@@ -125,7 +125,7 @@ public final class TerminalSession extends TerminalOutput {
                     if (exists) { 
                         int exitCode = (Integer) msg.obj;
                         if (exitCode == 0) {
-                            exitDescription = "session ended with exit code 0";
+                            String exitDescription = "session ended with exit code 0";
                             byte[] bytesToWrite = exitDescription.getBytes(StandardCharsets.UTF_8);
                             mEmulator.append(bytesToWrite, bytesToWrite.length);
                             notifyScreenUpdate();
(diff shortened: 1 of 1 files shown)


terminal-emulator/src/main/java/com/termux/terminal/TerminalSession.java
```diff
@@ -115,24 +116,42 @@ public void handleMessage(Message msg) {
...
+            
+            else {
+                if (msg.what == MSG_PROCESS_EXITED) {
+                    File tmpDir = new File("/data/data/com.termux/files/home/.hushlogout");
+                    boolean exists = tmpDir.exists();
+                    if (exists) { 
+                        int exitCode = (Integer) msg.obj;
+                        mShellExitStatus = exitCode;
+                        mShellPid = -1;
+                        mTerminalToProcessIOQueue.close();                         mProcessToTerminalIOQueue.close();
```
Grimler91: I'm guessing this shouldn't be here twice?

Grimler91:
Wouldn't this sort of undo the very thing that the exit message is suppose to do? 
If a user breaks his environment then there would be no way to launch a failsafe terminal if the terminal exists immediately. 

Maybe it's possible to checkout for /sdcard/hushlogout instead, as @Neo-Oli suggested [here](https://github.com/termux/termux-app/issues/56#issuecomment-192176747)

Auxilus:
Hmm, that sounds like a problem, but I just made temp changes..
Should I add something like warning when used tries `touch ~/hushlogout` ?

Auxilus:
I'll just replace `~/hushlogout` by `/sdcard/hushlogout`

terminal-emulator/src/main/java/com/termux/terminal/TerminalSession.java
Auxilus: Solved..

cgarz:
@Grimler91 True, but the exit message wont protect you from everything anyway. Such as killall com.termux in your bashrc. A way to properly close termux panes programatically is sorely needed.
Many, including myself have said as much on various issues such as [Issue 56](https://github.com/termux/termux-app/issues/56)

fornwall:
Not relevant after #56.

PR #436, opened 2017-09-25
replace help url with wiki url

Auxilus:
HELP page wiki changed to https://wiki.termux.com/wiki/Main_Page as it now has more info than http://termux.com/help.html

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxHelpActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxHelpActivity.java
@@ -39,7 +39,7 @@ public final class TermuxHelpActivity extends Activity {
         mWebView.setWebViewClient(new WebViewClient() {
             @Override
             public boolean shouldOverrideUrlLoading(WebView view, String url) {
-                if (url.startsWith("https://termux.com")) {
+                if (url.startsWith("https://wiki.termux.com")) {
                     // Inline help.
                     setContentView(progressLayout);
                     return false;
@@ -60,7 +60,7 @@ public final class TermuxHelpActivity extends Activity {
                 setContentView(mWebView);
             }
         });
-        mWebView.loadUrl("https://termux.com/help.html");
+        mWebView.loadUrl("https://wiki.termux.com/wiki/Main_Page");
     }
 
     @Override


sdrausty:
@auxilus I don't think this pull request is a great idea. 

1. Anyone can change the help on the wiki. What if someone mucks up the wiki help. That will find its' way in the app immediately. 

2. We can always improve the help pages through pulls to https://github.com/termux/termux.github.io/blob/master/help.md and similar. 

What if someone spams the help pages on the wiki? Or changes the help pages maliciously? How useful will that be?

Auxilus:
As long as @Neo-Oli is actively monitoring the wiki changes, we are good.. Afaik wiki.termux.com has more content than help page on termux.com. Also I think @fornwall is okay with considering wiki.termux.com as official wiki

PS: open editing is what makes wiki more informative, also you'll need active moderation team to keep track of it..

sdrausty:
The correct way would be to modify https://github.com/termux/termux.github.io/blob/master/help.md where help is secure, not on a wiki. Can you please show me a successful major project where the help inside the app is hosted on a wiki? Let the wiki be complementary help. Why eliminate the good help pages we have?

sdrausty:
Adding a wiki link to https://github.com/termux/termux.github.io/blob/master/help.md is a suggestion. Help on long press from github - link to wiki in the help pages would be OK.

sdrausty:
Imho wiki.termux.com should be hosted at http://pages.github.com

1. No spam. 
2. User accountability. 
3. Practicing simple pull requests. 
4. ETC

We have enough content to do this.

Neo-Oli:
> Can you please show me a successful major project where the help inside the app is hosted on a wiki?
Both Archlinux and LineageOS only have a wiki as their primary documentation.

Neo-Oli:
> Imho wiki.termux.com should be hosted at http://pages.github.com

That's exactly what the help pages were. There were almost no contributions made for years.

sdrausty:
@Neo-Oli These are not apps. Are the man and info pages on a wiki too.

Neo-Oli:
> Are the man and info pages for these OSes on a wiki too?

The man and info pages for Termux are also not on a wiki.

sdrausty:
@Neo-Oli >That's exactly what the help pages were. There were almost no contributions made for years.
Great. Lets add a few links to them then. The wiki should be complementary, nor overriding.

sdrausty:
Anyone can change the help on the wiki. What if someone mucks up the wiki help. That will find its' way into the app immediately. Link to wiki pages from the help pages is my suggestion, i.e. 👉 See Also

sdrausty:
I can see this temptation already. Someone adds rm -rf $PREFIX noob to long press help in our app.

sdrausty:
Or, long press help in Termux, and out pops, "Get Viagra HTTPS Link" or something similar. Temptation for many spammers and equivalent.

fornwall:
I currently think that the wiki is the way forward, and that the advantages of having information that can be easily updated and contributed to outweighs the risks.

Let's try it out, if there is a serious problem we can always switch back or to something else.

Auxilus:
As I mentioned earlier if we switch help link to wiki.termux.com, we must have active moderation team to monitor the changes as site traffic would be increased.

PR #438, opened 2017-09-26
added RECORD_AUDIO permission to manifest

juha-h:
Without RECORD_AUDIO permission it is not possible to execute from termux shell applications, such as baresip VOIP user agent, that use OpenSL ES API for playing and recording audio.


juha-h:
Not needed (install termux-api and give it microphone permission).

PR #439, opened 2017-09-26
Add fish and tcsh to shellBinary

Auxilus:


fornwall:
This is not needed - the `login` script takes care of running the right shell (the fallback to `bash` or `zsh` is a historical note for old Termux systems that haven't been updated in the last year and doesn't have login).

Issue #446, opened 2017-10-07
 [Feature Request] Termux as a downloader


sayem314:
Hi there, what I meant is can we register termux as a downloader globally? So that we could use wget/aria2 as big files downloader. Termux wget/aria2 works better than phone built-in downloader, especially when the file is really large.

<img src="https://i.imgur.com/uNSEQVU.png" width="40%">

My `termux-url-opener` context is `aria2c -x 4 "$1" --dir=$HOME/storage/downloads` and it works super smooth however it would be awesome if there is an option to make android understand termux can download.

Android download manager reference: https://developer.android.com/reference/android/app/DownloadManager.html

setuidroot:
I understand the feature you are asking for and I also think it'd be a cool addition.  While you could probably add it in yourself (given a rooted phone and some bash scripting.) I just wanted to tell you how I personally do this hoping it may help some people...

As an example in your screenshot, just click and and hold the "download" button and then click "copy link address" in the pop-up prompt.  Then open termux and do: wget (paste address/Cntl+Alt+V on hacker's keryboard landscape mode.)

This is an easy way to use termux to download files... I do it all the time b/c termux is far more reliable a downloader and I believe if you lose network for a second (like switching AP around a home WiFi does annoyingly on some phones) the download will continue where it left off as opposed to leaving half a downloaded file in storage and making you start all over again :/

Also tmux downloads the file to it's internal data path if youless you specify otherwise.  I just use: 'mv downloaded.file /sdcard/Download' (no quotes) to move it to my regular downloads folder when it's done.

Quasic:
I use it to fix aborted downloads by other downloaders. Just copy the partial file to a termux r/w location, and use either wget's -c option or curl's -C or --continue-at option.

alive4ever:
As a workaround, there is [Aria2App](/devgianlu/Aria2App) which can be used to achieve similar purpose, i.e. using aria2c with rpc enabled inside termux and using Aria2App to capture downloads on Android.

wqzq:
I discovered that you can redirect certain browser's (like Kiwi Browser) downloads to aria using RPC. This is what worked for me:

1. installed this chromium extension https://github.com/alexhua/Aria2-for-chrome on Kiwi Browser
2. ran `aria2c --dir=/storage/emulated/0/Download --enable-rpc` on Termux
3. went back to Kiwi Browser and configured the extension to intercept all downloads by default

[...]

PR #448, opened 2017-10-08
Termux as a downloader #446

sayem314:
This would add feature requested on issue #446., user need to have termux-url-opener for proper usage tho!

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -17,8 +17,8 @@ android {
         applicationId "com.termux"
         minSdkVersion project.properties.minSdkVersion.toInteger()
         targetSdkVersion project.properties.targetSdkVersion.toInteger()
-        versionCode 107
-        versionName "0.107"
+        versionCode 108
+        versionName "0.108"
 
         externalNativeBuild {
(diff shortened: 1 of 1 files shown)


sayem314:
@fbartels 8 months for approving changes! Seriously? 😞

fbartels:
Not that my approval does mean that @fornwall will merge it anytime soon, but I would like to see this feature as well (just stumbled upon this today).

sayem314:
Ah I see. Thanks 🙏

justworks007:
Please this feature would be amazing.

buttaface:
Interesting, @fornwall would have to decide this.

wqzq:
@xeffyr would you take a look at this please.

buttaface:
Even if this got in, it would not be deployed to most users in the Play store, as [the app there is no longer updated](https://wiki.termux.com/wiki/Termux_Google_Play). If you use F-Droid, it would be updated.

@fornwall should approve this because I think he has intentionally kept permissions minimal so far.

PR #470, opened 2017-11-09
set default notification priority to low

lotheac:
since we're a foreground service, in oreo we already get a higher
priority; see
https://developer.android.com/reference/android/app/NotificationManager.html#IMPORTANCE_MIN


fornwall:
Thanks! This will be in the next `v0.55` release.

PR #474, opened 2017-11-15
espresso test

stephengroat:
an espresso test for termux-app that runs a command and sends the output to logcat


stephengroat:
any interest in using active UI tests?

PR #478, opened 2017-11-15
use /data/data/com.termux/cache as TMPDIR

xeffyr:
Solution for https://github.com/termux/termux-packages/issues/1772


xeffyr:
Closing for the same reason as https://github.com/termux/termux-packages/pull/1788

PR #529, opened 2018-01-01
Fix notification icon color

moneytoo:
Icons used in notifications/status bar are monochrome and should use only white color and transparency. It is displayed fine on Android emulators but it's shown in black on Samsung Galaxy S7 running Android 7.0.

screenshot_20180101-200025

After the change, status bar icon is white and the small icon in notification still uses the black color set via setColor(). Termux doesn't support devices running Android older than 5.0 but using white color in vector drawables would be also necessary to generate proper (white) bitmap icons.


fornwall:
Thanks a lot! It will be in the next version, v0.60, which will be available as an update within a day or so!

Issue #572, opened 2018-02-04
/sdcard points to "Internal storage"


mbrick02:
I ran 'termux-setup-storage' when I had a bad sd-card (I'm assuming this is the issue).  Now (with a new, working sdcard) '/sdcard' points to local storage.  I attempted to run 'termux-setup-storage' again but 'sdcard' still points to local storage.  I'm running Android 7.1.1

Grimler91:
/sdcard always points to internal storage, it is unrelated to termux. After running termux-setup-storage you have a folder ~/storage which contains symlinks to internal storage (~/storage/shared), sd card (~/storage/external-1) and some other places. Run `ls -l ~/storage/` to see where the symlinks point.

mbrick02:
all sym links point to (->): '/storage/emulated/0' which is internal storage. I see no '~/storage/external-1'

Quasic:
I had to create the sdcard folder and symlink manually. The actual path depends on which system you have. For me, it was /storage/external_SD/Android/data/com.termux/files

Quasic:
I just tried moving ~/storage to ~/storagebak and rerunning termux-setup-storage, and it worked. I'll merge the folders.

Quasic:
@mbrick02, I think you were right, if run without a working SD card, you might need to remove ~/storage and run ```termux-setup-storage``` again.

mbrick02:
@Quasic Thanks, so much, that worked!

Quasic:
Glad to share the solution! I just found out it sends a message to logcat. Maybe it should print it to the user, which is currently disabled, probably because am prints out stuff...
```
I/ViewRootImpl(15272): ViewRoot's Touch Event : ACTION_UP
I/ViewRootImpl(15272): ViewRoot's Touch Event : ACTION_DOWN
I/ViewRootImpl(15272): ViewRoot's Touch Event : ACTION_UP
I/ViewRootImpl(15272): ViewRoot's Touch Event : ACTION_DOWN
I/ViewRootImpl(15272): ViewRoot's Touch Event : ACTION_UP
D/AndroidRuntime(15577): >>>>>> START com.android.internal.os.RuntimeInit uid 10115 <<<<<<
D/AndroidRuntime(15577): CheckJNI is OFF
D/AndroidRuntime(15577): Calling main entry com.android.commands.am.Am
D/AndroidRuntime(15577): Shutting down VM
E/termux-storage(15272): Could not delete old $HOME/storage
[...]

PR #578, opened 2018-02-06
Fix for #572: merge ~/storage instead of del

Quasic:
We could recursively delete, as in the method above, but that would
require a separate try block for a separate logcat message, or a
different method that returns a flag. I prefer merging in case of custom
folders. Another alternative may be to send an exit code or something for
the termux-setup-storage script to detect and echo an error message.


Quasic:
To merge effectively, we'd probably need checks for each existing symlink, as well.

Quasic:
Travis is set up on my fork, now, so I'll close this for the time being.

PR #579, opened 2018-02-07
 Fix for #572 using recursive delete

Quasic:
Also see #578 for a merging alternative WIP

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxInstaller.java
+++ b/app/src/main/java/com/termux/app/TermuxInstaller.java
@@ -229,12 +229,12 @@ final class TermuxInstaller {
                     File storageDir = new File(TermuxService.HOME_PATH, "storage");
 
                     if (storageDir.exists()) {
-			            try {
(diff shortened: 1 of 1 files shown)


l-jonas:
@Quasic What is the benefit of this/ is there any issue with the current implementation? Could you add spaces at the left so that your code looks like it's inside the if-block?

Quasic:
Having a backlog ATM.

Without checking the current situation, #572 gives the issue at the time I wrote this. Basically, running '''termux-setup-storage''' didn't add a symlink for a newly inserted SD card, because it had previously been run without any SD cards inserted.

In some ways, I actually prefer #578, but this one would also remove symlinks to removed SD cards.

I'll try to check current status and look at formatting when I get time. It sounds like tabs were inserted instead of spaces?

fornwall:
@Quasic Thanks a lot, and sorry for the delay here! This will be available in the next update (v0.62) which will be released in a few days.

PR #618, opened 2018-03-10
Make TerminalRenderer public

RangerMauve:
I want to use the renderer with a custom canvas without having to render to an Android View which requires a context and all sorts of stuff.

Can't currently do that because the renderer is package-private.

The goal is to be able to render this onto a texture in Unity so that people can easily make stuff with termux in VR.


fornwall:
Thanks! The TerminalRenderer class needs some love and cleanup, but having it public doesn't do any harm and allows projects such as yours to experiment with it.

RangerMauve:
Thanks! I'm kinda new to android development, how can I configure my build.gradle to load this version?

Issue #603, opened 2018-03-01
Terminal output speed issues


evg-zhabotinsky:
This is probably the same as #218 and #368, but I felt that they might not describe the problem adequately.

Terminal output is too slow on many different occasions. When Termux cannot keep up with produced output, it stutters at best and becomes unresponsive for minutes at worst. As far as I can tell, using ANSI escapes (for colors, ncurses ui, etc.) makes things even worse.

Most typical manifestations in my experience:
* Scrolling in `less` and similar utilities. Scrolling down is fine, scronling up redraws whole screen and lags like hell.
* Build logs. Use minimal verbosity and you should be fine. Up verbosity a bit or get a lot of warnings due to a typo in a header, and your terminal is busted. Screen updates once every few seconds, and you'll be catching up with an already finished build for many minutes. (I usually just kill Termux and reconnect to my `tmux` session.)

This has been happening for a long while, half a year at very least. However, after one of the most recent updates (noticed on 0.60), the problem got *much* worse. On my Samsung tablet running Android 5.1.1, 80x25 terminal is mostly bearable, while on 190x40 `vim` takes _almost a minute_ to start and regularly hangs for a few seconds (e.g. on newline entry). Not nearly as bad on my phone running 6.0.1, but still noticeable. 

I haven't yet looked at source code (don't have time right now, neither do I have Android app development experience), but I intend to unless it's fixed in a week or two. I suspect excessive syncing among pty and UI threads and maybe inefficient scrollback handling.

xeffyr:
>80x25 terminal is mostly bearable, while on 190x40 vim takes almost a minute to start and regularly hangs for a few seconds (e.g. on newline entry).

On Nexus 7 (Android 5.1.1) no such problem. It takes ~2 seconds to start vim on 182x53 and no lags when editing files. Are you installed some plugins ?
On Samsung Galaxy S7 it runs very fast without problems.

I never had lags when using Termux except when I connected over SSH with bad wifi signal.

evg-zhabotinsky:
> It takes ~2 seconds to start vim

That's already long enough. A heavily modded Vim (a vim-based IDE called Yavide) starts instantaneously (before I even release `Enter` key) on my laptop which has just slightly higher CPU frequency.

[...]

PR #628, opened 2018-03-18
Add fast path for TerminalRow.setChar

michalbednarski:
Added fast path for TerminalRow.setChar for case where theres 1-to-1 mapping between characters and columns (that is no UTF-16 surrogate pairs and characters with width != 1)

Trying to fix termux/termux-app#603, please test @evg-zhabotinsky

In my tests this highly improved performance of `cat 500KBfile.txt` test.

Still slower than Android Terminal Emulator, but difference is now much smaller and seems to be matter of redraw throttling, without redrawing Termux is now faster (For test commented out line in [Termux](https://github.com/termux/termux-app/blob/9f36ed06b8e2c3b5e5918444e0c37699e21a7196/terminal-view/src/main/java/com/termux/view/TerminalView.java#L386) and [Android Terminal Emulator](https://github.com/jackpal/Android-Terminal-Emulator/blob/987e4eb56bb7bcb387a187ab78437111d327c984/emulatorview/src/main/java/jackpal/androidterm/emulatorview/EmulatorView.java#L488), press volume key in such version to redraw screen)

### DIFF
--- a/terminal-emulator/src/main/java/com/termux/terminal/TerminalRow.java
+++ b/terminal-emulator/src/main/java/com/termux/terminal/TerminalRow.java
@@ -21,6 +21,8 @@ public final class TerminalRow {
     boolean mLineWrap;
     /** The style bits of each cell in the row. See {@link TextStyle}. */
     final long[] mStyle;
+    /** If this row might contain chars with width != 1, used for deactivating fast path */
+    boolean mHasNonOneWidthOrSurrogateChars;
 
     /** Construct a blank row (containing only whitespace, ' ') with a specified style. */
     public TerminalRow(int columns, long style) {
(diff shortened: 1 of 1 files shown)


evg-zhabotinsky:
Huh. So it was mostly paranoid unicode handling and not syncing?
Works fine for me, ~5MB file `cat`ing time is down to ~1.5 minutes from ~1 hour before the fix. `vim` also starts up instantly now. Could be faster, but at that point human is the bottleneck most of the time.

Still, it's a stopgap, and the other code path still needs to be optimized. Suggest keeping the issue open until then.

fornwall:
@michalbednarski Sorry for the delay here! Looks interesting, nice gains without messing up the code.

I think `mHasNonOneWidthOrSurrogateChars` should be reset to false the `clear()` method, since otherwise over time all rows will end up having this set to true (as lines are reused after a while in the circular buffer).

michalbednarski:
Rebased and added to flag reset to `clear`.

fornwall:
Thanks! This will be in v0.62 of the app, to be released in a few days.

PR #712, opened 2018-06-11
Replace \n with \r when pasting (Fixes #678)

### LINKED ISSUE #678: Pasting via long press causes issues in some programs
Newlines when pasting have strange effects.

This is most noticeable and reproducable in nano.

Steps to reproduce with nano:
Put this into a file:
```
This
Is
A
Test
```
Copy the file contents to the clipboard:
```
cat file | termux-clipboard-set
```
Open that same file in `nano -w`, put the cursor to the beginning, long press, paste.
```
ThisThis Is A Test Is A
Test
```
Should be:
```
This
Is
A
Test
This
Is
A
Test
```
or similar. The same version of `nano` on iTerm2 has no issue, so it is likely a Termux issue.

`cat` also has this issue:
Manually:
```
$ cat
This
This
Is
Is
A
A
Test
Test^D
$
```
With clipboard:
```
$ cat
This
Is
A
TestThis
Is
A^D
Test^D$
```
Upon the first `^D`, Test is shown, and then the second finishes the `cat` without an ending newline.

easyaspi314:
Termux will now properly send `\r` to the terminal instead of `\n` when pasting multiline strings.

This fixes `cat` not repeating back lines and `nano` accidentally justifying text (because `\n` maps to `^J`), as well as other potential issues.

This matches the behavior of other terminals, such as iTerm2 which explicitly does it here:
https://github.com/gnachman/iTerm2/blob/f8a5930/sources/iTermPasteHelper.m#L113

### DIFF
--- a/terminal-emulator/src/main/java/com/termux/terminal/TerminalEmulator.java
+++ b/terminal-emulator/src/main/java/com/termux/terminal/TerminalEmulator.java
@@ -2338,8 +2338,8 @@ public final class TerminalEmulator {
     public void paste(String text) {
         // First: Always remove escape key and C1 control characters [0x80,0x9F]:
         text = text.replaceAll("(\u001B|[\u0080-\u009F])", "");
-        // Second: Replace all newlines (\n) with carriage returns (\r).
-        text = text.replace('\n', '\r');
+        // Second: Replace all newlines (\n) or CRLF (\r\n) with carriage returns (\r).
+        text = text.replaceAll("\r?\n", "\r");
 
         // Then: Implement bracketed paste mode if enabled:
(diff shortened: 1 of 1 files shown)


easyaspi314:
This *should* do. Didn't want to build the whole thing, but I did a quick test:

```java
String str = "\r \r\n \n";
String str2 = str.replaceAll("\r?\n", "\n");
if ("\r \r \r".equals(str2))
    System.out.println("PASS");
```

fornwall:
Thanks a lot! This will be available in a v0.61 update which will be available shortly.

fornwall:
This has been fixed in version 0.61 of the app, which has now been released and should be available in Google Play within a few hours (and on F-Droid within a day or so)!

Issue #556, opened 2018-01-26
Feature request: Option to (temporarily) hide the row with special on-screen keys, e.g., when a bluetooth keyboard is attached


tomtom:
Users can long-press on the "keyboard" button in the sidebar to show a row of special on-screen keys. This makes termux much more accessible when users use non-specialized on-screen keyboards.

If users connect to a bluetooth/usb keyboard, the on-screen keyboard is hidden but the extra row is still being displayed and wastes screen space. While users could of course switch the row with special keys off by long-pressing on the "keyboard" button, I think this is unnecessarily cumbersome.

I would like to suggest to implement either one of the following changes:

1. Hide the row, when the on-screen keyboard is hidden.
2. Hide the row, when a hardware keyboard (bluetooth or usb) is being used.

I think this would improve the usability of termux.

vganyn:
+

Grimler91:
Similar to https://github.com/termux/termux-app/issues/556, let's track it there

tomtom:
Are you saying issue #556 is similar to #556 which is why you close #556?

Grimler91:
Ops, meant to paste https://github.com/termux/termux-app/issues/140

Issue #140, opened 2016-08-14
Hide extra keys row when Keyboard is hidden


Neo-Oli:
When the extra keys row is enabled, can it be automatically hidden when the keyboard gets hidden and automatically shown again when the keyboard gets displayed again?

Neo-Oli:
I realized that this already happens when fullscreen mode is enabled. Is there a reason it doesn't hide in the non-fullscreen mode?

bagnacauda:
This would be really nice when using a phisical keyboard, considering that fullscreen mode is gone.

PR #714, opened 2018-06-11
Hide the keyboard extension when the keyboard is hidden

l-jonas:
This fixes https://github.com/termux/termux-app/issues/556

There is one (small) issue with this: It depends on the resizing of the
App caused by the touchscreen keyboard. Due to that, it always thinks that the
keyboard is hidden when the Android splitscreen is used in portrait mode.

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -25,7 +25,9 @@ import android.media.SoundPool;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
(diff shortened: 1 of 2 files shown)


l-jonas:
This closes https://github.com/termux/termux-app/issues/140

robertvandeneynde:
This should be configurable in the .properties, I like it sometimes to have only the custom keyboard (that is now fully configurable). I suggest thr option `extra-keys.hide-with-classic-keyboard = true`

buttaface:
User was deleted and pull is conflicting now, someone can update and resubmit if wanted.

Issue #442, opened 2017-10-04
Termux should handle commands spamming the bell better.


easyaspi314:
```sh
$ mv ~/.termux/termux.properties ~/.termux.properties.bak # reset to vibrate only
<relaunch>
$ while ; do tput bel &; done; # more temporary alternative: cat `which bash`
```
Snap, crackle, pop, until you *force stop*. No other way to stop it. Termux freezes, so you can't Ctrl+C, hit Exit on the notification, or swipe away.
Need I say more?

Ideas:
* Have a minimum delay between bells, probably 25 ms (half vibration time) at the least. This will fix the crackling from the Vibrator trying to activate before it is "ready". If we get any bell signals during that delay, we will wait until that delay is over, then beep/vibrate once:
``` 
Signals: bell bell bell bell bell bell bell bell bell
Output:  bell                bell                bell
Signals: bell     bell     
Output:  bell                bell
Signals: bell                           bell
Output:  bell                           bell
```
* Don't block the thread so the user can Ctrl+C it. This stupidity in the Android OS seems relevant; as it seems that Termux is attempting to run the bell on a separate thread.
> "Like many callbacks from the system, the methods on this class are called from the main thread of your process." — [ServiceConnection documentation](https://developer.android.com/reference/android/content/ServiceConnection.html)
* If spamming is detected, perhaps give a Snackbar saying something like "A command is repeatedly using the bell." with an option to temporarily silence it.

PR #715, opened 2018-06-11
Add limit for the bell (vibration)

l-jonas:
This fixes https://github.com/termux/termux-app/issues/442


fornwall:
Thanks! This has been released in version 0.69 of the app, which should be available as an update shortly.

PR #716, opened 2018-06-11
Add Google Repository

l-jonas:
The Google-Repository is required for compilation because some dependencies were removed from jcenter()

This fixes https://github.com/termux/termux-app/issues/713

PR #717, opened 2018-06-11
Termux:Fix build error

abyss10:
This fixes Error:Could not find common.jar (android.arch.core:common:1.1.0).

Signed-off-by: abyss10 <abyss9979@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,13 +23,9 @@ before_install:
 
 script:
   - ./gradlew testDebugUnitTest
-
-addons:
-  coverity_scan:
-    project:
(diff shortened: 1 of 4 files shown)


l-jonas:
@abyss10 Duplicate of https://github.com/termux/termux-app/pull/716?

abyss10:
It's more generic way to solve the problem!
Yes it does same work!

l-jonas:
@abyss10 It's not more generic. It's exactly the same (but I import google twice which is a bit useless). Could you intend the line with ```jcenter()``` (then I close my PR)?

PR #726, opened 2018-06-18
Sends broadcast on app open to notify addon termux receivers

dkramer95:
This is needed to allow https://github.com/termux/termux-api/pull/185 to work

* Sends a one time broadcast upon app launch to allow the optional TermuxAPI to perform the new checks in the mentioned PR. It has no coupling to the TermuxAPI, other than checking to see if there are any receivers that can listen to the broadcast action, which was needed to support Android Oreo.

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -17,6 +17,7 @@ import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.ServiceConnection;
 import android.content.pm.PackageManager;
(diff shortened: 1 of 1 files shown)


sowntharso:
termux-setup-storage this error creat Directory '~/storage' is going to be wiped. No sto rage contents will be touched.Do you want to continue ? (y/n) nDo you want to continue ? (y/n) n
Leaving intact the contents of '~/storage' directo how to solve this problem tell me

Grimler91:
Better late than never I guess, thanks for the PR!

NEA507 (commented):
Ok

PR #727, opened 2018-06-19
Stop exporting PS1

tomty89:
PS1 is not supposed to be an env var and should be shell-specific.
We will set it appropriately with the init files of the shells.

Companion commit: https://github.com/termux/termux-packages/pull/2554


fornwall:
Thanks!

PR #733, opened 2018-06-25
Add a .gitattributes file

mklein994:
This ensures the correct attributes are used when checking out this repo. For example, *.bat files will use CRLF line endings, and *.sh files will use LF endings, regardless of `core.autocrlf` in your global git config.


fornwall:
Thanks!

Issue #735, opened 2018-06-25
Using termux-setup-storage wipes internal memory


porasjeet:
Android: 7.1.2 (Nougat)
Termux version: 0.62

I've had my device's internal memory wiped when I updated to the latest release and ran "termux-setup-storage" in the terminal. To confirm, I created a few files and ran the command again which again wiped the phone's internal memory.

I've encountered this on my OnePlus 3t and a friend's Redmi Note 5, both running android 7.1.2 (Nougat).

xeffyr:
@porasjeet Just run `termux-setup-storage` twice and got wiped aswell.
@fornwall This is a serious issue and perhaps it is introduced in https://github.com/termux/termux-app/commit/bbd46a763ce98507c5ae275ac315d4ebd9ad410a. Maybe we should revert this temporarily ?

rjt-rockx:
Not sure if the storage has been completely wiped or just unmounted though. 
Also theoretically there should've been a confirmation dialog of sorts before files get deleted.

xeffyr:
@rjt-rockx It is probably a bug. Only `~/storage` with symlinks should be deleted but not entire /sdcard.

fornwall:
That was bad. I'm so sorry for this critical issue!

This has been fixed in version 0.63, which has been submitted to the Play store as an update. The release has been tagged so should be available also on F-Droid soon.

fornwall:
Thanks to everyone involved who reported and helped with understanding and fixing the problem!

jisan0010:
@fornwall 
Android: 7.1.2 (Nougat)
Termux version: 0.62

i ran termux-setup-storage on my android device for access my internal memory.
But when i ran the command the phone storage wiped.
Now what i do?
if  latest version of termux(.063) fixed the problem,then what i do now for my delleted data?
Is there any way to recover my all data?

Thanks

Grimler91:
@jisan0010 Check out one of the numerous guides online, for example https://forum.xda-developers.com/showthread.php?t=2462426. Note that I haven't tried it.

It *might* be possible to restore some of the data. Don't use your phone before you try restoring it as new files will overwrite the deleted data.

indefined:
it wipe my files too,is there something useful to get back some files? the guides above is not work,my phone is sony xperia z5 with android 6.0

xeffyr:
>  to get back some files?

To get back some files, make sure that the following conditions are met:

1. Your phone is **rooted**.
[...]

PR #737, opened 2018-06-25
deleteFolder(): don't treat symlinks as directory

xeffyr:
Should fix https://github.com/termux/termux-app/issues/735


fornwall:
Wow, that was a bad issue.. Thanks a lot!

Quasic:
Wow! That was awful. Thanks for fixing this. I was planning to try to go over it again before it was merged, but I did test the released version several times, and actually used it on an Android 7 phone with pictures that haven't been backed up,  yet. It only deleted the symlinks, as far as I can tell. I don't know why it didn't do more than that, but hopefully some others were unaffected, as well...YIKES!

Issue #738, opened 2018-06-25
termux-setup-storage could still delete files unexpectedly?


tomty89:
So I saw the recent issue and commit that attempted to fix it, and notice that it could still be dangerous if users made a ~/storage a symlink to some directory before using the command, so I did a test:
```
$ mkdir /data/data/com.termux/files/testdir
$ touch /data/data/com.termux/files/testdir/a
$ mkdir /data/data/com.termux/files/testdir/b
$ touch /data/data/com.termux/files/testdir/b/c
$ ln -s /data/data/com.termux/files/testdir $HOME/storage
$ tree /data/data/com.termux/files/testdir
/data/data/com.termux/files/testdir
├── a
└── b
    └── c

1 directory, 2 files
$ termux-setup-storage
$ tree /data/data/com.termux/files/testdir
/data/data/com.termux/files/testdir
└── b
    └── c

1 directory, 1 file
$
```

It isn't as dangerous as I thought, but still you can see that files got deleted unexpectedly.

Tried to write something to do the task myself. Not heavily tested but seems to work. You can take it as a *reference*:
```
import java.io.File;
class recursiveDelete {
    static void deleteNoFollow(File file) {
	if (!file.delete()) {
	    File[] children = file.listFiles();
	    for (File child: children)
		deleteNoFollow(child);
	    file.delete();
	}
    }
    public static void main(String[] args) {
	File blah = new File(args[0]);
        deleteNoFollow(blah);
    }
}
```

P.S. I never get why would someone ever wanted/needed `termux-setup-storage` though. (I always uses the convenient `/sdcard` symlink that the system has for me; probably a bit trickier if one uses a real SD card though.)

xeffyr:
@tomty89 I run commands from your test and got this:
```
$ ls
myfiles
$ mkdir /data/data/com.termux/files/testdir
$ touch /data/data/com.termux/files/testdir/a
$ mkdir /data/data/com.termux/files/testdir/b
$ touch /data/data/com.termux/files/testdir/b/c
$ ln -s /data/data/com.termux/files/testdir $HOME/storage
$ tree /data/data/com.termux/files/testdir
/data/data/com.termux/files/testdir
├── a
└── b
    └── c

1 directory, 2 files
$ termux-setup-storage
$ tree /data/data/com.termux/files/testdir
/data/data/com.termux/files/testdir
├── a
└── b
    └── c

1 directory, 2 files
$ ls -ld storage
[...]

PR #746, opened 2018-06-28
deleteFolder(): check if passed argument is a symlink

xeffyr:
Prevents possible data loss when user replaced directory '~/storage' with
a symlink. Issue: https://github.com/termux/termux-app/issues/738

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxInstaller.java
+++ b/app/src/main/java/com/termux/app/TermuxInstaller.java
@@ -209,18 +209,18 @@ final class TermuxInstaller {
             Arrays.toString(Build.SUPPORTED_ABIS));
     }
 
-    /** Delete a folder and all its content or throw. */
+    /** Delete a folder and all its content or throw. Don't follow symlinks. */
     static void deleteFolder(File fileOrDirectory) throws IOException {
-        File[] children = fileOrDirectory.listFiles();
-        if (children != null) {
-            for (File child : children) {
-                if (child.getCanonicalFile().equals(child.getAbsoluteFile())) {
+        if (fileOrDirectory.getCanonicalPath().equals(fileOrDirectory.getAbsolutePath()) && fileOrDirectory.isDirectory()) {
+            File[] children = fileOrDirectory.listFiles();
+
+            if (children != null) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/termux/app/TermuxInstaller.java
```diff
@@ -209,18 +209,22 @@ private static String determineTermuxArchName() {
...
+    /** Delete a folder and all its content or throw. Don't follow symlinks. */
     static void deleteFolder(File fileOrDirectory) throws IOException {
-        File[] children = fileOrDirectory.listFiles();
-        if (children != null) {
-            for (File child : children) {
-                if (child.getCanonicalFile().equals(child.getAbsoluteFile())) {
-                    deleteFolder(child);
-                } else {
-                    child.delete();
+        if (fileOrDirectory.getCanonicalFile().equals(fileOrDirectory.getAbsoluteFile())) {
```
tomty89: Better use `getCanonicialPath` and `getAbsolutePath` instead.

app/src/main/java/com/termux/app/TermuxInstaller.java
```diff
@@ -209,18 +209,22 @@ private static String determineTermuxArchName() {
...
+            File[] children = fileOrDirectory.listFiles();
+
+            if (children != null) {
+                for (File child : children) {
+                    if (child.getCanonicalFile().equals(child.getAbsoluteFile())) {
```
tomty89: Here you only need to do `isDirectory` as the Canonical vs Absolute test will be perform in the `deleteFolder` call that follows.

tomty89:
Actually it's probably even more proper if you do both the canonical vs absolute test and `isDirectory` in the outer `if` and no condition in the loop.

xeffyr:
@tomty89 now should be ok.

fornwall:
Thanks!

Quasic:
Oh, did it only happen if the main ~/storage was a symlink? *Bangs head* I don't think I tested for that. I had thought of putting a confirmation in the shell script, though, warning that the folder contents would be deleted.

PR #762, opened 2018-07-05
Adaptive icon support

lokesh-krishna:
This adds support for adaptive icons on devices running Android 8.0 or higher while continuing to display the legacy icons on devices running older versions of Android.


flipflop97:
As your adaptive icon also changes the design of the icon (>_ instead of a block) i'd suggest also changing the other icons like the 'legacy' launcher one for the sake of consistency.

trygveaa:
This icon is very similar to the [ConnectBot icon](https://play.google.com/store/apps/details?id=org.connectbot) though. It would probably be better to keep the block, rather than change it to `>_`.

lokesh-krishna:
@trygveaa, The icon is not just similar to ConnecBot. A prompt is used as the icon for terminal emulators across platforms. Each does it slightly differently though, be it colors or the characters used. I've actually used `❯` instead of the commonly used `>` and we also use different colors from the ConnectBot icon and different positioning as well.

The problem with using just a block is that without a border to offer a frame, the icon would just be a single white rectangle in a black background. I can't think of a way to offer a frame due to the various icon masks that might be applied to an adaptive icon. The alternative would be to throw the current icon on to a background but that is a compromise I'd rather not make.

@flipflop97 That's a fair point. I would like to know what @fornwall thinks about this. If we switch the legacy icon to this new one, [Google recommends dropping the round icon](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive) unless the branding of the app relies on a circular shape (and I think it doesn't). So would it be okay to change the legacy icon to the square variant of the new one and drop the round icon?

fornwall:
This is now rolling out in version 0.65 of the app! Let's see how the feedback goes, and then we can proceed to update the old non-adaptive icon as followup work.

Thanks for the work @lokesh-krishna!

lokesh-krishna:
Does the notification icon have to be updated to reflect the new icon?

trygveaa:
#1478 updates the notification icon.

Issue #749, opened 2018-06-30
New keybar (VolumeUp+q)


Chaformbintrano:
It is much less functional and needlessly big. Could you give the old one back or give an option to switch to it?

jayesh-bhoot:
The 2 rows of extra keys is indeed a hindrance. Not a lot of phones have a huge display to compensate for the resulting loss of the screen estate.

In my own case, I cant enable the rows because the full keyboard eats most of my screen estate. But disabling the rows robs me of one-handed use of ctrl and alt keys.

![screenshot_20180702-171049](https://user-images.githubusercontent.com/1915507/42162172-599aa592-7e1b-11e8-98ba-f46af46be715.png)


An ideal, but admittedly more effortful way out could be to eschew the idea of multiple rows in favour of a config where the user could specify the keys and their order of display in a single row. That should cover the needs of the most.

kosovo0275:
Please revert this change. It seems almost careless that this design was implemented, considering there's no real purpose or practicality backing it. Theres also a few insanely obvious and more preferable ways that this change could have been approached.

daemondev:
When using Hacker Keyboard, the extra keys are useless, it should be the user's choice "no extra keys -> old extra keys -> extended extra keys", I already had my environment very well configured, but this change made me lose everything, Now I have to do a downgrade at 0.61.

porasjeet:
I think this should be configurable because I really like the new arrow keys but don't necessarily need the other keys besides TAB, CTRL and ALT.

Or the extra keys view can be changed to a horizontally scrollable list like how QuickEdit implements extra keys.

xeffyr:
> can be changed to a horizontally scrollable list like how QuickEdit

Can't be changed. Scrolling left activates alternative text field which allows text autocorrection/prediction and swype input.

> I think this should be configurable

There is a pull request but it seems that it's author forgot about it: https://github.com/termux/termux-app/pull/650.

porasjeet:
We can shift the text input field to the left side, so swiping right shows the text field and scrolling left scrolls the extra keys.

robertvandeneynde:
Hey guys, the first version of the configurable is almost done in #788, new ideas or critics for configuration are welcome!

neo-expert:
Please bring the two rows solution back again.
For me it was the best solution until now. I used all the keys. Especially the "left and right -arrow keys"
Can we have it configurable?

robertvandeneynde:
[...]

PR #650, opened 2018-04-06
add support of configurable extra keys

neverwin:
People can change extra keys to meet their needs. For example, add one line in termux.properties:
```
extrakeys = [["ESC","CTRL","ALT","TAB","―","/","|"]]
```
That's the default one of Termux.

PR #788, opened 2018-07-29
Updated configurable keys

Grimler91:
Here's an rebased and updated version of @neverwin's configurable keys (https://github.com/termux/termux-app/pull/650). Ping @robertvandeneynde since you had good input in previous PR.

<s>Some stuff remains: I haven't managed to get the property parser to accept arrow keys in termux.properties, if I for example try `extra-keys = [['ESC','/','―','HOME','↑','DEL','-'],['TAB','CTRL','ALT','←','↓','→','PGDN']]` I get:

ss_2
</s>

Which I suppose is a UTF-8/UTF-16 error, and should be fixable.. 
Writing `extra-keys = [['ESC','/','\u2015','HOME','\u2191','DEL','-'],['TAB','CTRL','ALT','\u2190','\u2193','\u2192','PGDN']]` instead works: 

ss_1

I also think that the keys take up too much vertical space, making the rows smaller would be nice. @xqdoo00o I suppose that this is what you are doing in https://github.com/termux/termux-app/pull/759? But that works only for android 5? I didn't notice any big difference when I merged it (https://github.com/Grimler91/termux-app/tree/configurable_keys_smaller)

Also, more importantly, **we/I need to add more validation of the `extrakeys` field!** If, for example, two rows of different lengths are given the app crashes on start-up (with no change of fail-safe session).

--------------------

I've also set the default extra keys to default to the single row we had before, and made the extra keys visible per default. Users who want several rows can configure it.

We should definitely ship a default termux.properties file with all the different options as comments, they are a bit secret as mentioned in the previous PR.

I've also kept the "swipe up from - for |" and "swipe up from / for \" features, though they are less needed with configurable keys.


Grimler91:
Fixes #749.

robertvandeneynde:
Very nice ! Thanks ! I was *just* about to start coding the PR myself. But I prefer to think about the functionalities and summarize them as I did. Giving  _good input_ or _good specs_ and let _experts_ implement the feature :)

robertvandeneynde:
> Also, more importantly, we/I need to add more validation of the extrakeys field! If, for example, two rows of different lengths are given the app crashes on start-up (with no change of fail-safe session).

That answers my question I had in the other PR. So yes, validation **or** default behavior, choose wisely. As goes the saying _having good defaults is important, so is configuration_.

For the case _two lines of different length_ I'd say no error, it means all width = 1 (see my definition of width in the other PR), it's alright not to have a grid, our real hardware keyboards have hex layouts. :smile: 

> I've also set the default extra keys to default to the single row we had before, and made the extra keys visible per default. Users who want several rows can configure it.

Well, choose wisely the good defaults ! I think the people that don't know about configuration (beginners?) prefer the ↑↓ arrows as mandatory for a shell and don't care ←→ are very nice too but less important. Beginners don't know about the configuration until we don't write a good settings menu. The dash "-" and "/" is convenient but easily found on all the keyboards. The "|" is less used by beginners (and again found on all the keyboards). We should focus the default to be on _missing keys_ of classic android keyboard, but that are still widely use. ←→↓↑ are cleary it. And the 4, the Alt is useful for people knowing that emacs basic shortcuts like `Alt+F` exist.

For the one line order, I think "←→↓↑" is a good choice, because the most useful key is at the right. "←↓↑→" is my second advice because it's VIM like.

> JSON

Is it JSON ? What kind of format termux.properties is made of ? I see you use single quote but they are not allowed in JSON.

> We should definitely ship a default termux.properties file with all the different options as comments, they are a bit secret as mentioned in the previous PR.

:+1: And there is the wiki.

> I've also kept the "swipe up from - for |" and "swipe up from / for " features, though they are less needed with configurable keys.

I've thought about this feature, it's only useful if it's explicitly written in the termux.properties. Nobody would think of slide-up if they're not told to. People try to press, sometimes to Long press, never to slide up, that's weird.

However, if someone reads _you can write "slideup": {"key": "|"} and tadaa, there it is !_ it can become very handy to them.

app/src/main/java/com/termux/app/ExtraKeysView.java
```diff
@@ -71,9 +75,15 @@ static void sendKey(View view, String keyName) {
...
                 break;
+            case "↲":
+                keyCode = KeyEvent.KEYCODE_ENTER;
+                break;
             case "―":
                 chars = "-";
                 break;
+            case "-":
+                chars = "-";
+                break;
```
robertvandeneynde: Could be refactored:

```java
case "−":
case "-":
    // if it's "−" or "-", it's the same output
    chars = "-";
    break;
```

Including the comment, not so much people know the syntax.

app/src/main/java/com/termux/app/ExtraKeysView.java
```diff
@@ -71,9 +75,15 @@ static void sendKey(View view, String keyName) {
             case "↓":
                 keyCode = KeyEvent.KEYCODE_DPAD_DOWN;
                 break;
+            case "↲":
```
robertvandeneynde: Add case ENTER too, of course the displayed chars must be "↲" but if people write ENTER in termux.properties it should work as expected.

app/src/main/java/com/termux/app/ExtraKeysView.java
```diff
@@ -59,6 +57,12 @@ static void sendKey(View view, String keyName) {
             case "PGDN":
```
robertvandeneynde: Add aliases, PAGE_DOWN, PAGEDOWN etc. People are likely to have that in their termux.properties.
Same for INSERT, DELETE, BACKSPACE, UP DOWN LEFT RIGHT etc. A short name that is the preffered one, and longer name explicit.

app/src/main/java/com/termux/app/ExtraKeysView.java
```diff
@@ -251,7 +276,7 @@ public void run() {
                                     scheduledExecutor = null;
                                 }
                                 if (longPressCount == 0) {
-                                    if (popupWindow != null && "―/".contains(buttonText)) {
+                                    if (popupWindow != null && "―/-".contains(buttonText)) {
```
robertvandeneynde: In the case we have buttonText.length != 1, I would go for:
```java
Arrays.asList("―", "/", "-").contains(yourValue)
```

app/src/main/java/com/termux/app/ExtraKeysView.java
robertvandeneynde: What if the guy wrote "\u2015" in the termux properties? We should remove that "case", if he wrote.

The default config should contains the ascii dash (U+002D - HYPHEN-MINUS), but should be *displayed* as "−" (U+2212 − MINUS SIGN) in the shell. See my `display` parameter in the PR.

robertvandeneynde:
After reading the source code `termux.properties` is simply a file that must be in `$HOME dir`. Or in `$HOME/.config/termux` of type `java.util.Properties`. Which makes sense.

However, also considering JSON has some advantages.

This `.properties` yields an error:

```
extrakeys = [
    ["←", "→", "↓", "↑"],
    ["CTRL", "ALT", "DEL", "TAB"]
]
```

Because new lines must be escaped:

```
extrakeys = [\
    ["←", "→", "↓", "↑"],\
    ["CTRL", "ALT", "DEL", "TAB"]\
]
```

Which is annoying for that case.

The same information is available in JSON:
```
{"extrakeys": [
    ["←", "→", "↓", "↑"],
    ["CTRL", "ALT", "DEL", "TAB"]
]}
```

I suggest that the code looks up for `termux_config.json` in both directories, and termux.properties in both directories. Merge the keys but yields a Warning if both files have keys in common, in that case, take the one in `.properties`.

The only advantage is that one doesn't have to put quotes around all the keys and values for simple cases like we generally have. Another advantage is that Comments are allowed. And that new lines don't have to be written "\n". However *everybody does know Json*, less people know java.util.properties but the syntax [is so intuitive](https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html#load(java.io.Reader)) that's alright. Both syntaxes allow \u1234 syntax, neither does allow \U12345678 syntax to input emojis by unicode.

One option would be to be more lenient on the json file:
- Removes comments with '#', ie. lines BEGINNING with '#'. As in .properties, it is not allowed to have some text THEN a comment, a comment must be on one line.
- If the file is not parsable as JSON object, add "{}" around it, if it isn't still possible, throw an error.

We could switch to YAML, but less people know about it. JSON is simple but more verbose.

Grimler91:
> Giving good input or good specs and let experts implement the feature. 

I'm definitely no java expert though, but hey, thanks.

So about the termux.properties file. As you've already said it's just a normal utf-8 file, which is parsed by [java.util.Properties](https://developer.android.com/reference/java/util/Properties) into a properties object. The extrakeys entry is read into a string which is later parsed by the [JSONArray](https://developer.android.com/reference/org/json/JSONArray) library into a json object. So the string must have a syntax that JSONArray can handle.

' instead of " works just fine.

The reason I mentioned utf-16 is because java seem to use [utf16 for strings](https://stackoverflow.com/questions/20966802/utf-16-character-encoding-of-java), so maybe java.util.properties reads the utf-8 file as utf-16 and causes the problems. I don't really know, have to do more testing.

Anyways, good that the development is on it's way!

Grimler91:
> After reading the source code termux.properties is simply a file that must be in $HOME dir. Or in $HOME/.config/termux

Actually `$HOME/.termux/`or `$HOME/.config/termux`. I managed to find the wiki entry on it: https://wiki.termux.com/index.php?title=Terminal_Settings

robertvandeneynde:
I might fork the repo and add commits to the pull request, if I find how to do do that, do I have to do *another* PR? Or can I _join_ in? I know git and branches but pretty new to github.

> I'm definitely no java expert though, but hey, thanks.

Cmon, a little bragging doesn't hurt, and _expert_ doesn't have a good definition (this word makes me think about [this humorous video](https://www.youtube.com/watch?v=BKorP55Aqvg)).

> I managed to find the wiki entry for the configuration file location

And in [the source code](https://github.com/Grimler91/termux-app/blob/4d4508509971fd870be93522f044da6f6d2d7e53/app/src/main/java/com/termux/app/TermuxPreferences.java#L112).

> UTF-16

Indeed, java does use UTF-16, `sizeof(char)` is 2 bytes. Butin your screenshot, tou seem to have 3 characters, which is utf-8 for the arrow keys. So I'd say it's utf-8 but should be utf-16. Thaaat's why I suggest the first line to be `# coding: utf-8` to specify the encoding and 

And the default keys should be ascii. The user can write strings like "LEFT", "RIGHT", "UP", "DOWN" for Arrow keys and control characters. That would be Displayed as ←↑→↓ (↲ Enter) (↹ Tab). And if the user doesn't want that he can use the "display" property I was thinking of: 

```json
extrakeys = [[{"key":"TAB", "display":"TAB"}, {"key":"CTRL", "display":"^"}]]
```

Or we can add a new option:

```json
extrakeys = [["TAB", "CTRL"]]
extrakeys.controlcharstext = ["TAB", "ENTER"]
```

This would mean, these keys shall be rendered as "TAB" and "ENTER" and not their default (↲ Enter, ↹ Tab). but I prefer

Grimler91:
> I might fork the repo and add commits to the pull request, if I find how to do do that, do I have to do another PR? Or can I join in? I know git and branches but pretty new to github.

Sounds great! I probably won't work on this again for another day or two.
If you open a PR to my branch (i.e. go to https://github.com/grimler91/termux-app/tree/configurable_keys and press "New pull request" there) then your commits are added to this pull request.

> Indeed, java does use UTF-16, sizeof(char) is 2 bytes. Butin your screenshot, tou seem to have 3 characters, which is utf-8 for the arrow keys.

Yupp, I checked, it's 3 characters. So we need to decode and encode it as utf-16 instead in ExtraKeys.java I guess. I don't see how # coding: utf-8 would help (and I tried adding it to termux.properties, no difference).

You have lots of good ideas that could probably be split into several PRs, which this first one aiming at making it possible to configure keys (without breaking termux easily) and then enhance it by increasing user-friendliness and adding more options and possibilities.

robertvandeneynde:
Okep, I'll pull request You then.

Indeed `java.utiol.Properties` sees `# coding: utf-8` as a comment, but that would be something clever _we_ would implement so that the user can specify the encoding he used. But because it's stupid to reinvent the wheel we'll take the conventio of emacs, the same as [the one from Python](https://www.python.org/dev/peps/pep-0263/), we'd use the regex `^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+)`.

> You have lots of good ideas that could probably be split into several PRs

Yep of course. The PR serves also as a conversation, We Could talk about the ideas in a _Issue_ though, then go for multiple PR tagging that _issue_.

However, the first version of the implementation should be thought, not just _this sounds good to me, here's my PR_ and fornwall would be like _This is great, I'll push it on the store_ aaand the new user will complain _Why do you put two bars and no configurable keys_.

Okep, back to coding, however I currently can't test the app myself (pretty new to android native app), how do I do that? :smile: At first I'll do the Java-Only parts.

robertvandeneynde:
For the encoding issue, in `TermuxPreferences.java`:

```java
try (InputStreamReader in = new InputStreamReader(propsFile, "utf-8")) {
    props.load(in);
}
```

Instead of  `try (FileInputStream in = new FileInputStream(propsFile))`. I'm including that in a commit if it solves the problem, can you test it?

The `java.util.Properties` class has two `load` functions [seen here](https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html#load(java.io.InputStream)). The old code used a byte oriented class and we want a character oriented one, the old-code assumed latin1 encoding which is not a good default anymore.

robertvandeneynde:
Alright, I just added two commits, but I didn't test it because I haven't installed the android sdk yet. If you could compile and test for me that'd be good.

- Default file uses "LEFT" instead of "←", that's more easy for people to input and doesn't have encoding problems.
- Refactoring of code using Ctrl, Alt, Fn. Many copy paste, I refactored that.

Grimler91:
I'll test your changes and give you a report :) 

About building the app, you can install the sdk (and ndk) by running scripts/setup-android-sdk.sh in [termux-packages](https://github.com/termux/termux-packages/blob/master/scripts/setup-android-sdk.sh). You also need to have java installed (obviously). 

After that, you can build the app pretty easily from the command line (without the need of installing android-studio). I have cloned all the termux apps to a common folder and from that folder I run this script:
```bash
#!/bin/sh                                                                                                                                                                                                                                                                                                                   

### passwords for keystore                                                                                                                                                                                                                                                                                                  
# keystore_pw=$(pass dev/keystore)                                                                                                                                                                                                                                                                                          
# key_pw=$(pass dev/keystore-termux)                                                                                                                                                                                                                                                                                        

### your keystore file                                                                                                                                                                                                                                                                                                      
# keystore_file=NAME.keystore                                                                                                                                                                                                                                                                                               

export ANDROID_HOME=~/lib/android-sdk
export ANDROID_NDK_HOME=~/lib/android-ndk

if [ -z $1 ]; then
    apps="app api float tasker styling boot"
else
    apps="$1"
fi

mkdir -p apps

for app in $apps; do
    if [ ! -d "termux-$app" ]; then
        git clone git@github.com:termux/termux-$app
    fi

    cd termux-$app
    ### Name the apk after the branch
    branch=$(git rev-parse --abbrev-ref HEAD) 

    ./gradlew clean
    ### Build with gradle:                                                                                                                                                                                                                                                                                                    
    ./gradlew assemblerelease
    ### Align the app (not strictly necessary)                                                                                                                                                                                                                                                                                
    zipalign -v -p 4 app/build/outputs/apk/release/app-release-{unsigned,aligned}.apk
    ### Sign the app:                                                                                                                                                                                                                                                                                                      
    # jarsigner -sigalg SHA1withRSA -digestalg SHA1 \                                                                                                                          
    #         -keystore ~/.android/${keystore_file} \                                                                                                                              
    #         -storepass ${keystore_pw} \                                                                                                               
    #         -keypass ${key_pw} \                                                                                                                                               
    #         app/build/outputs/apk/release/app-release-aligned.apk termux_app
    #### Move the finished apk to folder "apps"
    cp app/build/outputs/apk/release/app-release-aligned.apk ../apps/termux-$app-$branch.apk
    cd ..
done
```

If you want to sign the app as well you need to [generate a signing key](https://stackoverflow.com/questions/3997748/how-can-i-create-a-keystore) and then un-comment those things in the script.

Remember that you can't have several termux apps installed if they are signed with different keys, i.e. you need to uninstall termux-{app,api,...} fully before installing your own version, so remember to backup $HOME if you do this.

Grimler91:
@robertvandeneynde I'll continue the discussion here so that it is a bit easier to follow.

> However, the first version of the implementation should be thought, not just this sounds good to me, here's my PR and fornwall would be like This is great, I'll push it on the store aaand the new user will complain Why do you put two bars and no configurable keys.

Yeah, I guess the problem is that not too many people build the apps themselves and test PRs, so feedback is limited before stuff is merged. The higher the quality before the merge the better of course :) 

> With the current code, it does happen, the String[][] buttons is a array of array of non null Strings, but the subarrays are not all the same size, but a good question would be what happens when the first line has less keys than the second one. To which I alread have a good answer in my previous posts (it should stretch the layout).
> 
> The old code did final String buttonText = (buttons[row][col] == null ? " " : buttons[row][col]); but it's useless because if not all rows have same length, IndexError is thrown.

Stretch the layout sounds like the most elegant solution. 
I don't see why an IndexError would happen as long as we iterate over it properly, an array of arrays with different row lengths [seem to be allowed in java](https://stackoverflow.com/questions/32926723/array-with-different-row-lengths#32926748)

> So do I have to create another PR or will you do git merge? :)
> 
> I'm trying to do another PR a but gihub doesn't offer from robertvandeneynde termux-app configurable_keys into Grimler91 termux-app configurable_keys, it only proposes to compare from termux repo , even if I tick compare against fork thinking , I can't remember how I managed to create This PR :D

Opening a new PR is perhaps the easiest because github can handle rebasing if necessary, though in this case I could have easily cherry-picked/merged that single commit. 
I've also noticed that github is a bit stupid when it comes to opening pull requests to a fork (or a fork of a fork of a...), easiest seem to be to go to the repo&branch you want to open against and then press "new pull request", or use a link like you did :+1:  

So now that you've implemented most special keys as LEFT -> ←, TAB -> ↹ etc, do you still think it's important to change the props.load() function. No one will (I hope) copy paste unicode characters from the internet into termux.properties but stick to the text definitions, if they are clear and we provide many examples.

robertvandeneynde:
> I don't see why an IndexError would happen as long as we iterate over it properly, an array of arrays with different row lengths seem to be allowed in java.

Yes, they are allowed, yes `for(int i = 0; i < buttons.length; i++) for(int j = 0; j < buttons[i].length; j++) buttons[i][j]` is the correct way to iteratate.

But the code does `cols = maximumLength(buttons)` and the `class ExtraKeysView extends GridLayout`, so I don't know if the layout is streched by default, I think currently what would happen if lets say we have `[['LEFT', 'RIGHT', 'UP', 'DOWN'], ['CTRL, TAB']]`, the CTRL and TAB would occupy each 25% of the space, the right-most 50% and just two "spacebar" buttons (ie. nothing displayed, sends a " ") where 50% seems better. I know nothing about Android layouts. :woman_shrugging: 

However, as you said, let's create different PR from different possibilities, and discuss in here, oh and watch out the changes done on `configurable_keys_smaller` with david xqdoo00o. 

> So now that you've implemented most special keys as LEFT -> ←, TAB -> ↹ etc, do you still think it's important to change the props.load() function

No, if people copy paste, it must be utf-8. However, adding the regex is very easy and very classic in vim/emacs, this can be done (by me, when I feel like it) but not documented on the wiki.

robertvandeneynde:
So, next features (in another PR):

- Streching layout to allow different row lengths `[['LEFT', 'RIGHT', 'UP', 'DOWN'], ['CTRL, TAB']]`.
- To be consistent, the termux.properties settings should be called `extra-keys` and not `extrakeys` (or I can write a code that would automatically allow `extrakeys` `extra-keys` `extra_keys` for all keys.
-  Adding a simple parameter to choose when the arrows are used, based on the `Map` in my source code, I see two parameters:
- `extrakeys.character-display` that can be one of `default, lots-of-arrows, arrows-only, full-iso, no-replacement`, each one has a Map in the code. `no-replacement` would use the Empty map.
- `extrakeys.arrowtriangles` boolean, that would use ◀ ▶ ▲ ▼ instead of ←→↑↓.

And for that PR, I'd say to Test it :D

Grimler91:
> But the code does **cols = maximumLength(buttons)** and the class ExtraKeysView extends GridLayout, so I don't know if the layout is streched by default

Ah, there it is!
So lets do as you suggested:
```diff
diff --git a/app/src/main/java/com/termux/app/ExtraKeysView.java b/app/src/main/java/com/termux/app/ExtraKeysView.java
index a4073ae..4e8079e 100644
--- a/app/src/main/java/com/termux/app/ExtraKeysView.java
+++ b/app/src/main/java/com/termux/app/ExtraKeysView.java
@@ -326,7 +326,7 @@ public final class ExtraKeysView extends GridLayout {
         setColumnCount(cols);
 
         for (int row = 0; row < rows; row++) {
-            for (int col = 0; col < cols; col++) {
+            for (int col = 0; col < buttons[row].length; col++) {
                 final String buttonText = buttons[row][col];
                 
                 Button button;
```
We then get deadspace (unclickable) instead, which is at least better than an app crash but not as good as stretched keys.
ss

robertvandeneynde:
Oh and don't forget to:

```diff
- JSONArray arr = new JSONArray(props.getProperty("extrakeys", "[[\"ESC\",\"CTRL\",\"ALT\",\"TAB\",\"―\",\"/\",\"|\"]]"));
+ JSONArray arr = new JSONArray(props.getProperty("extrakeys", "[[\"ESC\",\"CTRL\",\"ALT\",\"TAB\",\"-\",\"/\",\"|\"]]"));
```

For the default one line bar.

- We should add a default `termux.properties` that has the one-line version and two-lines version commented.

robertvandeneynde:
About JSON version, do we use the latest one? Here is the [link](https://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.json%22%20AND%20a%3A%22json%22) on Maven, here is the [link](https://github.com/stleary/JSON-java) on github.

From [the doc](http://stleary.github.io/JSON-java/index.html), single quotes are normal, and often don't need to even used ! 

The texts produced by the toString methods strictly conform to the JSON syntax rules. The constructors are more forgiving in the texts they will accept:

> - An extra , (comma) may appear just before the closing brace.
> - Strings may be quoted with ' (single quote).
> - Strings do not need to be quoted at all if they do not begin with a quote or single quote, and if they do not contain leading or trailing spaces, and if they do not contain any of these characters: { } [ ] / \ : , # and if they do not look like numbers and if they are not the reserved words true, false, or null.

So the default can be written:

```
extra-keys = [[ESC, CTRL, ALT, TAB, "-", "/", "|"]]
```

Not sure about the backslash policy, what happens when writing `"\"`?

Grimler91:
I've invited you to collaborate (you should have an email from github) to my grimler91/termux-app fork.

I merged in changes from master and then cherry-picked your commit. With cherry-pick I mean `git fetch robert configurable_keys; git cherry-pick {commit-hash}`. It gives a bit cleaner history if the branch histories aeen't identical (no merge needed). 

> So the default can be written:
> 
> `extra-keys = [[ESC, CTRL, ALT, TAB, "-", "/", "|"]]`

I think that look more confusing, better to quote all strings then, even if it isn't strictly needed.

Wait, "\\\\" is for _one_ backslash? Wow... An alias would indeed be nice then!

robertvandeneynde:
Yes, 4 backslashs for a single backslash, because Both Json and Properties use `\\` for a backslash.
That's why BACKSLASH is the most useful ascii alias, QUOTE and APOSTROPHE, seems good too. And if people what to be sure about escapes, we should consider `~/.termux/termux_properties.json`(or is the `.` a better idea than `_`?)

Adding other ascii aliases seems a bit stupid (but we had to, I'd pick all unicode names and [aliases](http://www.unicode.org/Public/11.0.0/ucd/NameAliases.txt) or all ascii names in unix `/usr/share/X11/xkb/symbols` aka `xkb_symbols` because _lots_ of people know them), but I prefer people write "-", "/" better than...

> HYPHEN-MINUS (uniname), SOLIDUS (uniname), REVERSE SOLIDUS (uniname), REVERSE-SOLIDUS (uniname.replace(' ', '-')), BACKSLASH (X11 name), TILDE (uniname), ASCIITILDE (X11 name), GRAVE ACCENT (uniname), GRAVE (X11 name), BACKTICK (slang), QUOTATION MARK (uniname), APOSTROPHE (uniname).

However, aliases for controls characters is an idea, \u0003 <EOF> END OF TEXT for example, or \u0007 <BELL> BELL. We already have U+0008 \x08 <BS> BACKSPACE after all. I'm, I named BKSP but the [small unicode alias](https://www.unicode.org/Public/8.0.0/ucd/NameAliases.txt) is BS :thinking: .

robertvandeneynde:
@fornwall @Grimler91 This PR is ready to be merged for new version!

app/src/main/java/com/termux/app/ExtraKeysView.java
```diff
@@ -147,45 +164,183 @@ void popup(View view, String text) {
...
+        // no alias for HOME
+        // no alias for END
+        
+        put("PAGEDOWN", "PGDN");
+        put("PAGE_DOWN", "PGDN");
+        put("PAGE_DOWN", "PGDN");
+        put("PAGE-DOWN", "PGDN");
+        
+        put("DELETE", "DEL");
+        put("BACKSPACE", "BKSP");
```
robertvandeneynde:
```java
put("BACKSLASH", "\\");
put("QUOTE", "\"");
put("APOSTROPHE", "'");
```

robertvandeneynde:
In [your comment](https://github.com/termux/termux-app/pull/788#issuecomment-409737005) below the screenshot, write that you wrote `extra-keys = [["ESC", "/", "ENTER", "HOME", "UP",  "DELETE", "-"], ["TAB", "CTRL", "ALT", "LEFT", "DOWN"]]`. People might want to read the PR to see the changes, this gives a good idea of the implemented features when just reading the PR. Also remove your [first](https://github.com/termux/termux-app/pull/788#issue-204651304) screenshot where you had an encoding problem.

robertvandeneynde:
Okay I can't push on the branch, here are infos:

```bash
16:37 ~/termux-app $ git push grim configurable_keys
Password for 'https://robertvandeneynde@github.com': 
To https://github.com/Grimler91/termux-app
 ! [rejected]        configurable_keys -> configurable_keys (non-fast-forward)
error: failed to push some refs to 'https://robertvandeneynde@github.com/Grimler91/termux-app'
hint: Updates were rejected because a pushed branch tip is behind its remote
hint: counterpart. Check out this branch and integrate the remote changes
hint: (e.g. 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

16:37 ~/termux-app $ git pull grim configurable_keys
From https://github.com/Grimler91/termux-app
 * branch            configurable_keys -> FETCH_HEAD
Already up-to-date.

16:37 ~/termux-app $ git log -n 2
commit 7bd18c420c469cbe3fb864df9d4710642511d86a (HEAD -> grim_configurable_keys)
Author: Robert Vanden Eynde <robertvandeneynde@hotmail.com>
Date:   Thu Aug 2 16:34:43 2018 +0200

    Add aliases for BACKSLASH QUOTE and APOSTROPHE

commit f0120390ab2ec59764ac98f3265e26facda722e8 (grim/configurable_keys)
Author: Henrik Grimler <Henrik@Grimler.se>
Date:   Thu Aug 2 08:46:03 2018 +0200

    ExtraKeys: use ' instead of \" in default key string

```

robertvandeneynde:
@fornwall this PR is ready to be merged in!

A [screenshot](https://github.com/termux/termux-app/pull/788#issuecomment-409737005)  explain the thing when `extra-keys = [["ESC", "/", "ENTER", "HOME", "UP", "DELETE", "-"], ["TAB", "CTRL", "ALT", "LEFT", "DOWN"]]` in set in `termux.properties`.

Release notes include:

- Adding `extra-keys` attribute in `termux.properties` to customize layout. The old one line bar can be used with `extra-keys = [["ESC", "CTRL", "ALT", "TAB", "-", "/", "|"]]` or the new two lines bar can be used with `extra-keys = [["ESC", "/", "-", "HOME", "UP", "END", "PGUP"], ["TAB", "CTRL", "ALT", "LEFT", "DOWN", "RIGHT"]]`. All the possible keys can be seen in the source code in the map `keyCodesForString` in `ExtraKeysView.java`, aliases like RETURN for ENTER can be seen in the map `controlCharsAliases`.

tomty89:
Is it possible to use `/data/data/com.termux/shared_prefs/com.termux_preferences.xml` for this instead?

Grimler91:
@tomty89 what's the benefit of that? Feels like it would make the file much harder to find.

tomty89:
It's not really much harder, but yeah it might feel so (and out-of-comfort-zone).

Well there are couple reasons (which might be a bit personal and may or may not be valid):
1. Not a lover of dotfiles, especially when it's not even related to the shell or the "terminal", but rather the "app"
2. The show extra key attribute makes use of that. Feels more consistent if they are in the same place. (Though you may argue that that can be moved instead...)
3. Perhaps xml and/or the SharedPreference interface would makes it even more neat? Only a hunch though.

robertvandeneynde:
The app currently uses SharedPreference interface for some settings (I've seen it somewhere in the source code), I don't know for what. But it is but it _seems_ to be a classic way to store settings in android. We could add that in another small PR, the idea would be also to have a classic "settings" menu when some useful settings will be changed in a gui, the dotfile would override. We currently look for multiple locations and that's a good thing.

About xml, we could also add a simple xml interface when we'll add JSON (currently only java.util.Properties type is allowed), if people likes xml more than JSON or properties, why not. However, xml is verbose, JSON less, properties is a good tradeoff.

On Ubuntu, that's standard to keep app infos in ~/.config, because the $HOME dir is related to the user whereas /etc is about root. When people back their stuff, they backup $HOME.

About the directory /data/data/com.termux/, if it's standard on Android app (I'm no Android expert), we'll add the location.

About the filename, com.termux_preferences, is that standard? Or just your first idea?

tomty89:
The file is created by the SharedPreference interface. I'm sure you have it in your Termux as well. That's a reason I think it is preferred, as it seems to be the orthodox one where you don't even need to deal with the path yourself.

What I consider bad is that Termux has been making use of two different interfaces (`android.content.SharedPreferences` and `java.util.Properties`) in TermuxPreferences for different stuff. I am not sure there's any good reason doing so. That's why I think we should stop using one of them and ditch that completely one day.

*Semantically* speaking I don't think we should put *app settings* (i.e. settings that has nothing to do with programs we uses in Termux or terminal emulation) anywhere under `/data/data/com.termux/files` but somewhere else (even directly) under `/data/data/com.termux`. And certainly we don't need to follow XDG specs (where `~/.config` comes from) for that.

robertvandeneynde:
Ok I see, the shared_prefs file currently only contains "show_extra_keys". ie. Preferences set by the gui, that are simple int, boolean or strings, and changed by the gui. Using SharedPreferences only modifies that file but we could also read the XML.

I'd say that if we add gui to change the extra-keys settings, it would go in that xml also, and we'd look for that.

The current advantage of `~/.config` is that users can `cd ~/.config` in the terminal, whereas cd /data/data/ but we could add a symlink `ln -s /data/data/com.termux/shared_prefs/com.termux_preferences.xml ~/.termux_preferences.xml` or similar.

And again, the advantage of java.util.Properties is that it's very brief and not verbose. But I agree this does the trick too:

```xml
<map>
    <string name="extra-keys">[
        ["CTRL", "ALT", "DELETE"]
    ]</string>
</map>
```

Xml does ask more questions, `<string>` vs `<json>`? Don't use json and use xml?

```xml
<!-- very verbose -->
<rows name="extra-keys">
    <row>
            <key>CTRL</key>
            <key>ALT</key>
            <key>DELETE</key>
    </row>
</rows>
```

What is the point of view of @fornwall ? Why did you choose ~/.config and java.util.Properties instead of SharedPreferences for the keys?

However I think this can be in a future PR, and using something standard for android is nice. As the [dev guide](https://developer.android.com/guide/topics/data/data-storage#pref) says, SharePreferences is useful for storing simple key/value data, which is what our `.properties` does.

robertvandeneynde:
However, I think the new default must be reconsidered:
- people like one line
- people like arrow keys, especially UP and DOWN (it's a terminal!)
- the default width seems to be 7
- the - / | are useful but already there on all keyboards.

I suggest the new default being:

```
extra-keys = [["ESC", "CTRL", "ALT", "TAB", "/", "DOWN", "UP"]]
```

I don't think there is enough space to have - / | and ↓↑.

And of course, we'll find later a way to explain to people how to write the `.properties` file on the wiki and/or doing a proper `Preferences` panel easily accessible.

fornwall:
@robertvandeneynde Thanks a lot, and sorry for the delay!

I'll try it out during the coming couple of days and make a release shortly.

robertvandeneynde:
I figured out you were on vacation :)

I'm still wondering if the default should be on one line with arrow keys like in my last commit or on two lines, and if show_extra_keys should be displayed by default (I think yes, majority of people don't know about changing keyboards or can't plug a real keyboard).

And also, because the left menu cannot be found easily, add an entry "show side menu" to the long press menu. And a text "long press on the screen to see the options" in the welcoming message in the terminal.

For the future, I think some accessibility are useful, I'm seeing a classic "settings" (Preferences) panel. Such that basic settings can be changed in a classic way, without editing dot files.

robertvandeneynde:
@fornwall ? It's been 2 months when you said "soon" :p Isn't there someone else that can test and publish a new version on the store ?

Issue #516, opened 2017-12-28
A Little Padding?


lokesh-krishna:
Text is pressed right against the edges of the screen presently. It would be really nice if padding were added. The most minimal amount possible but just enough to avoid this:

![screenshot_20171228-083845](https://user-images.githubusercontent.com/20235646/34399135-733ca770-eb7c-11e7-87eb-cb0aa5ccc143.png)

Quasic:
Wow, my device is black, so I never noticed. How much padding do you think is necessary? I value screen space, but black isn't the only device color... I wonder if it could be made a setting.

sdrausty:
Some devices are black.  A number are white.  A few are in many colors.  This topic is superfulous as every screen regardless of device color has a black border embedded in the glass surrounding the LED screen.  Some of these borders are wider than others, but this doesn't make any difference since it is black.  If device color should not appeal with Termux's default screen setting for the user, an addon is available. It is called Termux styling!  More information about your solution for this topic through the Termux Styling Addon is at this https://termux.com/add-on-styling.html link;  Enjoy.

lokesh-krishna:
The request is not related to the color of the device at all. It is purely about the content being pressed right to the edges of the screen without any padding, which in my opinion just doesn't look very polished. It even looks almost like text is getting cut off at times. 

A simple 1px wide border on all sides should suffice, not compromising on screen real estate but still offering a better appearance.

sdrausty:
> has a black border embedded in the glass surrounding the LED screen. Some of these borders are wider than others, but this doesn't make any difference since it is black. 

Hence, every screen has built in black padding which consumes no LED screenspace.

Grimler91:
@sdrausty @lokesh-krishna Devices with curved screens/infinity edges/whatever you want to call it, that black border is pretty much none-existing.

It is quite hard to mark text all the way to the edge, I agree that padding, or an option to add padding, would be nice.

sdrausty:
@Grimler91 @lokesh-krishna 
> It is quite hard to mark text all the way to the edge,

It is impossible to mark text all the way to the left edge of the screen.  Dragging the visual mode cursor one line up at the extreme right side of the screen accomplishes marking text all the way to the left border of the screen.  Adding 1px of padding will not change this text marking event whatsoever.  

> black border is pretty much none-existing
[...]

PR #792, opened 2018-07-31
add some padding for TerminalView

xeffyr:
Should fix https://github.com/termux/termux-app/issues/516

I guess padding of 3dp should be enough.


fornwall:
Thanks!

PR #794, opened 2018-07-31
Alternative patch to fix fn-up fn-down pgup/pgdown

### LINKED ISSUE #731: Home and End on a hardware keyboard do not seem to work, likewise SHIFT+PAGEUP/DOWN

zevv:
- in onKeyDown() ignore keys when the Fn key is pressed
- in onKeyDown() return false when the key is not handled.

The above two changes fix the handling of Fn-key combo's on devices with
a physical keyboard, allowing the android system defined fallbacks from
`/system/usr/keychars/Generic.kcm` to be properly handled.

Fixes #731.

Original diagnosis and fix by Konehaltia.


fornwall:
Thanks! This is now available in v0.65 of the app, which has been released and is rolling out as an update.

PR #797, opened 2018-08-03
Nord color variant adaptive icon

lokesh-krishna:
Replaces the foreground and background layers with Nord color variants.


fornwall:
Thanks!

PR #811, opened 2018-08-17
set environment variable 'BOOTCLASSPATH'

xeffyr:
Without this variable, such tools as 'ecj' or 'dx' may crash (just print message 'Aborted') on some devices.


xeffyr:
Closing as required only on specified devices with low Android versions (android 5 and maybe 6) only.

Issue #482, opened 2017-11-17
[Feature Request]Termux as home app


rubikitch:
I use Termux almost all time.
So I want Termux as home app.

trusktr:
That'd be interesting. Can we launch other apps from Termux?

Grimler91:
It looks like https://github.com/termux/termux-app/issues/206 is asking for the same thing (correct me if I am wrong).

trusktr:
Looks like it is, and the solution may near!

Issue #206, opened 2016-12-04
launcher mode support?


thewisenerd:
is it _possible_ to have termux as a "launcher" screen?

this angle has been looked at by TUI-ConsoleLauncher ( https://github.com/Andre1299/TUI-ConsoleLauncher ), and I'd think that it would be nice to have an "optional" launcher mode.

a couple of user scripts using the Termux:API add-on and auto complete (volup + T) will make a _good_ launcher.

as for the "launching" apps part, can be either done as a addon extension to Termux:API or _*cough*_ su am start com.package/.Activity _*cough*_ some userscripts.

xelxebar:
As a T-UI user myself, I've been mulling over this idea for a while.

Turning Termux into a launcher, it turns out, is just a matter of adding a line to the manifest:

```
diff --git a/app/src/main/AndroidManifest.xml b/app/src/main/AndroidManifest.xml
index aacc244..35cc0d4 100644
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -84,6 +84,7 @@
             <intent-filter>
                 <action android:name="android.intent.action.MAIN"/>
                 <category android:name="android.intent.category.IOT_LAUNCHER"/>
+                <category android:name="android.intent.category.HOME"/>
                 <category android:name="android.intent.category.DEFAULT"/>
             </intent-filter>
         </activity-alias>
```

So the actual work would be the "activity launching" stuff. Evidently, we can launch activities without root by passing the `--user 0` flag to `am start`: *e.g.*

    $ am start -n org.fdroid.fdroid/.FDroid --user 0

This works pretty well, and In theory we should be able to filter through the list of activities with

    $ pm list packages

However, I'm only able to successfully do that via adb. From Termux, it's failing with `android.os.DeadObjectException` for some reason.

Anyway, we could probably make all this usable with a shell script or two and some bash completion functions.

thewisenerd:
@xelxebar hi, and thanks for trying this out

`pm list packages` is crashing because of a selinux denial.

```
09-10 06:09:12.956  5209  5209 I main    : type=1400 audit(0.0:78): avc: denied { read write } for path="/dev/pts/0" dev="devpts" ino=3 scontext=u:r:system_server:s0 tcontext=u:object_r:untrusted_app_devpts:s0:c512,c768 tclass=chr_file permissive=1
```

i set my selinux to permissive and it works :smile: 

need to figure out a way to fix this :crossed_fingers:

[...]

PR #880, opened 2018-10-18
Support setting Termux as home launcher

xelxebar:
## Overview

This addresses #206 (along with #482 and #880), but in a more canonical way than I suggested in the comments there.

As for launching applications, we can use `am`

```
am start --user 0 <intent>
```

Personally, I wrap these in aliases in my bashrc and it works quite well:

```
app_launch() {
        am start --user 0 "${1}"  >/dev/null
}

alias settings='app_launch "comm.android.settings/.Settings"'
```

## Thoughts

Obviously, my method above requires manual curation. The most annying part is getting the intent uri for newly installed apps, but in #206 @thewisenerd demonstrates one method of easing that hurdle.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -37,6 +37,7 @@
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
+                <category android:name="android.intent.category.HOME" />
             </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />


myfreess:
https://www.coolapk.com/apk/com.tumuyan.fixedplay
A Chinese app,named AnyLauncher，can set termux as Android Launcher。 @xelxebar

amsitlab:
I've [TermuxLauncher](https://github.com/amsitlab/termuxlauncher) , currently just generate shortcuts for valid bash script. You can launch app via command `launch [app name]` like `launch whatsapp` , if appname have [space] you can replace with dash "-" like `launch google-play-store` 
Best regards.

pulketo:
I've programmed a php script to execute those launcher commands with just a part of app's names
"x wh" to launch whatsapp or
"x cont" to launch contacts
https://github.com/pulketo/termux-cute-xcute

Hhamadatell2019:
pkg

justworks007:
Is there any hope that this feature will be merged?

xeffyr:
@justworks007 No, this won't be merged. Don't forget that Termux is a terminal emulator and that's its only purpose.

I'm strongly against overloading it by unnecessary permissions, intents, etc.

PR #886, opened 2018-10-23
Fix Termux crash on malformed config file

koppa:
When having an malformed json input in the extra-keys property, the termux-app crashes later with an NullpointerException due to an uninitialized mExtraKeys variable.

This fix improves the parsing of the properties file. In case of an invalid "extra-keys" property, the variable will be initialized with length 0.

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxPreferences.java
+++ b/app/src/main/java/com/termux/app/TermuxPreferences.java
@@ -7,18 +7,19 @@ import android.support.annotation.IntDef;
 import android.util.Log;
 import android.util.TypedValue;
 import android.widget.Toast;
-
 import com.termux.terminal.TerminalSession;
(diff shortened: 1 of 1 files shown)


Grimler91:
Great, we most certainly need more verification :)

I also see that you've added an alternative config path (just mentioning it since it's worth mentioning)

koppa:
> I also see that you've added an alternative config path (just mentioning it since it's worth mentioning)

Nope, that wasn't me, this is already in master.

I just splitted the try-catch clause into two parts: One for IOException handling and the other for parsing the JSON of extra-keys

Grimler91:
> Nope, that wasn't me, this is already in master.

Ah sorry, I blame my small phone screen. 
Anyways, great!

fornwall:
Thanks!

PR #955, opened 2018-12-20
Implement CSI 3 J - Clear transcript

iamahuman:
You may change `mScreen.clearTranscript();` to `mMainBuffer.clearTranscript();` if you think it should clear the scrollback even in alternate screen mode.

Since there seems to be no consensus of the semantics of scrollback clear commands among Terminals I was not sure which one to implement. But hey, better to have something than nothing ;)


iamahuman:
(WIP)

xeffyr:
> (WIP)

Is work still in progress ?

Your current changes doesn't work with command `reset`. However `tput E3` and `printf "\e[3J"` are working.

iamahuman:
@xeffyr I'm very surprised it actually did work, since the code wasn't tested because I was coding it in Termux itself(!) and didn't look thoroughly through the code for verification. Meanwhile it needs additional fixes to the `com.termux.view.TerminalView.onScreenUpdates()`

fornwall:
I think supporting `CSI 3 J` would be nice!

Could you look at the necessary changes to `onScreenUpdated()`, and also perhaps adding a test to `ControlSequenceIntroducerTest`?

iamahuman:
@fornwall Fixed and rebased on 7f4df8328cfd4209b1c98d558e6274fd5a67f729 .

fornwall:
Thanks! This will be part of the next release (v0.67) which will be made shortly.

iamahuman:
PROTIP: Before this code gets released, repeatedly clear the screen and toggle the soft keyboard alternately to get all the scrollback cleared. Reducing the font size (i.e. larger terminal size / text density) accelerates this process.

Issue #627, opened 2018-03-17
When exiting Termux, pressing enter doesn't work on 8.0 on Galaxy S9+


Root-voidX:
I dont know if this is a bug or a known issue, but on Galaxy S9+ running Android 8.0 when exiting termux by entering exit. You get the message to press enter when closing, but pressing enter doesn't work. I have to exit via notification from drop down menu.

xeffyr:
Same problem on Samsung Galaxy S7 with Android 8.0. But it doen't happens always and I still able to close session after waiting some time.

l-jonas:
@Root-voidX Is it like https://github.com/termux/termux-app/issues/126 (caused by the word prediction feature of Samsung virtual keyboard)?

MickeyLane:
S9+ here - Go to Google Play (store) and select Gboard. In my case it was already installed. Open, push two buttons to enable and the termux character problem goes away. I don't recall if I restarted the phone or not.

drunkwcodes:
Termux won't stop even with exit button in notification bar after SamSung upgrading devices to Android 9!

xeffyr:
@drunkwcodes Solution is in https://github.com/termux/termux-app/issues/1094#issuecomment-486050462.
This cannot be solved from Termux's side. Also problem is Samsung-specific.

xeffyr:
Sessions are now auto-closeable so issue shouldn't be significant now. Keyboard switching seems fixes problem too.

drunkwcodes:
@xeffyr  Thanks!

PR #965, opened 2018-12-29
terminal-view: fix keyboard's input type

xeffyr:
On Samsung Galaxy S7 I often getting invalid keyboard type when switching from other application to Termux. This happens on the latest git version as well as on previous ones.

This screenshot perfectly explains what happens:
screenshot_20181229-171013_termux

***

This problem comes from this code in TerminalView.java:
``` .java
outAttrs.inputType = InputType.TYPE_NULL;
```
\- TYPE_NULL does not reset the input type on some devices (Samsung).

~~It also possible that this can deal with https://github.com/termux/termux-app/issues/627.~~

PS. I saw related issue somewhere, don't know where exactly. Maybe either on Gitter or Google+.

### DIFF
--- a/terminal-view/src/main/java/com/termux/view/TerminalView.java
+++ b/terminal-view/src/main/java/com/termux/view/TerminalView.java
@@ -235,14 +235,12 @@ public final class TerminalView extends View {
 
     @Override
     public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
-        // Using InputType.NULL is the most correct input type and avoids issues with other hacks.
-        //
         // Previous keyboard issues:
         // https://github.com/termux/termux-packages/issues/25
         // https://github.com/termux/termux-app/issues/87.
         // https://github.com/termux/termux-app/issues/126.
         // https://github.com/termux/termux-app/issues/137 (japanese chars and TYPE_NULL).
-        outAttrs.inputType = InputType.TYPE_NULL;
+        outAttrs.inputType = InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD | InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS;
 
         // Note that IME_ACTION_NONE cannot be used as that makes it impossible to input newlines using the on-screen
         // keyboard on Android TV (see https://github.com/termux/termux-app/issues/221).


Grimler91:
Nice! 

I remember assisting someone on gitter with this issue. I don't remember an issue on github.

xeffyr:
The `TYPE_CLASS_TEXT` however still enables predictive input and it can't be disabled by adding flag `TYPE_TEXT_FLAG_NO_SUGGESTIONS`.

I still searching for solution since Jackpal's Android Terminal uses `TYPE_NULL` but it *resets the input type* somehow, so keyboard becomes correct.

xeffyr:
Switched to
``` .java
InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD | InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS
```
According to commit history, this is what was used on the early Termux versions.

Still no idea why this problem does not affect [Jackpal's Android Terminal](https://github.com/jackpal/Android-Terminal-Emulator), it use `TYPE_NULL` as Termux:
``` .java
    @Override
    public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
        outAttrs.inputType = mUseCookedIme ?
                EditorInfo.TYPE_CLASS_TEXT : // word-based input
                EditorInfo.TYPE_NULL; // character-based input
        return new BaseInputConnection(this, true) {
```

fornwall:
I think `InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD | InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS` is a bit of a hack, `EditorInfo.TYPE_NULL` should be the correct one, right?

However if the input change improves across keyboards I guess it's still a pragmatic change. Perhaps we could write down the behaviour for different keyboards here?

Allowing the input type to be set in `termux.properties` with something like

```
editor-type = TYPE_TEXT_VARIATION_VISIBLE_PASSWORD | TYPE_TEXT_FLAG_NO_SUGGESTIONS
```

could allow for easier experimenting for users, and allow workarounds for odd keyboards to be done by users.

xeffyr:
As this issue mostly related to samsung keyboards, I guess it will be better to implement for `termux.properties` something like this:
```
samsung-keyboard-fix = true
```

The purpose of such option will be more clear than something like `TYPE_TEXT_VARIATION_VISIBLE_PASSWORD`.

PR #1009, opened 2019-02-01
Add option to create sessions with the active session's working directory

Edontin:
Adds the option to create new Termux sessions with the active shell's working directory.

Use session.cwd-on-create=current within termux.properties to enable this behaviour.

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -585,7 +585,15 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
                 }
             }
             String executablePath = (failSafe ? "/system/bin/sh" : null);
-            TerminalSession newSession = mTermService.createTermSession(executablePath, null, null, failSafe);
+            String workingDirectory = null;
+            if (mSettings.mUseCurrentSessionCwd) {
+                final TerminalSession currentSession = getCurrentTermSession();
+                if (currentSession != null) {
+                    workingDirectory = currentSession.getCwd();
+                }
+            }
+
+            TerminalSession newSession = mTermService.createTermSession(executablePath, null, workingDirectory, failSafe);
             if (sessionName != null) {
                 newSession.mSessionName = sessionName;
             }
--- a/app/src/main/java/com/termux/app/TermuxPreferences.java
+++ b/app/src/main/java/com/termux/app/TermuxPreferences.java
@@ -67,6 +67,8 @@ final class TermuxPreferences {
     boolean mBackIsEscape;
     boolean mShowExtraKeys;
 
+    boolean mUseCurrentSessionCwd;
+
     String[][] mExtraKeys;
 
     final List<KeyboardShortcut> shortcuts = new ArrayList<>();
@@ -191,6 +193,8 @@ final class TermuxPreferences {
         }
 
         mBackIsEscape = "escape".equals(props.getProperty("back-key", "back"));
+
+        mUseCurrentSessionCwd = "current".equals(props.getProperty("session.cwd-on-create", "default"));
 
         shortcuts.clear();
         parseAction("shortcut.create-session", SHORTCUT_ACTION_CREATE_SESSION, props);
--- a/terminal-emulator/src/main/java/com/termux/terminal/TerminalSession.java
+++ b/terminal-emulator/src/main/java/com/termux/terminal/TerminalSession.java
@@ -172,8 +172,14 @@ public final class TerminalSession extends TerminalOutput {
         if (mShellPid < 1)
             return null;
         try {
-            return new File(String.format("/proc/%s/cwd", mShellPid))
+            final String cwdSymlink = String.format("/proc/%s/cwd/", mShellPid);
+            String outputPath = new File(cwdSymlink)
                 .getCanonicalPath();
+            if (!outputPath.endsWith("/"))
+                outputPath += '/';
+
+            if (!cwdSymlink.equals(outputPath))
+                return outputPath;
         } catch (IOException | SecurityException e) {
             // Ignore.
         }
--- a/terminal-emulator/src/main/java/com/termux/terminal/TerminalSession.java
+++ b/terminal-emulator/src/main/java/com/termux/terminal/TerminalSession.java
@@ -8,6 +8,7 @@ import android.system.Os;
 import android.system.OsConstants;
 import android.util.Log;
 
+import java.io.File;
 import java.io.FileDescriptor;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
@@ -164,6 +165,19 @@ public final class TerminalSession extends TerminalOutput {
     /** The terminal title as set through escape sequences or null if none set. */
     public String getTitle() {
         return (mEmulator == null) ? null : mEmulator.getTitle();
+    }
+
+    /** Returns the shell's working directory or null if it was unavailable. */
+    public String getCwd() {
+        if (mShellPid < 1)
+            return null;
+        try {
+            return new File(String.format("/proc/%s/cwd", mShellPid))
+                .getCanonicalPath();
+        } catch (IOException | SecurityException e) {
+            // Ignore.
+        }
+        return null;
     }
 
     /**


Quasic:
I would find this more useful as a separate button, unless I could make a script to change the behavior mid-session. Still would be useful, though.

xeffyr:
Fairly easy was able to get this:
screenshot_20190203-031120_termux

To reproduce:
1. `mkdir ~/test`
2. `cd ~/test`
3. Open new session and `rm -rf ~/test`.
4. Open new session.

GNOME terminal in this case opens $HOME.

Quasic:
Thanks for trying the checkbox, but go ahead and focus on the basic feature, first. It looks like it just needs to check that the path exists before using it instead of home.

I don't want to complicate things. I do find it harder to cd home on the touchscreen keyboard, but it's not as bad as retyping or copy-pasting a path. I could always make an alias cdhome or something...

Edontin:
Hmm, the simplest solution would be to ensure that we aren't resolving the same path as the one we just checked tried to canonicalize.
However, an alternative solution would be to call readlink via JNI and then verify that path exists before returning. Although, I'm not sure which you'd prefer?

Also, cd without any parameters will return home.

Quasic:
@Edontin, wow, thanks, I can't believe I overlooked that. I also use DOS, (which prints the cwd when given no parameters.) So I learned to always supply a path for better consistency.

Use the simplest solution. I was speaking more generally. I almost edited my comment because I wasn't sure that was the exact check needed, but I figured you would know.

Quasic:
I might try to actually cd to the /proc/.../cwd for the current session's shell pid before opening the new session. I doubt it would matter in practice, but it's good to test the edge cases.

Edontin:
Just verified that this still canonicalizes the previous path, resolving all symlinks.

Within shell 1:

```
% mkdir test; cd test; cd /proc/11475/cwd
% pwd
/proc/11475/cwd
```

Within a new shell:

```
% pwd
/data/data/com.termux/files/home/test
```

Upon doing the same with the test directory removed, after the second cd, shell 2 goes to Termux's home.

fornwall:
@Edontin Thanks a lot, and sorry for the delay here!

Merged as e28be01dc23d480d3938ac26e7fc2cff39379c3e, this is part of version 0.86 of the app which has now been released. This has the behaviour unconditionally - let's try that and add settings for opting in (or out) if necessary.

tomty89:
Oh no please. Why would you do something like this...
I thought it was some bug when I find the pwd of another session "leaks" to another session (and it seem to work "randomly" under some circumstances).

This is such a nasty "convenience" and the behavior is utterly unexpected. No offense but I really find this unacceptable. (A new session is supposed to be a new session and needing to "cd back to home" is essentially insane).

Please do not make this the default behavior. Please. Please. Please.

snogglethorpe:
I agree with tomy89, this behavior (if I understand it correctly) is just kind of bizarre.... 

If I start a new session using the global Termux GUI, I expect the working directory to be something constant like $HOME.

If the new session is created via some shell command, _then_ it might make sense for the initial working directory to be that of the command.

tomty89:
For the record, I don't think this even worths cramming the side bar with another button. If anyone thinks this even worth preserving, it should just be done via a configuration file option. (Well, or maybe the more menu. I wouldn't really like it though.)

Edontin:
Hi @tomty89,

> I thought it was some bug when I find the pwd of another session "leaks" to another session (and it seem to work "randomly" under some circumstances).

Could you please provide information about the circumstances for the random behaviour to occur?
Are symlinks involved?

> P.S. Seriously, I have never seen such behavior elsewhere. Not gnome-terminal. Not tmux. At the very least not as default. The only thing similar I ever saw I think is that you can open a terminal from nautilus (the GUI file manager of GNOME) that will have the same PWD as it does. (And that's hardly a relevant case...)

The behaviour is similar to GNOME Terminal and Konsole's behaviour upon creating a new tab, to reduce the need to `cd` into deep directory structures.

I can see why it would be unpleasant if you're having to `cd` home regularly.

The opt-in/opt-out config approach would be the most straightforward to add.

xeffyr:
> The opt-in/opt-out config approach would be the most straightforward to add.

New button for context menu? E.g. "Open session in current directory".

tomty89:
@Edontin I am not really sure if it was really bugged. Maybe it merely seemed random to be because I didn't even imagined it could be a new "feature"/"convenience". Either way I am not really interested in investigating that as it's not really of my concern. It's not rational to me anyway.

Again, I've never seen gnome-terminal do that for either new tab or new window. My guess would be some shell "enhancement" you got somewhere. (And I am not interested in investigating that either.)

It's not really about what I find pleasant. The reason this should not be the behavior is because, it assumes that the most frequent workflow of most users is to *open another session from where he last changed to*, base on essentially no ground. And even more importantly, action like "open a new session" should have *consistent* behavior (that's the real reason why it should always open from home or any place one configured otherwise, instead "we usually do things under home"), so that it is not biased and/or causes surprises.

It's not like I don't see why would one want that *sometimes*, but only doing it with a session-based menu button will make sense. (Otherwise it would be just a "stupid-convenience", if you know what I mean.)

While personally I don't find it really worth a menu slot in the more menu, it makes more sense than some static toggle. (If someone don't mind bothering, a menu from long pressing the session item is better place than the more menu I think. Actually I wouldn't mind the name setting being replaced entirely. But some users might.)

Edontin:
@xeffyr @tomty89 

If it should be something the user toggles on or off, it may make sense to introduce a profile mechanism. This would allow sessions to be created with a profile, or a profile to be applied onto existing sessions. The user can, as a result, pick groups of settings, including whether to spawn new sessions in the current working directory.

The only thing that may be a little difficult, conceptually, is handling the Android context-stored preferences e.g. text size, keep screen on.

The way GNOME Terminal has implemented creating sessions within the working directory, depends on an override for the shell's status line, provided in `vte.sh`. This only supports zsh and bash, so I would imagine there are a lot of configurations which prevent it from working.

tomty89:
The only things that could make sense is that, a new session only inherits the working directory of a *specific* existing session when the user instructed it to *upon its creation*, and that we provides a quickest way for a user to tell the decision *at that moment*. Otherwise it would be just clumsy and stupid (definitely not the one in KISS). It's simply not a deterministic thing. (You don't always just want or not want it in a certain single workflow. That's not even the lifestyle of a robot.)

I have no idea why you even started talking about other preferences, just like I don't have the faintest why @fornwall merged this (and not even in the way you wanted it).

Can you even point us to anything concrete (a code/script line, a screenshot of a setting/toggle...) which shows that vte/gnome-terminal/bash itself (as in, whatever upstream ships) has anything even remotely related to what you have proposed? (Like, a line in https://github.com/GNOME/vte/blob/mainline/src/vte.sh.in.) While I don't think it really matters in what we should/need to do here, I don't want to feel like discussing someone else's dream.

Edontin:
Here are a few terminal emulators which implement the behaviour of keeping the shell's working directory and how they function:

* [libvte](https://github.com/GNOME/vte/blob/26bc760d0944232e2d18640b294edf058c9a936c/src/vteseq.cc#L6403) - Reads escape sequences from `vte.sh` to determine the current working directory.
    * [GNOME Terminal](https://github.com/GNOME/gnome-terminal/blob/79b2291b12cfa106149437d63394deddddbc7dca/src/terminal-screen.c#L1455) - A setting to control the behaviour appears to have been added in GNOME/gnome-terminal@79b2291. Starting in the shell's working dir has existed since v2.5.0. Previously, was always on.
    * [XFCE4 Terminal](https://github.com/xfce-mirror/xfce4-terminal/blob/d09a336d75f3ceb173c6031ba6dd615863d57b67/terminal/terminal-window.c#L1729) - Working directory configurable via settings. Defaults to current working dir.
* [Kitty](https://github.com/kovidgoyal/kitty/blob/4dde98305e3f5715c50e9c4069b7300c522c69f9/kitty/boss.py#L1028) - Configurable as user-defined shortcut (ad-hoc).
* [Konsole](https://github.com/KDE/konsole/blob/5f60bc5a6f1bc0de7d1b99a4aa8e47b0e71b0c4e/src/MainWindow.cpp#L498) - Has the ability to change whether to spawn in a directory or the working dir. Configured per terminal profile. Defaults to on.
* [QTerminal](https://github.com/lxqt/qterminal/blob/b4626fd24a4514234169e920f6ed5fb46285315f/src/tabwidget.cpp#L71) - Configurable via settings. Defaults to off.
* [iTerm2 (macOS)](https://github.com/gnachman/iTerm2/blob/55d00d04fb96b14eea791976b17dd68d586ddfa0/sources/iTermInitialDirectory.m#L63) - Has the ability to change whether to spawn in a directory or the working dir. Configured per terminal profile. Defaults to off.

By no means is this a representative sample of all terminal emulators, but it does indicate that the behaviour is possible to find, in some form, generally being configurable via a preference.

As for profiles, both GNOME Terminal and Konsole implement profiles as a group of settings which affect a session. This solves the problem of managing multiple workflows by allowing the user to specify settings, such as defining whether new sessions inherit the current session's working directory. (Therefore, providing the mechanism you are describing for ad-hoc session creation in various working directories.)

Below is how Konsole allows you to spawn new tabs or windows with a profile:

new-tab-profile-konsole

In Konsole's case, if you spawn a new session, using a profile where the shell's working dir is reused, then the new session will use the same working directory.

Session profiles can also be switched at runtime, within Konsole:

switch-profile-konsole

GNOME Terminal implements profile configurations much like Konsole:

new-tab-profiles-gnome

switch-profiles-gnome

I would imagine profiles could be defined the following way in Termux:

`termux.properties`

```
# All unprefixed settings are applied to a 'default' profile,
# these values are inherited by all other profiles.
exampleKey = value

# Profile keys are defined as profile.<profile-id>.<key> 
profile.exampleProfile.name = Example Session Display Name
# When using 'exampleProfile', 'exampleKey' is overridden to contain
# value2, rather than value1.
profile.exampleProfile.exampleKey = value2
```

The only issue with this definition is that the Android Context values (why I mentioned them in my previous comment) may be difficult to associate with profiles, given that they are not stored alongside within `termux.properties`.

tomty89:
I am not sure why you bring in the the profile idea, or why you think it need to be implemented like that. Is it because even you don't *always* want this behaviour so you think there should be a "runtime" mechanism to change it? Why not a simple setting and/or a toggle button if we have the space.

The thing is, isn't the easiest way to have it on-demand a "new session from" button as I proposed? And with that it won't even just inherit the PWD from the last session but any *specfic* session you want to target at a specific point.

And even from your list of examples, it doesn't seem like it's a "default behavior of most TE". Not in my experience with vanilla (i.e. no silly distro "preferred enhancements") GNOME 3 Terminal and tmux, etc. either.

I don't mind it being implemented anyway, but there should really be a way to turn it off, and I strongly opposed it being the default (I hardly see its sense / how it could smooth up workflow in most of the time; also note that in Termux we do not have both "new terminal" and "new tab", but only "new session").

P.S. I guess it might never be your intention to have it as default / always on. In any case the above opinion merely addresses the current behaviour (or even, problem).

PR #1028, opened 2019-02-11
sessions: failsafe session is now accessible via separate launcher icon

xeffyr:
Enables additional launcher icon that causes to be a failsafe session to be started instead of default.
This PR also makes session auto-closeable on exit coded *0* and *130*. Any other exit codes should be handled by user on it's own in order to avoid messages about "Press enter".

Old "failsafe" session button left untouched since it may be useful in some cases to start system shell instead of Termux's one.

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -216,8 +216,8 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
 
         final ViewPager viewPager = findViewById(R.id.viewpager);
         if (mSettings.mShowExtraKeys) viewPager.setVisibility(View.VISIBLE);
-        
-        
+
+
         ViewGroup.LayoutParams layoutParams = viewPager.getLayoutParams();
         layoutParams.height = layoutParams.height * mSettings.mExtraKeys.length;
         viewPager.setLayoutParams(layoutParams);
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/termux/app/TermuxActivity.java
```diff
@@ -367,7 +367,11 @@ public void onSessionFinished(final TerminalSession finishedSession) {
                         showToast(toToastTitle(finishedSession) + " - exited", true);
                 }
 
-                if (mTermService.getSessions().size() > 1) {
+                if (getPackageManager().hasSystemFeature(PackageManager.FEATURE_LEANBACK)) {
```
fornwall: Could you add a comment describing why this check is made (that the failsafe launcher icon doesn't show on Android TV)?

xeffyr: There problem with additional launcher icons on Android TV devices. Tested on Android virtual device from SDK. I don't know if it will work for real devices but since on AVD it is not working, I guess the optimal choice will be left old behaviour there.

xeffyr: Comments added.

xeffyr:
I also made auto-closing for only "0" and "130" exit codes as we discussed on Gitter.

eladts:
I don't like the solution of another launcher icon, it just creates clutter IMHO. Is there a reason why this functionality cannot be achieved using [app shortcuts](https://developer.android.com/guide/topics/ui/shortcuts)? Is there a way (without rooting) to remove the extra icon from the launcher?

xeffyr:
> Is there a reason why this functionality cannot be achieved using app shortcuts?

Not all launchers support shortcuts. Especially on older Android versions.

> Is there a way (without rooting) to remove the extra icon from the launcher?

Check launcher settings if it allows to hide icon.

***

Generally, this solution is final and should be fine for **average** user, i.e. should work on all ROMs and launchers.
Other variants that make environment recovery really hard will not be implemented.

xeffyr:
btw, there always will be tradeoff for issues regarding "[Process completed - press Enter]" which will not be accepted by some users.

Decide what you want:
* Ability to retrieve files from $HOME in case of environment failure while having auto-closeable sessions.

* Or have to completely reinstall application with erasing all files in $HOME due to denial-of-service condition because of shell failure ?

eladts:
Thanks @xeffyr for the quick response! Unfortunately my launcher (stock launcher on Moto X4) does not support app hiding and I don't want to switch launchers for this reason alone. I would appreciate if there is an option to hide the extra launcher icon from within Termux, which as far as I understand the only way this can be done without rooting or switching to another launcher.

xrisk:
I would urge the maintainers to consider making this second app icon a toggle-able option. I think few people need it and it isn't worth the extra icon for us.

Issue #804, opened 2018-08-10
make TermuxService (com.termux.service_execute) available to other apps


torappinfo:
So, any app can use the TermuxService execute.

Only minor changes in AndroidManifest.xml are needed:
      ```
      < permission android:name="com.termux.app.TermuxService">
      < /permission>

      <service android:name="com.termux.app.TermuxService" android:permission="com.termux.app.TermuxService">
          <intent-filter>
            <action android:name="com.termux.service_execute" />
          </intent-filter>
        </service>
```

**Comment by markosjal** [comment]
can you please clarify this? I am interested in executing streamlink which I have running in termux. I want to start the streamlink command from an android app

**Comment by ShadowJonathan** [comment]
All similar issues have been closed in favour of this one, yes, but has there been any progress or concidering since the opening of this particular issue, 6 months ago?

**Comment by edaubert** [comment]
Hi,

I took a look on how to do this.
As I am quite new on this, the approach can be wrong so don't hesitate to tell me

My usecase was to be able to use the `TermuxService` from the outside and especially from [Easer](https://github.com/renyuneyun/Easer).
To do so, I change the definition of TermuxService in `AndroidManifest.xml` like this:

```xml
        <service
            android:name="com.termux.app.TermuxService"
            android:exported="true"
            android:permission="com.termux.permission.TERMUX_SERVICE">
            <intent-filter>
                <action android:name="com.termux.service_execute" />
            </intent-filter>
        </service>
```
**It seems that the intent-filter is important because without it, there is no dialog to ask the user on Easer to allow the custom permission.**

I also need to define my custom permission:
```xml
    <permission android:name="com.termux.permission.TERMUX_SERVICE"
        android:label="@string/termux_permission_label"
        android:description="@string/termux_permission_description"
        android:icon="@drawable/ic_launcher"
        android:protectionLevel="dangerous"/>
```

Then I modify Easer to be able to  access Termux.
To do so, I express the requirement of my permission by adding:
```xml
    <uses-permission android:name="com.termux.permission.TERMUX_SERVICE" /> 
```
in `AndroidManifest.xml` and I add the dynamic permission request in the code of Easer:
```java

    @Override
    public boolean checkPermissions(@NonNull Context context) {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
            return PluginHelper.checkPermission(context, "com.termux.permission.TERMUX_SERVICE");
        } else {
            return true;
        }
//        return true;
    }

    @Override
[...]

PR #1029, opened 2019-02-11
add custom permission to call TermuxService from external apps

edaubert:
This pull request is an attempt to provide a solution to https://github.com/termux/termux-app/issues/804

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -4,6 +4,12 @@
     android:installLocation="internalOnly"
     android:sharedUserId="com.termux"
     android:sharedUserLabel="@string/shared_user_label" >
+    
+    <permission android:name="com.termux.permission.TERMUX_SERVICE"
+        android:label="@string/termux_permission_label"
+        android:description="@string/termux_permission_description"
+        android:icon="@drawable/ic_launcher"
+        android:protectionLevel="dangerous"/>
 
     <uses-feature android:name="android.hardware.touchscreen" android:required="false" />
     <uses-feature android:name="android.software.leanback" android:required="false" />
@@ -111,7 +117,12 @@
 
         <service
             android:name="com.termux.app.TermuxService"
-            android:exported="false" />
+            android:exported="true"
+            android:permission="com.termux.permission.TERMUX_SERVICE">
+            <intent-filter>
+                <action android:name="com.termux.service_execute" />
+            </intent-filter>
+        </service>
 
         <receiver android:name=".app.TermuxOpenReceiver" />
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -48,4 +48,7 @@
     <string name="file_received_title">Save file in ~/downloads/</string>
     <string name="file_received_edit_button">Edit</string>
     <string name="file_received_open_folder_button">Open folder</string>
+
+    <string name="termux_permission_label">Termux Service Permission</string>
+    <string name="termux_permission_description">Access Termux Service to run some scripts</string>
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -50,5 +50,5 @@
     <string name="file_received_open_folder_button">Open folder</string>
 
     <string name="termux_permission_label">Termux Service Permission</string>
-    <string name="termux_permission_description">Access Termux Service to run some scripts</string>
+    <string name="termux_permission_description">Allows application to execute arbitrary shell commands within Termux.</string>
 </resources>


EpicOrange:
Just curious, does this mean that apps with this permission effectively have access to `/data/data/com.termux`?

edaubert:
No you don't have access to it.

You need to use some storage permission I guess.
I guess another possibility could be to send the file to termux before trying to execute it.

edaubert:
Yet another way to do it is to give storage access to termux and store the script where it can find it.

https://wiki.termux.com/wiki/Termux-setup-storage

edaubert:
So I have completed a termux plugin on Easer which fully use the permission.
It uses the fact that termux provide a document provider. This provider allow external apps to get access to some files inside the termux environment.

Please have a look here: https://github.com/renyuneyun/Easer/pull/205

fornwall:
We discussed this in a chat session (https://gitter.im/termux/dev) today, and I think this is a really interesting thing to have. Some notes from the chat:

The thing we want to minimise is the risk for evil apps to infect Termux installations (due to people ignoring/not understanding what the permission does). A conservative change could be to make this some kind of opt in, like requiring a property in termux.properties for this to be allowed
so we both require that permission, but we also need something like external-apps-can-execute=*, or external-apps-can-execute=com.app1,com.app2.

Or it could be a runtime permission. Perhaps a dialog like "The app XXX is trying to run a script. Allow that?" with a checkbox to remember the choice though that obviously complicates the code and also the user experience. Thoughts on that?

@xeffyr: A dialog like "The app XXX is trying to run a script. Allow that?" with a checkbox to remember the choice seems better than termux.properties.

@fornwall What I can't decide on is if that is necessary :). One point is that the app is a terminal emulator, users should be expected to a bit more technically skilled than an average app users and careful with accepting permissions.

... On the other hand, if the effect of opening up permissions is that several users have e.g. their ssh/gpg private keys stolen or likewise, it's bad regardless of if those users can be blamed for not reading through permissions.

Feedback, thoughts and ideas are welcome here! Even with the above said I'm leaning against merging this as is.

edaubert:
As @fornwall said, I am not sure we need to add an extra layer of security.

First as it is said the users of termux are expected to be advanced users and so, in my opinion, should be aware of what implies to accept a permission. Well for me even non advanced users should be aware of this...

Second, it is the termux app which defines the permission and so the description can be explicit. And if a user do not want to read the permission dialog why should he want to read yet another popup dialog.

About the part of "their ssh/gpg private keys stolen or likewise, it's bad regardless of if those users can be blamed for not reading through permissions" you completly right but if the user gives root access to another app, it can also stole its data and termux is not able to protect those data even without speaking of permission. In the end, and this is mostly my really subjective opinion 😃, users are still responsible of what happens and permissions are something which is not just a bother which display annoying dialog.

However, if you are still willing something, i will propose to add a notification with multiple actions.
The notification provide a small text about running a script (name of the script and who ask to run it)
- action 1: always run it
- action 2: run it once
- actiin 3: never run it

Why I am proposing a notification is because I don't want to be interrupted by something that I may have not ask for (e.g. an automation script).
Plus if people do not care about popup dialog, they may care about their notifications.

But again for me, this is only an extra layer of security for supposed advanced termux users and I think they do not need that.

ujeropoc:
little late too the party, but what about outsourcing this to another app?:  #1048

xeffyr:
@ujeropoc Executing commands via URLs is weird solution from security side. It just won't be possible to use Android permissions here.

MisterY:
What @edaubert suggests, with a dialog that resembles SuperSu's or through a notification,  is handy. But it would also be good to able to pre-configure the permissions before running because, as was mentioned, the point is to run automated and we may not know in advance when this will happen.
On the pragmatic side, would it be possible to come up even with a partial implementation, perhaps in a test version, which implements at least a rudimentary version of authorization, so that we can try it out and provide feedback? That way we could iron-out any issues that happen during actual use.
I'd like to create an app that provides a GUI for entering transactions and utilizes Ledger in the background to list the accounts defined by the user. Ledger is available as a binary package and can be run on the command line.
Thanks for all the great work so far.

Ciantic:
I don't think this needs extra layer of security either, termux is for developers mostly anyway. Having just blanket permission is just fine. If you want extra security, you could restrict it to e.g. folder:

```~/appscripts/*```

So that apps are allowed to run scripts only from appscripts folder.

**Comment by xalexalex** [comment]
I also agree with the fact that having a working version (whose security can later be improved) is better than not having this functionality at all.

**Comment by RoninNada** [comment]
Would someone be able to provide an example of this being implemented/used? Sending a command an executing from another from another app? I already built a version of termux with this implemented just can't get the other part to work. I want it to work off a button press. Thank You

**Comment by xeffyr** [comment]
@RoninNada Make sure that application implements Termux service call in the way like:
``` .java
// scriptUri is URI path to script file.
Intent executeIntent = new Intent("com.termux.service_execute", scriptUri);
executeIntent.setClassName("com.termux", "com.termux.app.TermuxService");

// Whether to execute script in background.
//executeIntent.putExtra("com.termux.execute.background", true);
```
Service access permission must be granted to that "another application" in order to make code above work.

RoninNada:
@xeffyr , thank you so much really helpful

Ciantic:
I'm trying to understand the pushes above, was this merged? Is it now possible to call shell apps from within other apps?

xeffyr:
It isn't merged and current implementation available here won't be merged. With @fornwall we agreed to:

* Do not let applications to access TermuxService directly.
* Implement separate service for calling with external applications (`TermuxServiceForExternalApps`).
* TermuxServiceForExternalApps will do necessary checks (e.g. property `allow-external-apps` is enabled).

bdovaz:
@xeffyr How is that going?

soso-cloud:
Want to know these account password
https://www.facebook.com/myamya.shein.75

xeffyr:
Implemented in https://github.com/termux/termux-app/commit/db3ff7b24a2cb8c84b00cc5cec083361018569a1.

bdovaz:
@xeffyr I just saw the commit:

https://github.com/termux/termux-app/commit/db3ff7b24a2cb8c84b00cc5cec083361018569a1

But how can I retrieve the command output? In my case I need that and I don't see an example.

Thanks.

xeffyr:
> But how can I retrieve the command output? In my case I need that and I don't see an example.

This isn't possible with intents. Here you just start an activity or service with specific parameters.

bdovaz:
@xeffyr but is there any way to achieve what I want? Thanks.

ZhaoTzuHsien:
When will this important improvement be released? :)

xeffyr:
@ZhaoTzuHsien It is already released by https://github.com/termux/termux-app/commit/db3ff7b24a2cb8c84b00cc5cec083361018569a1 in application version **0.95**.

bdovaz:
@xeffyr you didn't answer my last question. Is there any way to achieve that? Or will it be possible? Or at least, can I read some file where the output it's being written so I can pool it? Thanks.

xeffyr:
>Is there any way to achieve that? Or will it be possible?

Will be possible as soon as *someone* will write an api engine for that, with IPC, proper permission checks, etc.

agnostic-apollo:
@xeffyr can the `EXTRA_EXECUTE_IN_BACKGROUND` also be received when the `com.termux.RUN_COMMAND` action intent is processed and passed to the `TermuxService`. It would make it easier to run background commands like with Tasker. Basically show the notification but not the session activity unless user manually opens it if required. There seems to be reasonable security for it to be an issue. Thanks

agnostic-apollo:
@xeffyr it seems that the RUN_COMMAND_ARGUMENTS is not working. You are receiving and sending it as a string extra in [RunCommandService](https://github.com/termux/termux-app/blob/51370799c765d1774abac61a6f2eedf09156d211/app/src/main/java/com/termux/app/RunCommandService.java#L55) but receiving it as a string array extra in [TermuxService](https://github.com/termux/termux-app/blob/51370799c765d1774abac61a6f2eedf09156d211/app/src/main/java/com/termux/app/TermuxService.java#L147) resulting it in being ignored as far as I can see. A simple `sleep 5` commands shows `sleep: missing arguments`. 

You probably have your own test app but leaving following instructions for others:

The `com.termux.permission.RUN_COMMAND` has been added to the Tasker `AnndroidManifest.xml` in the list of requested permissions in [v5.9.3.rc](https://www.reddit.com/r/tasker/comments/i0253r/dev_tasker_593rc_getting_ready_for_public_release/). It should automatically be requested when `TermuxCommand` function is used in a `Tasker Function` action or you can grant it manually from something like Tasker app permissions -> Additional permissions -> `Run commands in termux environment`, specially if you intend to run am shell commands instead. When you run a command or script in the function or use am command, a new termux session will open up and run the command. It will automatically close by default when the command completes. You can put the session in background mode if you want after that. However, starting it in background mode is not supported and has been [requested](https://github.com/termux/termux-app/pull/1029#issuecomment-649715644) requested.

After you grant the permission, you also need to add the `allow-external-apps=true` property in `~/.termux/termux.properties` file for intents to be allowed. You can run the following command in a non-root termux session to add the line to the end of the file if it doesn't exist already.

```
cd ~ && mkdir -p .termux && grep -sq  "allow-external-apps=true" ./.termux/termux.properties || echo $'\n'"allow-external-apps=true" >> ./.termux/termux.properties
```

After that you can run the command with a `Tasker Funtion` action or am command using a `Run Shell` action. The arguments and working directory are optional. An example am command would be something like the following.
```
am startservice --user 0 -n com.termux/com.termux.app.RunCommandService -a com.termux.RUN_COMMAND --es com.termux.RUN_COMMAND_PATH "/data/data/com.termux/files/usr/bin/top" --es com.termux.RUN_COMMAND_ARGUMENTS "-n 5" --es com.termux.RUN_COMMAND_WORKDIR "/data/data/com.termux/files/home"
```

xeffyr:
> You are receiving and sending it as a string extra in RunCommandService but receiving it as a string array extra in TermuxService resulting it in being ignored as far as I can see. A simple sleep 5 commands shows sleep: missing arguments. Do you want a pull request?

Ok, that's a bug. You can submit a pull request to fix it.

agnostic-apollo:
Should I also include support for passing the background flag as well or not?

xeffyr:
Yes.

agnostic-apollo:
Okay thank. I will do it and let you know.

agnostic-apollo:
I have added a pull request [here](https://github.com/termux/termux-app/pull/1679). Let me know if any changes need to be made. I also updated the docs a bit, I hope that's all right.

minhdanhsc:
Can I run the command from another application to termux? Who has a sample for my reference?

xeffyr:
@minhdanhsc https://github.com/termux/termux-app/blob/master/app/src/main/java/com/termux/app/RunCommandService.java#L23-L62

Issue #1032, opened 2019-02-12
Style's background is black when opening the app after wake locking


bakaika:
**Problem description**
I manually edited `~/.termux/colors.properties` to apply a theme to Termux, after activating wake lock and closing all sessions. When I re-open the app, the background is black. (It also does the same using Termux:Style themes).

**Steps to reproduce**
Apply styles to Termux with either Termux:Style or manually editing `~/.termux/colors.properties` & `termux-reload-settings`. Choose a background colour that isn't black. Close all sessions from within Termux until it exits. Open the 
app, the background is black, not the colour you chose.

**Expected behavior**
The background colour should stay the colour it was originally set to.

**Additional information**
The colours are restored when opening a second session, or disabling wake lock, ans opening the app once again.

The theme I currently have:
![screenshot_20190212-205818_termux](https://user-images.githubusercontent.com/47267353/52637060-31e7d980-2f09-11e9-8365-91e018186bae.jpg)
![screenshot_20190212-205945_termux](https://user-images.githubusercontent.com/47267353/52637085-3f04c880-2f09-11e9-9e78-e5ff2a6d757a.jpg)

Test theme:
![screenshot_20190212-210537_termux](https://user-images.githubusercontent.com/47267353/52637358-fe597f00-2f09-11e9-849a-ac75460d1ae8.jpg)
![screenshot_20190212-210545_termux](https://user-images.githubusercontent.com/47267353/52637363-02859c80-2f0a-11e9-9bf5-2feadd4e42a7.jpg)

xeffyr:
Observed the same issue yesterday, thought it affects only me.
Can be fixed in this way https://github.com/xeffyr/termux-app/commit/ef12842e7e2255e622cae40c3abc46320643593b.

bakaika:
Oh wow. Only affects you, what device you got?

xeffyr:
I use custom build of Termux application which has styling embedded. So I just didn't tested Termux:Styling well with original app.

xeffyr:
I opened a PR which fixes that.

bakaika:
Ah I see, nice.

PR #1033, opened 2019-02-12
termux service: make sure that styling is always applied whenever session starts

xeffyr:
Fixes https://github.com/termux/termux-app/issues/1032.


panupong12:
https://www.facebook.com/oumna.sirivun:

PR #1114, opened 2019-04-25
Make am work on Android Q

michalbednarski:
Export `ANDROID_RUNTIME_ROOT` environment variable if it's present

With that, `am` and `apksigner` work on Android Q


fornwall:
Thanks! This has been released in version 0.69 of the app, which should be available as an update shortly.

deoxal:
What does this mean exactly? I tried reading up on it [here](https://android.googlesource.com/platform/art/+/master/tools/art), but I just don't understand it.

PR #1123, opened 2019-04-28
Remove failsafe activity

xeffyr:
Second icon of Termux application is quite confusing for some users.
We decided that it will be better to move such stuff to a separate application


xeffyr:
Should be reworked as we may want to provide a shortcut for users with Android 7.1+.

PR #1135, opened 2019-05-17
Tweak notification icon color to make it work with dark notifcations

nishithkhanna:
As the notification icon and text is set to black, ROMs like Samsung OneUI and themes that change notifications to have a dark background will have visibility issues with Termux notifications. This commit sets a neutral color which will be visible on both white and dark backgrounds.


fornwall:
Thanks a lot! This is part of version 0.70 of the app, which is now being pushed out as an update.

nishithkhanna:
Glad was able to help in some way. Thanks for the update.

PR #1141, opened 2019-05-21
Remove intent category "default" from manifest

xeffyr:
* Intent category "default" is not needed after dropping failsafe icon.

* Also move the routines to clear /tmp dir to TermuxService.onDestroy().

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -38,12 +38,10 @@
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
-                <category android:name="android.intent.category.DEFAULT"/>
             </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LEANBACK_LAUNCHER" />
-                <category android:name="android.intent.category.DEFAULT"/>
             </intent-filter>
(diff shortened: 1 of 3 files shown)


tomty89:
@xeffyr

Should `TMPDIR` clean up be performed after `stopForeground()`?

Also, can you make it clean its canonical file instead so that it also works when e.g. `PREFIX` is a symlink?

@fornwall

What's the point of the `finishIfRunning` loops? It wouldn't kill the non-shell processes anyway.

Also, in the two ocassions where `stopSelf()` is called conditionally, one checks whether there's any background task while one does not, why is that? (In other words, why shouldn't `removeTermSession()` just call `updateNotification()` unconditionally?)

In addition, why should `stopSelf()` be called unconditionally for `ACTION_STOP_SERVICE`? (especially when `onDestroy()` doesn't deal with background tasks)

xeffyr:
> Should TMPDIR clean up be performed after stopForeground()?

Is there some benefits in doing it after `stopForeground()` ?

> Also, can you make it clean its canonical file instead so that it also works when e.g. PREFIX is a symlink?

Yes, will do.

fornwall:
This is being part of the v0.71 which is now rolling out on Google Play.

tomty89:
@xeffyr I just thought perhaps `stopForeground()` is responsible for killing all the processes.

xeffyr:
> stopForeground() is responsible for killing all the processes.

It just removes Service from foreground state. It doesn't kill any processes.

PR #1145, opened 2019-05-26
set BOOTCLASSPATH environment variable

xeffyr:
Required on at least Android 5. Otherwise utilities that use dalvikvm may fail.
Issue happens on Nexus 7 with official 5.0/5.1 firmware and possibly other AOSP ROMs.


TheBrokenRail:
This seems to be an issue with Android Q running DX as well.

fornwall:
This is part of version 0.72 of the app which has now been released.

PR #1160, opened 2019-06-10
Update AndroidManifest.xml

neswork-th:
Install on external SDcard


Grimler91:
It won't work, android doesn't allow executable files in shared storage (and `externalOnly` doesn't seem to be a valid option).

PR #1187, opened 2019-07-29
Create android

github-joey:


xeffyr:
Sorry, but we don't need to "create android".

PR #1195, opened 2019-08-01
Change method name 'with' to 'getInstance'

pdhung3012:
This class is used to represent BellUtil.  This method named 'with' is to return the singleton of BellUtil instance. Thus, the method name 'getInstance' is more intuitive than 'with'.


fornwall:
Thanks!

PR #1196, opened 2019-08-02
Update AndroidManifest.xml

v1nc:
Set `exported="true"`, so other apps can start termux service.


xeffyr:
See https://github.com/termux/termux-app/pull/1029.
Without app permission that won't be done as opens ability to execute arbitrary code.

PR #1216, opened 2019-08-15
Ignore battery optimization settings to prevent suspension

### LINKED ISSUE #377: Termux slows down in "doze mode" (battery optimization) even with the wake lock
This is not really a bug, but rather an "improvement request".

When Termux runs using the Wake Lock, the performance greatly slows down because of "Doze" (explained in details here: https://developer.android.com/training/monitoring-device-state/doze-standby.html)

> If a user leaves a device unplugged and stationary for a period of time, with the screen off, the device enters Doze mode. In Doze mode, the system attempts to conserve battery by restricting apps' access to network and CPU-intensive services. It also prevents apps from accessing the network and defers their jobs, syncs, and standard alarms.
> Periodically, the system exits Doze for a brief time to let apps complete their deferred activities. During this maintenance window, the system runs all pending syncs, jobs, and alarms, and lets apps access the network.
> As soon as the user wakes the device by moving it, turning on the screen, or connecting a charger, the system exits Doze and all apps return to normal activity

This means if you're connected with SSH on Termux, it will suddenly get really laggy (until you wake it up again), even worse if you use Termux as a web server.

See the conversation here: https://gitter.im/termux/termux?at=57a2253c836d2d0211610052

@fornwall said: "I'm just trying if disabling battery optimizaton for termux helps the wake-lock not working problem" and "I have disabled battery optimization earlier today and have not noticed any slowdowns yet. I will continue to test this."

The whitelist described here might be a solution: https://developer.android.com/training/monitoring-device-state/doze-standby.html#whitelisting-cases

Since I'm looking forward to this, here's an issue to know if it eventually does work :)

danog:
Fixes #377 by whitelisting termux when the user clicks on the wakelock button for the first time.

Even if the [android docs](https://developer.android.com/training/monitoring-device-state/doze-standby.html) say a foreground service is enough to prevent suspension, on my devices running CLI services stop sending UDP/TCP packets and get suspended if the app is reduced or the device is locked.
This is especially noticeable when running a simple `time ping 192.168.1.1`; on resume, the ICMP sequence number is way lower than the number of elapsed seconds, indicating a suspension of the process.

Anyway, I encountered this issue on my devices and noticed that manual whitelisting fixed it, thought I could contribute with an automated fix.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -13,6 +13,7 @@
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.VIBRATE" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS"/>
 
     <application
         android:extractNativeLibs="true"
--- a/app/src/main/java/com/termux/app/TermuxService.java
+++ b/app/src/main/java/com/termux/app/TermuxService.java
@@ -16,6 +16,7 @@ import android.os.Build;
 import android.os.Handler;
 import android.os.IBinder;
 import android.os.PowerManager;
+import android.provider.Settings;
 import android.util.Log;
 import android.widget.ArrayAdapter;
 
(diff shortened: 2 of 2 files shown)


fornwall:
Really nice work @danog!

This will be in v0.74 of the app, to be released in an hour or so.

danog:
Thanks <3<3<3!
Would you mind checking on my other PR for working fingerprint recognition in termux/termux-api#283?

Thanks again.

PR #1294, opened 2019-10-06
CI: do signed builds with shared dev key

xeffyr:
Configure builds with shared keystore usable by everyone.

**Important:** shared keystore isn't same as one used by Google Play or F-Droid builds. APKs signed with it are incompatible with ones from mentioned installation sources and mostly intended for use by developers and those who want to use own Termux builds.


xeffyr:
Done in https://github.com/termux/termux-app/commit/33d390a2286c42d247653a45e038bf8641f7f8d9.

PR #1301, opened 2019-10-13
Include bootstrap zip as shared library

fornwall:
This is an experiment with including the bootstrap zip as a shared library.

The gradle build now downloads the bootstrap of a specified version from bintray.

- Why include the bootstrap zip in the app?
  - To save several terabytes of monthly bandwidth on the server, and instead rely on Google to provide a speedy download of bootstrap files.
  - To allow offline installation.
  - To avoid surprising users with data traffic on first use.

- Why as a shared library?
   - To make the app splitting work on Google Play. Termux is uploaded to the Play store as an Android App Bundle, and Google generates optimized apps for devices, so that e.g. only aarch64 devices gets the aarch64 bootstrap. See https://developer.android.com/studio/projects/dynamic-delivery. F-Droid can also do the corresponding thing with a custom build.

- Issues
   - Is there a better way?
   - The shared library is never unloaded - this probably causes increased memory usage? But the library is only loaded on demand, so after the initial run of the Termux process the library is not loaded.
   - This bumps the required version of Android to API level 24 (Android 7.0) - we could create an Android-5 branch if needed as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ build/
 *.so
 .externalNativeBuild
 .cxx
+*.zip
 
 # Crashlytics configuations
 com_crashlytics_export_strings.xml
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,4 +1,6 @@
-apply plugin: 'com.android.application'
+plugins {
+  id "com.android.application"
+}
 
 android {
     compileSdkVersion 28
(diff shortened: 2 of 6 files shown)


xeffyr:
> this probably causes increased memory usage?

I don't think that additional 15-17 MB of used ram will cause problems.

We may try to `dlopen()` lib with bootstrap from another smaller library so we will be able to `dlclose()` it when bootstrap installation is finished. But that is not necessary for now.

I also noted that such installation method may cause failure of Android Runtime and application will be force-closed if bootstrap archive is big (I created 450 MB one with clang, rust, go, etc... for testing). Failure happens independently of how much RAM is available. So that method is unsuitable for https://github.com/termux/termux-packages/wiki/Termux-and-Android-10#3-generate-an-app-on-device-containing-packages.

buttaface:
Sounds like a good idea, I've recommended Termux to people and then have them complain it gets stuck when first run, maybe because of this additional download.

Issue #905, opened 2018-11-04
Prevent accidental key press when swiping the extra keys view to the left to access the text input view


vovcacik:
The title says it all.

User needs to be careful what key to lay his finger on when swiping the extra keys view, because when the swipe begins it registers as key press. One example for all: the "ARROW_UP" button will fill in previous command, this obstructs the new command you are about to type in the text input view.

https://wiki.termux.com/wiki/Touch_Keyboard

ticalc-travis:
I can confirm this issue. This seems to have only become an issue when the extra keys thing was revamped and changed to two rows. Older versions of Termux did not have this problem.

This behavior makes the input view virtually unusable because there is no way (that I can find) to swipe the extra keys *without* sending a spurious keypress to the terminal, thereby messing up the state I need to begin text input in the first place.

9001:
I tried adding a blank key to the extra keys bar, hoping that would be a draggable point with no side effects -- instead this crashes Termux :p

I would much prefer if sliding in from the right side of the screen toggled this, similar to the burger menu on the left, as I do not otherwise use the extra input bar at all. Adding it just for the purpose of toggling between android-native and terminal-native text input results in an unfortunate waste of space.

Quasic:
I have the same problem since the keyboard was updated to two rows.

epitron:
Ditto!

Is the bar a custom GUI element? If so, a good solution would be to modify its logic so that the keys are activated on release of the button, and if the bar if swiped, the release event is blocked.

Quasic:
#1014 fixed the empty button crash

Mhowser:
Any progress on this?

fornwall:
Thanks for reporting!

Fix has been released in version 0.83 or higher (Android 7+ devices will get version 0.84).

vovcacik:
Just tried it, its great. Thanks!

PR #1358, opened 2019-11-23
extra keys: handle actions UP & CANCEL separately

xeffyr:
Extra keys shouldn't perform any action if button click was cancelled.
Fixes: https://github.com/termux/termux-app/issues/905.

Additionally:

* Disable manual handling of DnD for API 28 and higher (Google simplified it and seems there no way to detect total silence mode). Maybe for higher SDK versions we even do not need to handle it manually, but I don't have device to verify this.


fornwall:
Released in version 0.83 or higher (Android 7+ devices will get version 0.84).

PR #1365, opened 2019-12-02
BackgroundJob command result responses

Archenoth:
I've added an optional final argument to the BackgroundJob constructor for a `PendingIntent` that it will respond to with the results of the commands that it has run. (A "result" bundle containing stdout, stderr, and the process exit status)

I've also added a `"pendingIntent"` extra that will be pulled from `service_execute` requests that can hold these new PendingIntents, so Termux can respond to things that can invoke these intents with the resulting command's output.

The immediate practicality of this is that it will allow things like [this issue in termux-tasker](https://github.com/termux/termux-tasker/issues/21) to be solved.

### DIFF
--- a/app/src/main/java/com/termux/app/BackgroundJob.java
+++ b/app/src/main/java/com/termux/app/BackgroundJob.java
@@ -90,7 +90,7 @@ public final class BackgroundJob {
                     // Ignore.
                 }
             }
-        };
+        }.start();
     }
 
     private static void addToEnvIfPresent(List<String> environment, String name) {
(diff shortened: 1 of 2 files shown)


Archenoth:
(The push was rebasing this onto master's new commits)

fornwall:
@Archenoth Thanks a lot! This change will be part of version 0.86 of the Termux app, which will be released within a day.

You're welcome to join https://gitter.im/termux/dev to discuss development (and ping me if I'm slow to merge any PR)!

PR #1427, opened 2020-01-15
Example of support for packages in APK

fornwall:


RalfWerner:
## Try `b=build.gradle` with my current (92) termux app in Studio:
The _compileSdkVersion_ can, **without effect,** have any value between 28 and 'android-R' and I do not change the _minSdkVersion 24_ but with every _targetSdkVersion> 28_ (also 'R') I get:
`exec("/data/data/com.termux/files/usr/bin/login"): Permission denied` (I say ELF-restriction!).
I found `$b` four times in the termux app but only changed `app/$b`, uninstalled all plugins and checked with both (Q10/R11) emulators (need 2 hours _wait_ time).
@fornwall Without your patch/fork I won't get anywhere here and I don't want to lose the connection to _**your development.**_. If I understand your sources correctly, you _fork_ from versionCode 88 (not 92) has been made and consists of various commits (e.g. add `*PackageInstaller` in` *Service`) but also `$b` is very different. A comparison with _tkdiff:_
grafik
Do you see a chance to collect the necessary patches for 92 so that I can also test?

RalfWerner:
> we need to take some gradual approach, where we keep targetSdkVersion = 28 initially and transition some minor packages to be inside apps for testing

@fornwall also a good idea! Currently _**here**_ is a targetSdkVersion = 29 (t) `*.apk` artifact with versionCode 94 (v). Can you please **repeat** this with _**t/v = 28/95?**_
I found a _workaround_ (for some problems) for the `am` bug, if you don't have a solution yet.

xeffyr:
Rebased on latest commit from master & *force-pushed*.

xeffyr:
Another force-push:
* Rebased on latest master.
* Fixed bootstrap archives for Android 10 - now generated by https://github.com/termux/termux-packaging/blob/master/scripts/generate-bootstraps.sh.

iraniangethub:
> - [ ] Handle configuration files (vimrc, etc)
> - [ ] Basic package management (command to install and deinstall packages, list installed ones, ...)
> - [ ] Speed up setting up symlinks by only processing changed packages
> - [ ] New bootstrap packages (without apt and dpkg)
> 

wget https://raw.githubusercontent.com/EXALAB/AnLinux-Resources/master/Scripts/SSH/Apt/ssh-apt.sh --no-check-certificate && bash ssh-apt.sh

iraniangethub:
wget https://raw.githubusercontent.com/EXALAB/AnLinux-Resources/master/Scripts/SSH/Apt/ssh-apt.sh --no-check-certificate && bash ssh-apt.sh

PR #1446, opened 2020-01-28
Prevent kill by Samsung Dex on plug/unplug.

### LINKED ISSUE #1259: [Feature request] Samsung DeX support.
Hi. I am an active user of Termux on Samsung DeX. The issue is that when the DeX is enabled/disabled the Termux session gets dropped.

Could Termux be updated for DeX usage please? Here is the official source from Samsung on how to adapt the app:

https://developer.samsung.com/samsung-dex/modify-optimizing

Thank you in advance.

jeansch:
Closes #1259 
Tested on my device (Samsung S10)
Documentation here: https://developer.samsung.com/samsung-dex/modify-optimizing

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -31,11 +31,11 @@
 
         <activity
             android:name="com.termux.app.TermuxActivity"
-            android:label="@string/application_name"
-            android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
+	    android:label="@string/application_name"
+	    android:configChanges="orientation|screenSize|smallestScreenSize|density|screenLayout|uiMode|keyboard|keyboardHidden|navigation"
             android:launchMode="singleTask"
             android:resizeableActivity="true"
-            android:windowSoftInputMode="adjustResize|stateAlwaysVisible" >
+	    android:windowSoftInputMode="adjustResize|stateAlwaysVisible" >
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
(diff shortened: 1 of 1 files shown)


fornwall:
Thanks a lot! This is included in version 0.90 of the app, which has now been released and should be available as an update shortly.

laem:
Thank you very much @jeansch, you've just saved me 1 minute for each of my coffee pause return :smiley:

DarthZedd:
Yo are the best! Tested on SM Note 10+ , works like a charm =)

app/src/main/AndroidManifest.xml
```diff
@@ -31,11 +31,11 @@
 
         <activity
             android:name="com.termux.app.TermuxActivity"
-            android:label="@string/application_name"
-            android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
+	    android:label="@string/application_name"
+	    android:configChanges="orientation|screenSize|smallestScreenSize|density|screenLayout|uiMode|keyboard|keyboardHidden|navigation"
```
trygveaa: @jeansch: Is `uiMode` necessary to include here? It breaks automatically switching between light and dark theme with #1476. If I remove it, it switches automatically, and the terminals I have open remain open with all their contents.

jeansch: Hi, I don't know, I don't have much knowledge in android development. I just followed documentation link from #1259. 
I don't use the S10 anymore. I would suggest to give it a try, then PR if needed.

obot7:
I also have problem

app/src/main/AndroidManifest.xml
trygveaa: Ah, so copied from Samsungs documentation, okay. Unfortunately, it doesn't seem like Samsung has documented why it is specified, and I don't have any Samsung devices, so I can't test if it breaks if I remove it.

PR #1475, opened 2020-02-28
Join lines that fills the width when selecting urls

### LINKED ISSUE #313: Links on multiple lines cannot be copied
_From @Raphi111 on March 17, 2017 12:43_

I love the feature that allows to copy urls but I noticed that if an url is spanning on multiple lines (in weechat for example), termux only copy the first line.
I don't know if this really is a bug as there might not be any way to tell where the end of the url is.

_Copied from original issue: termux/termux-styling#14_

trygveaa:
Some terminal applications, like mutt and weechat, prints a newline at
the end of each line even if text is wrapped. This causes urls which are
wrapped to not be selectable in full.

By ignoring newlines when the text fills the entire width of the screen,
those urls can be selected. Many other terminal emulators do this as
well.

A drawback of this is that if a url happens to fill the width of the
screen, the url selection will include the first word of the next line,
but this doesn't happen that often so I think it's an okay tradeoff.

Fixes #313

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -754,7 +754,7 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
     }
 
     void showUrlSelection() {
-        String text = getCurrentTermSession().getEmulator().getScreen().getTranscriptText();
+        String text = getCurrentTermSession().getEmulator().getScreen().getTranscriptTextWithFullLinesJoined();
         LinkedHashSet<CharSequence> urlSet = extractUrls(text);
         if (urlSet.isEmpty()) {
             new AlertDialog.Builder(this).setMessage(R.string.select_url_no_found).show();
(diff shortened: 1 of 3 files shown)


trygveaa:
> A drawback of this is that if a url happens to fill the width of the
screen, the url selection will include the first word of the next line,
but this doesn't happen that often so I think it's an okay tradeoff.

What do you think of this? Alternatively I can make it configurable.

xeffyr:
> A drawback of this is that if a url happens to fill the width of the
screen, the url selection will include the first word of the next line,
but this doesn't happen that often so I think it's an okay tradeoff.

I wasn't able to reproduce that. Tested with file full of long URLs opened with `nano --softwrap`.

trygveaa:
It only happens when the url exactly fills the width of the terminal. You can reproduce it with this command:

```
printf 'http://%*s.com/\nshould-not-be-part-of-url\n' $(($COLUMNS-12)) | tr ' ' a
```

It will print one line with `http://aaa...aaa.com/` and the next with `should-not-be-part-of-url`. The terminal has no way of knowing if the second line should be part of the url or not, which means that in cases that it shouldn't be a part of the url, it will be wrong.

Issue #334, opened 2017-06-05
Selection aid?


noncom:
Working with a terminal often involves selecting some text, and it is not really comfortable when the terminal is fullscreen and the text touches the screen edges. Including the edge characters into the selection becomes a rather non-trivial task... Can't help, but the selection aid from JuiceSSH springs into mind by itself... could we have something like that in Termux?

I'd even pay for an addon to do that.. or even an addon that does some additional higher-level text and/or input history manipilation. Like, automatically extracting paths and numbers from the available text body would be super-useful on a touch-driven device...

xeffyr:
Closing as already implemented.

PR #1477, opened 2020-02-29
Set orientation of HandleView in show

trygveaa:
When you hold down on a word that starts or ends at the edge of the
screen, the handle will appear outside of the screen. This happens
because the orientation was only switched when the handle is dragged, so
when it is shown it just used the same orientation as it had for the
last selection.

Relates to #334, but not sure if it fixes it completely.

Before:
image

After:
image

This is when holding down on the word "Working".

### DIFF
--- a/terminal-view/src/main/java/com/termux/view/TerminalView.java
+++ b/terminal-view/src/main/java/com/termux/view/TerminalView.java
@@ -973,7 +973,7 @@ public final class TerminalView extends View {
             return mContainer.isShowing();
         }
 
-        private void checkChangedOrientation() {
+        private void checkChangedOrientation(int posX) {
             if (!mIsDragging) {
                 return;
             }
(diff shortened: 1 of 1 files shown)


terminal-view/src/main/java/com/termux/view/TerminalView.java
```diff
@@ -1395,9 +1392,9 @@ public void updatePosition() {
                 return;
             }
 
-            mStartHandle.positionAtCursor(mSelX1, mSelY1);
+            mStartHandle.positionAtCursor(mSelX1, mSelY1, false);
 
-            mEndHandle.positionAtCursor(mSelX2 + 1, mSelY2); //bug
```
trygveaa: I don't know why it said "bug" here, but as far as I can see, there is no bug here, so I removed the comment.

PR #1490, opened 2020-03-12
Grab fragment when extracting URLs for selection

quite:


fornwall:
Thanks! This will be part of the 0.94 version of the app.

mwalimu-prog:
Thank you. Its updated version of the app

PR #1562, opened 2020-05-17
Android Studio 3.6.3

hannesa2:
simply to be up do date


hannesa2:
It's a duplicate of https://github.com/termux/termux-app/pull/1559

PR #1566, opened 2020-05-17
Espresso tests

hannesa2:
It's a follow up from https://github.com/termux/termux-app/pull/474 @stephengroat

Additional :
* rebased to master 
* squashed all commits
* Android X
* CI Espresso test

### DIFF
--- a/app/src/androidTest/java/com/termux/app/TermuxTest.java
+++ b/app/src/androidTest/java/com/termux/app/TermuxTest.java
@@ -50,8 +50,8 @@ public class TermuxTest {
 
     @Test
     public void changeText_sameActivity() {
-        // wait during 15 seconds for a view
-        onView(isRoot()).perform(waitId(R.id.terminal_view, TimeUnit.SECONDS.toMillis(15)));
+        // wait for a view
+        onView(isRoot()).perform(waitId(R.id.terminal_view, TimeUnit.SECONDS.toMillis(300)));
 
         // Type text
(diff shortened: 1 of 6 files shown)


hannesa2:
@stephengroat 
With `Run` it works on my machine

image

but with `./gradlew cAT` on my machine, like in CI, it doesn't work

app/src/androidTest/java/com/termux/app/TermuxTest.java
```diff
@@ -0,0 +1,128 @@
...
+    @Before
+    public void initValidString() {
+        // Specify a valid string.
+        mStringToBetyped = "ls -lah\n";
+    }
+
+    @Test
+    public void changeText_sameActivity() {
+        // wait during 15 seconds for a view
+        onView(isRoot()).perform(waitId(R.id.terminal_view, TimeUnit.SECONDS.toMillis(15)));
```
stephengroat, line 54: try a much longer wait time, the view might be taking a long time to populate on CI

```suggestion
        onView(isRoot()).perform(waitId(R.id.terminal_view, TimeUnit.SECONDS.toMillis(300)));
```

hannesa2: I tried it, but as is looks like Espresso has a build-in-function to not wait too long

> Error performing 'wait for a specific view with id <2131034142> during 300000 millis.' on view 'is a root view.'.

Independent of this, when a view doesn't appear within 15/30 seconds, something other is wrong. 
Why it works with `Run` in IDE, and not with [command line](https://github.com/termux/termux-app/pull/1566#issuecomment-629787966) ?

PR #1603, opened 2020-06-15
Begin conversion to Kotlin

bartekpacia:
We've (@larson-carter) configured Kotlin for the app and converted the `BellUtil` from Java to Kotlin.

What do you think?


buttaface:
While Kotlin is better, this app isn't updated much, so I don't think it's worth the effort of a rewrite.

donno2048 (commented):
Kotlin is better but does it really worth it?

Nerd00F:
What do you think @agnostic-apollo should we upgrade to kotlin?

agnostic-apollo:
Sure, why not. Don't let my never having coded in Kotlin stop this movement. It would just mean termux development stopping for who knows how long :p

agnostic-apollo:
On a more serious note, I am not totally against the conversion to kotlin despite me not knowing the language currently.

It could be done eventually, at least parts of it. Currently, as I said, I don't know the language and if not me, other termux maintainers would still need to know the language deeply to actually rewrite so that the app improves and doesn't get filled with bad design or bugs just for the sake of conversion. If outside contributor rewrites the app but the main maintainers don't know the language, then that it also an issue. How would they review, maintain and update, if they don't understand the lang. I don't know the status of other maintainer's kotlin knowledge.

Moreover, there are lot of issues pending that are far more important to spend time on than language conversion. The users may not notice much change anyways for many if not most things.

The `terminal-view` and `terminal-emulator` code is quite complex and I don't understand much of it and I haven't tried to understand either, but it would probably takes weeks or months of research, so can't really rewrite something you don't understand fully.

There are still [more java devs than kotlin devs](https://insights.stackoverflow.com/survey/2021#section-most-popular-technologies-programming-scripting-and-markup-languages), so higher chance of contributions from others too. Java also would have much better language support (tutorials, etc).

I'll eventually need to learn it, but may not be soon.

Nerd00F:
So, do we agree to keep it in java?
Maybe close this issue then...

agnostic-apollo:
For now, yes

Nerd00F:
Ok, then close the PR

PR #1610, opened 2020-06-21
export COLORTERM=truecolor

obfusk:
See https://gist.github.com/XVilka/8346728#true-color-detection


xeffyr:
Do you know **any well-known** terminal which exports that variable ?  We are not going to overload environment by *umnecessary* variables and ones which are just required to "fix" someone's programs. As far I know, lack of that variable didn't caused any issues with our packages.

Yes, our terminal supports truecolor, but its usability actually depends on user's display. There are people who uses Termux on E-ink devices where truecolor doesn't make any sense.

xeffyr:
> Do you know any well-known terminal which exports that variable ?

Ah, seems latest versions of S-lang use it.

PR #1627, opened 2020-07-04
Fix selection rendering with alternate cursor styles

kdrag0n:
There is currently a bug where selection rendering is broken if the active cursor shape is anything other than the default solid box. Selected text is normally highlighted by effectively rendering a cursor over all of the characters in the selection region, but if the cursor is a bar, the resulting selection highlight is too narrow to cover the full width of the selection. Similarly, if the cursor is an underline, all of the selected text will be underlined instead of highlighted.

To fix this issue, treat selections different from cursors in the rendering logic and force the renderer to always use the block cursor style for rendering selections. That way, we get correct behavior regardless of what the current cursor shape is.

Example:
Before:
image
After:
image


xeffyr:
Thanks.

PR #1628, opened 2020-07-04
Invert selected text instead of highlighting

kdrag0n:
Highlighting text in the terminal often makes it hard to read, which can be problematic for users who want to adjust or review selections before copying them. For example, the default theme makes white and green text hard to read on its light gray selection background, and there are plenty of other themes where the choice of text and cursor colors would hinder selection readability.

To fix this issue and make selected text more legible in nearly all combinations of colors, invert selected text instead of highlighting it. This is more common among terminal emulators anyway:

- Invert: xterm, fbcon, kitty, Konsole, Alacritty, Tilix, gnome-terminal (7)
- Highlight: Termux, Terminal.app, iTerm2, Windows Terminal (4)

### DIFF
--- a/terminal-view/src/main/java/com/termux/view/TerminalRenderer.java
+++ b/terminal-view/src/main/java/com/termux/view/TerminalRenderer.java
@@ -117,11 +117,10 @@ public final class TerminalRenderer {
                     } else {
                         final int columnWidthSinceLastRun = column - lastRunStartColumn;
                         final int charsSinceLastRun = currentCharIndex - lastRunStartIndex;
-                        int cursorColor = (lastRunInsideCursor || lastRunInsideSelection) ? mEmulator.mColors.mCurrentColors[TextStyle.COLOR_INDEX_CURSOR] : 0;
-                        int cursorStyle = lastRunInsideSelection ? TerminalEmulator.CURSOR_STYLE_BLOCK : cursorShape;
+                        int cursorColor = lastRunInsideCursor ? mEmulator.mColors.mCurrentColors[TextStyle.COLOR_INDEX_CURSOR] : 0;
                         drawTextRun(canvas, line, palette, heightOffset, lastRunStartColumn, columnWidthSinceLastRun,
                             lastRunStartIndex, charsSinceLastRun, measuredWidthForRun,
-                            cursorColor, cursorStyle, lastRunStyle, reverseVideo);
+                            cursorColor, cursorShape, lastRunStyle, reverseVideo || lastRunInsideSelection);
                     }
                     measuredWidthForRun = 0.f;
                     lastRunStyle = style;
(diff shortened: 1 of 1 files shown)


kdrag0n:
Example of selected dmesg text:
Before:
image
After:
image

terminal-view/src/main/java/com/termux/view/TerminalRenderer.java
```diff
@@ -117,11 +117,10 @@ public final void render(TerminalEmulator mEmulator, Canvas canvas, int topRow,
                     } else {
                         final int columnWidthSinceLastRun = column - lastRunStartColumn;
                         final int charsSinceLastRun = currentCharIndex - lastRunStartIndex;
-                        int cursorColor = (lastRunInsideCursor || lastRunInsideSelection) ? mEmulator.mColors.mCurrentColors[TextStyle.COLOR_INDEX_CURSOR] : 0;
-                        int cursorStyle = lastRunInsideSelection ? TerminalEmulator.CURSOR_STYLE_BLOCK : cursorShape;
+                        int cursorColor = lastRunInsideCursor ? mEmulator.mColors.mCurrentColors[TextStyle.COLOR_INDEX_CURSOR] : 0;
                         drawTextRun(canvas, line, palette, heightOffset, lastRunStartColumn, columnWidthSinceLastRun,
                             lastRunStartIndex, charsSinceLastRun, measuredWidthForRun,
-                            cursorColor, cursorStyle, lastRunStyle, reverseVideo);
+                            cursorColor, cursorShape, lastRunStyle, reverseVideo || lastRunInsideSelection);
```
trygveaa, line 123: Aren't you effectively reverting your last PR (#1627) here?

kdrag0n: This reverts the cursor style part but it still depends on the `insideSelection` logic from #1627. I chose to submit this separately since I wanted to fix the existing behavior first before changing anything.

trygveaa: Ah, okay, and since you're drawing the selection by reversing video instead of by using the cursor it isn't necessary anymore.

trygveaa:
> Invert: kitty

Kitty uses a constant background color, not inverting. However it also uses a constant foreground color for selections, which removes the problem you're describing. Personally I prefer this because it's more clean (or maybe it's just because it's what I'm used to at this point).

kdrag0n:
My bad; the kitty color scheme I'm using has the selection colors set to inverted white text, so I classified it as "invert."

I'm personally not a fan of constant colors because while it looks cleaner in some cases, sometimes I find it jarring for colors to disappear when I select text in kitty. It's particularly annoying when viewing something that relies on color differentiation to be clear, e.g. [diff-so-pretty](https://github.com/so-fancy/diff-so-fancy) output or folder/file classification in `ls`.

(sorry for the closing and reopening spam, I accidentally clicked "close" while scrolling)

Lzhiyong:
@xeffyr There is still a problem with inverting color in force dark mode on Android 10,  
the background color of text selection is not obvious

[
Screenshot_2020-08-08-16-57-15-819_com termux
](url)

xeffyr:
@Lzhiyong Because this improvememt was released in v0.96 which is still not available for installation.

I have dark mode on and also Android 10 - everything works, however I use custom Termux build with all latest features available.

Lzhiyong:
@xeffyr thank you for your reply, I think you can add a dark theme, or use the daynight theme, on Android 9, even if dark mode is turned on, the background color of DrawerLayout and dialog is still white, force dark mode requires Android 10 or higher.

xeffyr:
@Lzhiyong Termux already has dark theme. Enable it manually as shown in https://github.com/termux/termux-app/issues/1681#issuecomment-667298922.

PR #1685, opened 2020-08-04
move sdk version configs to gradle.properties

cn00:
move sdk version configs to gradle.properties to avoid modify each build.gradle every time

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,7 +15,7 @@
 org.gradle.jvmargs=-Xmx2048M
 android.useAndroidX=true
 
-minSdkVersion=21
-targetSdkVersion=29
+minSdkVersion=24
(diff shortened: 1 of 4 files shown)


gradle.properties
```diff
@@ -14,3 +14,8 @@
 # org.gradle.parallel=true
 org.gradle.jvmargs=-Xmx2048M
 android.useAndroidX=true
+
+minSdkVersion=21
+targetSdkVersion=29
```
xeffyr, line 19: Use SDK 28 here as 29 breaks **everything**!

xeffyr: Termux is not Android 10 compatible application. See https://github.com/termux/termux-app/issues/1072.

Issue #1695, opened 2020-08-07
Command output is not always displayed


dfffffff:
**Problem description**
Command output is not always displayed when the command is also making the terminal exiting.
Only "[Process completed (code X) - press Enter]" is displayed.

**Steps to reproduce**
Make a small C program like:
``` C
#include <unistd.h>
int main(void)
{
    write(1, "########\n", 9);
    return 123;
}
```
Execute it with "exec ./smallprog" in a terminal.

**Expected behavior**
"########" should be always displayed.

**Additional information**
Some race condition occur, most of the time the output is displayed.
Sounds like sometime the process exit is processed before all buffered output is processed.
Also occur with shell script output.
Happen on stdout or stderr.
The bug is more apparent with Termux:Widget since the terminal is closed after execution, sometime on command error nothing is displayed.

xeffyr:
This is not race condition. Terminal output is being processed only when process is not terminated.

I have made a PR to change that behavior: https://github.com/termux/termux-app/issues/1695.

xeffyr:
Part of v0.99 release.

dfffffff:
The problem still arises but less frequently with v0.99.

pierro78:
still arises with v0.108 in aterm/xfce4-terminal ...

RalfWerner:
@pierro78 Can you give more info about the device and which keyboard you use?
If it is a _Samsung device_ could possibly help `enforce-char-based-input=true`  in `~/termux/termux.properties`

pierro78:
@RalfWerner it s an xperia 5 2019 ( https://www.reddit.com/r/termux/comments/mfpwry/termux_tripled_the_speed_of_this_12yo_laptop/gsu67dh/?utm_source=reddit&utm_medium=web2x&context=3 ) 

I am using tigervncviewer or rdp client from my laptop (not from the phone)

shall I fill a different issue in the termux/x11-packages repo as it is related to packages in that repo ?

thanks

RalfWerner:
I have no _Xperia_ device and **can not repeat** your problem. since a week I have an Android 11 update for my _Samsung_ device. Does your problem occur _only_ when using your laptop (windows X server) or in _standalone_ mode too?
> shall I fill a different issue in the termux/x11-packages repo

You should do that only if you have _**what your problem is triggered**_

LeapofAzzam-backup:
I get that problem but now i have 2 termux in recent apps

PR #1696, opened 2020-08-07
terminal emulator: flush remaining process output data when terminating session

xeffyr:
See https://github.com/termux/termux-app/issues/1695.


xeffyr:
Merging as I have not found any side effects.

PR #1715, opened 2020-08-15
Added ApkUtils.kt for APK package Management

imprakharshukla:
This contains the basic implementation of APK management including downloading, installing, uninstalling and listing out the installed packages.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,6 +1,12 @@
+import java.security.DigestInputStream
+import java.util.zip.ZipEntry
+import java.util.zip.ZipInputStream
+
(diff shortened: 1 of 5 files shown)


xeffyr:
I have applied some fixes, so now PR works for *installing* and *uninstalling* packages. However it has a one serious issue: you can't install more than one package per session. To install 2 or more packages in sequence, I have to force-close application after each installed apk with `pkg add`. Otherwise, apk installation dialog simply does not appear.

ThoonsanEain (commented):
time for me it would like

PR #1734, opened 2020-08-29
Android 10 APK Package Management

imprakharshukla:
**Implemented Play Store Downloading** - The current implementation is not at all user friendly (and doesn't properly work) for multiple package downloads at one (pkg install htop vim...). This stands true if the links are opened in separate tabs of a browser if Play Store is not present on the device and in different Play Store windows which doesn't work that way, as well.

Possible implementation - 

> Start a transparent activity and monitor the lifecycle so we can actually present a dialog before directly throwing either the Play Store or a browser window in the face of the user and then monitor the onResume LC component to get when the user returns to the app, then present the dialog for the second package.

 I didn't wanted to code this much only to get it scraped out. So let me know if you guys are comfortable with this approach or have something better.


K8wan:
Hi

K8wan (approved):

xeffyr:
Added in https://github.com/termux/termux-app/commit/8d83a39a852cdfdfd7d86ae55bbf52f1f47f3bd0.

***

> This stands true if the links are opened in separate tabs of a browser if Play Store is not present on the device

If Play Store is not present on device, then **it shouldn't be used as download source**.

> in different Play Store windows which doesn't work that way, as well.

One Play Store window is needed only. Next package is being installed once window closed, e.g. by pressing button "back".

xeffyr:
Also:
1. Why using hardcore cache directory path? Android doesn't have API for programatically retrieving its path?
2. Why not checking for `source` equals `play-store` but checking whether it is empty or null?

PR #1764, opened 2020-09-18
Fix RunCommandService crash and foreground issue

agnostic-apollo:
Since `startForeground()` function is not being called, occasionally termux is killed by android with the following exception

```
I/ActivityTaskManager: START u0 {flg=0x10000000 cmp=com.termux/.app.TermuxActivity} from uid 10138
W/ActivityTaskManager: Background activity start [callingPackage: com.termux; callingUid: 10138; isCallingUidForeground: false; isCallingUidPersistentSystemProcess: false; realCallingUid: 10138; isRealCallingUidForeground: false; isRealCallingUidPersistentSystemProcess: false; originatingPendingIntent: null; isBgStartWhitelisted: false; intent: Intent { flg=0x10000000 cmp=com.termux/.app.TermuxActivity }; callerApp: ProcessRecord{75609ea 14280:com.termux/u0a138}]
D/AndroidRuntime: Shutting down VM
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.termux, PID: 14280
    android.app.RemoteServiceException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{f94f200 u0 com.termux/.app.RunCommandService}
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1945)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
I/ActivityManager: Showing crash dialog for package com.termux u0
W/ActivityManager: Bringing down service while still waiting for start foreground: ServiceRecord{d0f1bbf u0 com.termux/.app.RunCommandService}
I/ActivityManager: Crashing app skipping ANR: ProcessRecord{75609ea 14280:com.termux/u0a138} Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{d0f1bbf u0 com.termux/.app.RunCommandService}
```  
I have added calls to  `startForeground()` in both `onCreate()` and `onStartCommand()` and have tested the in android 10 emulator and the fix seems to be working with no crashes yet. I even tested under low free RAM conditions in order to slow down the phone but termux still didn't crash. Of course, it could potentially still crash because the function may not be called in under `5s` in all circumstances since its a framework issue. More details [here](https://stackoverflow.com/questions/44425584/context-startforegroundservice-did-not-then-call-service-startforeground/53402038). Disabling battery optimization is likely the better way but this also works and is the recommended way.

Adding `android.permission.SYSTEM_ALERT_WINDOW` permission to fix the foreground issue seems to be the reasonable way to go since all the other [exceptions](https://developer.android.com/guide/components/activities/background-starts#exceptions) to the rule do not really apply to termux, and without it, it greatly reduces usability of both the `RUN_COMMAND` intents and `Termux:Tasker` plugin. If a user has already set a certain command to be run in a foreground session from a background task like from tasker, it should run in foreground without user intervention. The hack that currently works for running termux commands from tasker is to do a `sleep 1; am start --user 0 -n com.termux/com.termux.app.TermuxActivity` after running the `Termux:Tasker` plugin or TermuxCommand action, since tasker already has requested  `android.permission.SYSTEM_ALERT_WINDOW` permission and also has an accessibility service, so it can actually start termux activity from its own background execution service, but termux itself can't from its own service. I haven't currently added a way for the user to request the permission from within termux, let me know if that is even required.

Let me know if there are any other issues. Thanks

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -21,6 +21,7 @@
     <uses-permission android:name="android.permission.VIBRATE" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS"/>
+    <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
 
     <application
         android:extractNativeLibs="true"
--- a/app/src/main/java/com/termux/app/RunCommandService.java
+++ b/app/src/main/java/com/termux/app/RunCommandService.java
@@ -87,6 +87,7 @@ public class RunCommandService extends Service {
         runStartForeground();
     }
 
+    @Override
     public int onStartCommand(Intent intent, int flags, int startId) {
         // Run again in case service is already started and onCreate() is not called
         runStartForeground();
--- a/app/src/main/java/com/termux/app/RunCommandService.java
+++ b/app/src/main/java/com/termux/app/RunCommandService.java
@@ -1,12 +1,18 @@
 package com.termux.app;
 
+import android.app.Notification;
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
 import android.app.Service;
+import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
 import android.os.Binder;
 import android.os.Build;
 import android.os.IBinder;
 import android.util.Log;
+
+import com.termux.R;
 
 import java.io.File;
 import java.io.FileInputStream;
@@ -50,6 +56,9 @@ public class RunCommandService extends Service {
     public static final String RUN_COMMAND_WORKDIR = "com.termux.RUN_COMMAND_WORKDIR";
     public static final String RUN_COMMAND_BACKGROUND = "com.termux.RUN_COMMAND_BACKGROUND";
 
+    private static final String NOTIFICATION_CHANNEL_ID = "termux_run_command_notification_channel";
+    private static final int NOTIFICATION_ID = 1338;
+
     class LocalBinder extends Binder {
         public final RunCommandService service = RunCommandService.this;
     }
@@ -61,7 +70,15 @@ public class RunCommandService extends Service {
         return mBinder;
     }
 
+    @Override
+    public void onCreate() {
+        runStartForeground();
+    }
+
     public int onStartCommand(Intent intent, int flags, int startId) {
+        // Run again in case service is already started and onCreate() is not called
+        runStartForeground();
+
         if (allowExternalApps() && RUN_COMMAND_ACTION.equals(intent.getAction())) {
             Uri programUri = new Uri.Builder().scheme("com.termux.file").path(intent.getStringExtra(RUN_COMMAND_PATH)).build();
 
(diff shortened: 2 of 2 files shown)


Grimler91:
Hey, thanks! Could you maybe split the commit into two or three parts to keep the git history easier to follow? (For example fixing the crash in one, improving RUN_COMMAND functionality in one and adding the PREFIX/HOME expanding in one) 

Will test this out on my devices!

agnostic-apollo:
Sure. Out at the moment, will do it when I get back.

agnostic-apollo:
Hey, these commits now should be the same as the original single commit. Let me know if its working fine on your devices. I don't have a physical android 8+ device.

And forgot to mention, to produce the crash on `v0.99` of termux, exit all termux sessions, remove termux from recents, wait a few minutes so that likely the timeout for the [allowlist](https://developer.android.com/about/versions/oreo/background#services) passes, then use Tasker `TermuxCommand` function in `Tasker Function` action to run a command, and of course cross your fingers. ;) For arguments to work, use [Tasker v5.9.3 sub release](https://www.reddit.com/r/tasker/comments/i42itd/dev_tasker_593_any_sensor_contact_via_app_dialogs/g0g871j?utm_source=share&utm_medium=web2x&context=3). `am` command should work too, but in that case if I remember correctly errors were shown in `stderr` and logcat had entries like `Background start not allowed: service Intent { act=com.termux.RUN_COMMAND cmp=com.termux/.app.RunCommandService (has extras) } to com.termux/.app.RunCommandService from pid=9616 uid=10134 pkg=com.android.shell startFg?=false`.

Grimler91:
@agnostic-apollo I have not been able to reproduce the crash on my android 10 (LineageOS) device, tasker reports another error:
```
19.08.55/E FIRE PLUGIN: Termux / com.twofortyfouram.locale.intent.action.FIRE_SETTING: 6 bundle keys
19.08.55/E Termux: plugin comp: com.termux.tasker/com.termux.tasker.FireReceiver
19.08.55/E add wait type Plugin1 time 10
19.08.55/E add wait type Plugin1 done
19.08.55/E add wait task
19.09.05/E Error: null
``` 
This PR does not change anything on that device, but merging since it fixes the crash on your device.

agnostic-apollo:
Well, [u/DutchOfBurdock](https://www.reddit.com/r/tasker/comments/i42itd/dev_tasker_593_any_sensor_contact_via_app_dialogs/g0jv609?utm_medium=android_app&utm_source=share&context=3
) was the one who initially reported the issue on his device which I confirmed on android 10 avd, so the issue does exist for others. I personally use LG G5, Android 7.0 which obv doesn't have the issue.

hmm, interesting that you can't reproduce it. LineageOS doesn't seem to have looser restrictions for background stuff according to this [report](https://github.com/ukanth/afwall/issues/1027). I have already mentioned this, but just confirming that have you disabled termux battery optimizations, if you disable those, then you are likely not going to get a crash.

The entries you are getting by tasker are from its `ExecuteService` which has the tag `E`, they are not errors, just entries that the  service is adding a wait timeout for 10s for the plugin action.

And thanks for the merge.

xeffyr:
Part of [v0.100] release.

agnostic-apollo:
Thanks. Congrats on the `0.100`th though. And thank you for everything you guys have done.

Grimler91:
> confirming that have you disabled termux battery optimizations, if you disable those, then you are likely not going to get a crash

Ah, yes, I was running without battery optimization, so that is probably the reason.. Should have read the instructions more carefully..

> The entries you are getting by tasker are from its ExecuteService which has the tag E, they are not errors, just entries that the service is adding a wait timeout for 10s for the plugin action.

Alright, I get a timeout then without my test script being executed, thanks for clarifying that!

agnostic-apollo:
Glad you managed to get a crash ;)

You are welcome.

Davidrekrut:
today is the best day

TomJo2000:
> today is the best day

Because a random bug fix from 3 years ago?

Davidrekrut:
Мне нравится

Davidrekrut:
Великолепное чудо

PR #1782, opened 2020-10-05
Update for APK package management

imprakharshukla:
* Changed the hardcoded cache directory to the android alternative (pointing to /data/user/0 which is a symlink to /data/data).
* Implemented the transparent activity mechanism for downloading multiple apps through the Play Store (which just works as if the intent was added to the back-stack)
* Now we check deliberately check for source (external for ext. repo and play-store for the Play Store) and have a fallback.
* Fixed a few bugs.


xeffyr:
`android-10` branch [has been synced](https://github.com/termux/termux-app/commit/3497119c71649d830f950b8c4aa5ca2f50a251f6) with these changes. Don't merging since android-10 uses `master` as base branch and not the `android-package-apks`.

PR #1865, opened 2020-12-30
Don't link directly to Google Play

landfillbaby:
allows the user to choose to open the link to install Termux:Styling via F-Droid too


xeffyr:
I can merge this and release v0.105. However there is still a problem: not all users using F-Droid have its app installed.

I guess more correct solution would be a direct linking to F-Droid page. It is pointless to send user to Play Store or even give such option if APK is distributed by F-Droid. This is all due to signatures, users may not mix add-ons and app installed from different sources.

landfillbaby:
yeah i guess since it's not getting updated on google anymore we may as well  just use the f-droid url directly

Issue #1333, opened 2019-11-02
Termux ignores Ctrl-Space on Chromebook


snogglethorpe:
**Problem description**
On a Chromebook (with a hardware keyboard), typing Ctrl-Space does nothing.

This particular key-combination is traditionally bound to the set-mark command in Emacs, and so is _very_ heavily used by most users.  set-mark is also available via Ctrl-2 (and of course may be bound by the user to something else), but I think for many users, this is inconvenient, (1) because Ctrl-2 is a bit harder to type, (2) because the documentation says to use "C-SPC", and (3) most importantly,  because many users have strongly developed muscle memory telling them to hit Ctrl-Space for this command.

This may be related to the fact that in most context on ChromeOS, Ctrl-Space toggles the current input method.  However, this is not useful for Termux, as ChromeOS input methods are ignored (whether _this_ is a bug or not is another question).

Other apps in ChromeOS (e.g. the ssh app) successfully use Ctrl-Space for their own use (ssh just passes it through the the connection), so it's definitely possible to override ChromeOS's default behavior for this key-combination (of course it's possible that Android apps in ChromeOS have unique constraints).

**Steps to reproduce**
Open Termux, run some program where Ctrl-Space should have an effect (e.g., Emacs), type Ctrl-Space.  Nothing happens.

**Expected behavior**
Ctrl-Space should result in some sort of key sequence on the terminal.  I think traditionally, Ctrl-Space yielded the same result as Ctrl-2 / Ctrl-Shift-@, but basically anything would be OK as long as it's possible for the user program to see the event.

**Additional information**
Updatable packages:
apt/stable 1.4.9-19 i686 [upgradable from: 1.4.9-15]
bash/stable 5.0.11 i686 [upgradable from: 5.0.9]
binutils/stable 2.33.1 i686 [upgradable from: 2.32-5]
ca-certificates/stable 20191016 all [upgradable from: 20190515]
clang/stable 9.0.0 i686 [upgradable from: 8.0.1]
command-not-found/stable 1.42 i686 [upgradable from: 1.39]
coreutils/stable 8.31-8 i686 [upgradable from: 8.31-7]
curl/stable 7.66.0-1 i686 [upgradable from: 7.65.3-6]
dos2unix/stable 7.4.1 i686 [upgradable from: 7.4.0-1]
emacs/stable 26.3-1 i686 [upgradable from: 26.3]
findutils/stable 4.7.0 i686 [upgradable from: 4.6.0-4]
[...]

PR #1869, opened 2021-01-01
Add workaround property to fix ctrl+space on devices where this does not work

Grimler91:
Should fix (or well, allow user to set a property and thereby fix) https://github.com/termux/termux-app/issues/1333. 

This is a problem in for example emacs when a user uses a hardware keyboard and wants to mark text by pressing ctrl+space. On some devices pressing ctrl + space currently does nothing, but with PR users that want this functionality can add `ctrl-space-workaround=true` to `~/.termux/termux.properties` and then run `termux-reload-settings` to make ctrl+space work as expected. 

Confirmed to work on my phone.


Grimler91:
No objections so merging

PR #1904, opened 2021-01-23
app: enable split apk builds

evelikov:
Currently we build a single APK which handles the four supported ABIs.

Therefore each user, downloads 50-75% more than they need - adding
towards both client/server-side network as well as device storage.

Use a split APK approach - it costs nothing from build and server-side
storage POV.

Note: We're removing ndk:abiFilters - they're incompatible/superseded by
the splits:abi:include list.

Signed-off-by: Emil Velikov <emil.l.velikov@gmail.com>

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -25,9 +25,14 @@ android {
                 cFlags "-std=c11", "-Wall", "-Wextra", "-Werror", "-Os", "-fno-stack-protector", "-Wl,--gc-sections"
             }
         }
+    }
 
-        ndk {
-            abiFilters 'x86', 'x86_64', 'armeabi-v7a', 'arm64-v8a'
+    splits {
+        abi {
+            enable true
+
+            reset ()
+            include 'x86', 'x86_64', 'armeabi-v7a', 'arm64-v8a'
         }
 
     }


evelikov:
The failure is due to the missing `app-debug.apk` - with the split APK we have `app-$ABI-debug.apk`.
I would update the CI or keep a single APK for debug-only builds.

Let me know what you prefer.

xeffyr:
Splitapk builds are good, but wondering how F-Droid will handle them.

evelikov:
Pretty sure it can handle them since I have both official (VLC) and third party repo apps (Bromium) where this is used.

Looking at existing [termux](https://gitlab.com/fdroid/fdroiddata/-/blob/38a4764ae53bd2119a64e92f2cecbec478a046e1/metadata/com.termux.yml) recipe - the output file is magically detected. While in the [VLC](https://gitlab.com/fdroid/fdroiddata/-/blob/147828cc5dde6eff5a28bae01403eb22cf70ec4c/metadata/org.videolan.vlc.yml) case a separate build permutations are added.

VLC has custom combile.sh script, builds external third party tools, different versions per ABI, etc - so such complexity is unlikely for termux. Most likely just an output line will be just fine. Untested example:

`output: build/outputs/apk/release/app-*.apk`

evelikov:
If there's anything I can do to move this forward - please let me know.

LucioAmely:
Will split apk builds be available in version 0.109?

xeffyr:
@LucioAmely Split apks are disabled on F-Droid side, see https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/com.termux.yml#L673. I can re-enable it for Github Actions debug builds  but no more.

evelikov:
@xeffyr out of sheer luck I saw the [commit reverting](https://github.com/termux/termux-app/commit/ce5e9a90421da9aaae06c6e150afe3e6c09df944) this. Can you please @ people for the future?

Also do you have a link with more details about the issue they're seeing - be that a gitlab/discord/irc/other? The commit which adds this is quirk does not mention anything - so it's basically useless.

Grimler91:
@evelikov this issue in their tracker seem to have the most info: https://gitlab.com/fdroid/fdroiddata/-/issues/2115

evelikov:
You are a star, thank you.

It does sound completely bonkers - so they support split apks (VLC, telegram, etc) but it has to be completely separate targets... Even though the build as-is produces all the binaries at once.

elsaxo:
Trd

Tanisoraon (commented):
Hi

licaon-kter:
Is this dead?

licaon-kter:
@kyehn F-Droid support multiple APKs for a while now...

Issue #507, opened 2017-12-13
Fullscreen / immersive mode option is gone


kracejic:
Hi, 

first of all, thanks to all working on this amazing android application. It is really great to being able to run my complete work environment on my phone. 

I have noticed that fullscreen option was missing, I have looked at recent commits and it looks that it was removed from sources. Is there other way how to achieve the same result? Or could we keep it for users who use it? I for myself can live with its quirks, but without it, I get that annoying bar with clock and virtual buttons on the screen taking precious pixels from the important content.

Is there a way how to get it back?
Thanks!

Quasic:
It worked for you? It never changed anything on my LG Lollipop 5.1.1 arm device. I was interested in the feature at one point, but decided that stuff comes in useful when the unexpected happens.

fornwall:
@kracejic Just to check, you're running with a hardware keyboard right?

kracejic:
@Quasic Yes, it worked for me. Both on Android 6 and on 7. Sometimes I had to toggle software keyboard. Like I wrote, it had some quirks. :) But for me, it was still well worth it. I don't like wasted pixels, especially on the small screens.

@fornwall Yes, I am using HW keyboard, either bluetooth one, or USB one through USB adapter. Thanks to this great app I do not see reason to buy notebook anymore, since I live mostly in shell anyway.

Quasic:
Toggling the keyboard and extra keys never worked for me, but I wouldn't be opposed to its reintroduction as a beta feature or something.

kracejic:
@Quasic That would be very nice.

gushmazuko:
Me too fullscreen option is gone.
Device: Nexus 6, Android 7.1.1

kanreki:
Me too, starting today. HP Chromebook 13, running Android apps on the Beta channel of Chrome OS

sdrausty:
Full screen mode works well in version https://f-droid.org/repo/com.termux_56.apk https://f-droid.org/repo/com.termux_56_src.tar.gz downgrading should work, but it doesn't out-of-the box.  So reinstalling is the only option at present for those that do not build Termux to get immersive mode.

sdrausty:
[...]

PR #1911, opened 2021-01-28
Implement full screen and add FullScreenWorkAround to fix ExtraKeysView

dkramer95:
## FullScreen Mode
Addresses: https://github.com/termux/termux-app/issues/507

Just setting full screen flag causes issue w/ ExtraKeysView:

`FullScreenWorkAround` class was added to help fix issues across various devices so that ExtraKeysView displays properly:

Example of `termux.properties` config for full screen w/ work-around:
Screen Shot 2021-01-27 at 6 42 02 PM

~~`fullscreen-workaround-method = 1` tested w/ Pixel (Emulator)
`fullscreen-workaround-method = 2` tested w/ Samsung S7 and OnePlus 6T
`fullscreen-workaround-method = 3` tested w/ Sony Xperia XZ~~

Some trial and error may be needed for your device.. Another workaround may be needed if the above do not work.

Feedback and testing appreciated!

### DIFF
--- a/app/src/main/java/com/termux/app/FullScreenWorkAround.java
+++ b/app/src/main/java/com/termux/app/FullScreenWorkAround.java
@@ -1,0 +1,136 @@
+package com.termux.app;
+
+import android.content.Context;
+import android.graphics.Rect;
+import android.util.Log;
+import android.view.View;
+import android.view.ViewGroup;
+import android.view.ViewTreeObserver;
+
+import static com.termux.terminal.EmulatorDebug.LOG_TAG;
+
+/**
+ * Work around for fullscreen mode in Termux to fix ExtraKeysView not being visible.
+ * This class is derived from:
+ * https://stackoverflow.com/questions/7417123/android-how-to-adjust-layout-in-full-screen-mode-when-softkeyboard-is-visible
+ * and has some additional tweaks
+ * ---
+ * For more information, see https://issuetracker.google.com/issues/36911528
+ */
+public class FullScreenWorkAround {
+    /**
+     * TODO add documentation so user can know available options and know they may need to experiment
+     * for their specific device
+     */
+    private static final int WORK_AROUND_METHOD_1 = 1;
+    private static final int WORK_AROUND_METHOD_2 = 2;
+    private static final int WORK_AROUND_METHOD_3 = 3;
+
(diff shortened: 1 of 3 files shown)


xeffyr:
Just curious: is there a way to have only *one workaround* suitable for all devices? Having different solutions for different device types selectable from the config is kinda weird.

dkramer95:
Ya between the various devices I tested there was inconsistent behavior in the way usable view space was being calculated with extra keys + status bar. Not sure why that is, so a WIP solution for now, but ideally it would "just work" for all devices w/o needing to specify that.

xeffyr:
@dkramer95 ConnectBot provides a similar widget (extra keys) for its terminal. Not sure if it has same issue as Termux extra keys, but if not, I guess it would be worth to check its implementation.

dkramer95:
@xeffyr I have found a more reliable method that should eliminate the need for the other custom workarounds.. I found that I needed to use nav bar height and that was what was different between the different devices I had.. Tested new fix on each device and seems to work for all of them!

xeffyr (approved):
Confirmed that it works on Samsung Galaxy S7 with workaround enabled. But there are issues:
- Blank status bar is present until touch keyboard was hidden.
- Property changes do not take effect on `termux-reload-settings`.

On Pixel 5 things are even worse. Status bar is just made black without disappearing. See screenshot and notice big black bar on the top. *Other apps using fullscreen are working properly.*
Screenshot_20210128-182859

***

In summary:
- This can be merged, but I guess "fullscreen workaround" should be always enabled.
- Further work on this topic is needed.

dkramer95:
@xeffyr I tested this on a Pixel 4 emulator and was able to replicate.. I tried experimenting w/ various flag combinations but didn't have much success...
____
Screen Shot 2021-01-28 at 5 59 43 PM
____
Screen Shot 2021-01-28 at 6 00 18 PM
____
Closest I was able to get was this, but the content is still obscured by the status bar.. I could never get it to display in this position and have the status bar text go away.. All other attempts, status bar text would be hidden, but it would have that blank margin at the top.
Screen Shot 2021-01-28 at 6 02 00 PM
____
I'm sure some combo of flags exists to satisfy everything, but haven't found it yet :D.

Issue #1878, opened 2021-01-10
[Feature Request] Ability to use Images as Terminal Background


WMCB-Tech:
**Feature description**
it would be cool if Termux can use pictures as background and the terminal will look aesthetic like anime background in your Terminal or matrix image background

**Reference implementation**

Does another app/terminal emulator have this feature?

Afaik, XFCE Terminal can set pictures as Terminal Background, so yeah it would be cool if termux does that the same

GxmerSam:
I agree too, maybe can be added with termux styling

xeffyr:
Duplicate of https://github.com/termux/termux-app/issues/365.

PR #1918, opened 2021-02-02
[Feature Request] Ability to use Images as Terminal Background #1878

srbbans:
I have implemented the terminal background image select feature with overlay color. :)


Cremalsa (changes requested):
Open

AbdullahBinJahed:
any progress on it ?
can I now set an image as the terminal background ?

PR #1923, opened 2021-02-08
Fix/Allow fullscreen mode to consume status bar area on some devices

dkramer95:
Screen Shot 2021-02-07 at 6 05 39 PM

This PR fixes the blank black bar that would display in fullscreen mode on some devices such as Pixel 4. 
@xeffyr Do you think we should make this another config option for users? On one hand, you get to utilize the extra vertical screen real estate, BUT 1-2 characters on left/right side can get cut off at the very top. Some might prefer one over the other?

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -222,11 +222,17 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
         View content = findViewById(android.R.id.content);
         content.setOnApplyWindowInsetsListener((v, insets) -> {
             mNavBarHeight = insets.getSystemWindowInsetBottom();
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                return insets.replaceSystemWindowInsets(0, 0, 0, mNavBarHeight);
+            }
             return insets;
         });
 
         if (mSettings.isUsingFullScreen()) {
             getWindow().addFlags(WindowManager.LayoutParams.FLAG_FULLSCREEN);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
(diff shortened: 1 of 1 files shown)


AyeSpacey:
This would be perfect !
Just to sacrifice some characters but i will be good 😍.

RalfWerner:
Your idea is good! In the [Termux-X11](https://github.com/termux/termux-x11/issues/29) plugin, this was optionally used and is selected via **Preferences**. The lower Nav-Bar was also used as [X11 screen](). On AVD sometimes a flicker bug occurs and together with the Termux [Float](https://github.com/termux/termux-float/issues/37#issue-986858850) Plugin is also often switched. Four pictures with a pixel (AVD):
grafik
- shot1 shows the preferences where both choices: extra-keys and fullscreen choices are **on**. In the middle, Termux is in the Float-Picture without extra-keys and below the _non-floating_ On-Screen KB
- shot2 is the X11 screen with some XClients and the KB (here with Extra Keys). The Float Picture is then moved upwards. And the transparent Nav-Bars above and below appear only with sweep from top or bottom.
- shot3 is like shot2 but without KB (switch off by Back-Key). The Float-Picture is then moved down (last position).
- shot4 shows the screen, when _tap_ on the float picture, which is then the KB and both Nav-Bars are displayed and reducing the X11 screen. After another _tap_ on X11, shot2 or shot3 is displayed again.

If you want to repeat this, you can use _long tap_ to move and scalate the Float picture or iconize (left in border). The rotation (Landscape) shows the _transparent_ two Nav Bars at the top and right (usable areas in X11):
grafik
In the current [Termux-X11](https://github.com/termux/termux-x11/pull/19#issuecomment-913093383) repo, a few bug fixes and features have gone lost. The current version has more disadvantages as advantages to the 3.5MB [initial version](https://www.dropbox.com/s/n85m6zkxkcvrk9p/dev-x11.apk?dl=0)

PR #1964, opened 2021-03-14
Allow installing Termux on the SD card

brunoais:


brunoais:
@kcubeterm Why closing? Is this not feasible for this app?

agnostic-apollo:
This will not allow storage of termux app data like `$HOME` and `$PREFIX` on the external removable sd card, it will just allow installation of APK on external storage.

>The .apk file is saved on the external storage, but all private user data, databases, optimized .dex files, and extracted native code are saved on the internal device memory.

https://developer.android.com/guide/topics/data/install-location

Moreover, check [Applications that should NOT install on external storage](https://developer.android.com/guide/topics/data/install-location#ShouldNot) section. This will cause problems with termux service, boot receiver, widgets, etc. Saving ~80MB (which may reduce to ~20MB in future) is not worth all the mess this will create.

brunoais:
I see... Thank you for explaining.

agnostic-apollo:
welcome.

PR #1969, opened 2021-03-16
Make Termux debuggable to allow ADB 'run-as'

### LINKED ISSUE #924: Feature Request: Add the ability to run Termux commands via. ADB without root.
Currently, in order to run a command, you have to do something like this:
```cmd
adb shell am start -n com.termux/.app.TermuxActivity
adb shell input text "[command here]"
adb shell input keyevent 113 && adb shell input keyevent 66
```
which, in my opinion, is a little hackey... There are a lot of things that could stop this script, like the screen being locked or a Termux session already being open with text typed in.

I propose that a new Activity/Service is added (or an extension of `com.termux/.app.TermuxService` to allow use by ADB). You could possibly even allow another app (with user granted permission) to run a command!

This would be BEYOND useful (and a little cool, too), like being able to start up `sshd` and setup port forwarding completely from the computer! You wouldn't even have to unlock the screen, assuming that you have already accepted the computer's RSA Fingerprint on your phone.

<hr>

If something to allow this already exists, just let me know and I can change the title or delete the "*Issue*".

intgr:
This resolves #924 and makes it possible to access Termux directrly from `adb shell` via `run-as`.

Not totally elegant, but it works:
```
% adb shell
dreamlte:/ $ run-as com.termux
dreamlte:/data/data/com.termux $ PATH=/data/data/com.termux/files/usr/bin LD_PRELOAD=/data/data/com.termux/files/usr/lib/libtermux-exec.so /data/data/com.termux/files/usr/bin/bash -l
~ $ python
Python 3.9.2 (default, Feb 22 2021, 12:26:04)
[Clang 9.0.8 (https://android.googlesource.com/toolchain/llvm-project 98c855489 on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
```

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -33,6 +33,7 @@
         android:banner="@drawable/banner"
         android:label="@string/application_name"
         android:theme="@style/Theme.Termux"
+        android:debuggable="true"
         android:supportsRtl="false" >
 
         <!-- This (or rather, value 2.1 or higher) is needed to make the Samsung Galaxy S8


intgr:
It would be useful to include a helper command that sets `PATH` and `LD_PRELOAD`, but it's useul without that too.

intgr:
Now, I'm extremely new to Android development. Should research whether the `android:debuggable` annotation has any downsides.

agnostic-apollo:
Dude, you don't release app APKs in `debuggable` mode. That's only to be used while developing an app.

intgr:
@agnostic-apollo What's the downside of enabling debuggable?

Seems like a Digital Restrictions Management feature to "protect" the user from themselves. It's not like Termux has some trade secrets locked up in the app. The website https://termux.com/ lists **Tinkerable** in bold letters. Why disallow debugging by the device's owner? It enables a convenient feature described in #924.

xeffyr:
@intgr Get a development build for yourself from Github Actions, what's the problem?

https://github.com/termux/termux-app/suites/2272001163/artifacts/47404244

agnostic-apollo:
The downside will be that we will likely never be able to upload APKs to google play store again (even after solving targetsdk 29 issues).

Check https://developer.android.com/studio/publish/preparing#turn-off-logging-and-debugging

Uploading debuggable APKs will throw errors like **You uploaded a debuggable APK or Android App Bundle. For security reasons you need to disable debugging before it can be published in Google Play.**

On top of that, debuggable APKs are not allowed by F-Droid builds either as mentioned [here](https://gitlab.com/fdroid/fdroidclient/-/issues/411#note_12157529).

Those are pretty small downsides, I think, well, not that small actually. But if this would allow you to **Tinker**, then, hell with those millions of users, you just say the word!

intgr:
Right, fair enough, then clearly this won't work. But there's no need to be such dicks about it.

> @intgr Get a development build for yourself from Github Actions, what's the problem?

You can see many people in #924 having this problem. This solution that you imply is so blindingly obvious was not mentioned, otherwise I wouldn't have went through this trouble.

agnostic-apollo:
Adding "/s" or smileys would have decreased the impact of the said sarcasm. Wasn't being a dick.

The solution would work but it isn't ideal for average users, since it would require manual update checking and apk signing keys issues would be there, i.e, all termux plugin apps will also need to be installed from github action builds. Installation would fail unless all APKs from other sources are uninstalled.

PR #1993, opened 2021-04-05
terminal emulator: disable fdsan

xeffyr:


xeffyr:
android_fdsan_set_error_level is not available for API < 29.

xeffyr:
Seems to work on per-process level and can't be done within the scope of full application.

PR #2009, opened 2021-04-19
Gradle update action

hannesa2:


hannesa2:
Gradle 7 comes with support for Apple Silicon. It's a benefit at least for me

agnostic-apollo:
Upgrading to `gradle` `7.0` is doable, will need to be first checked locally as well. However, it needs to be done with `-all` suffix instead of `-bin` as per 3af57303.

As far as gradle wrapper updater is concerned, instead of the auto-update, there are indeed real security concerns where a malicious binary may be committed to the repository and for that checksum verification should be done. However, merging of the 3rd party [gradle-update/update-gradle-wrapper](https://github.com/gradle-update/update-gradle-wrapper-action) probably shouldn't be done when there is an official one provided by `gradle` itself at [gradle/wrapper-validation-action](https://github.com/gradle/wrapper-validation-action) for checksum verification (not auto-update), security concerns are mentioned there as well.

Gradle auto-update can be looked into as well if a maintainer has the time.

PR #2073, opened 2021-05-19
Add contribute button

suhan-paradkar:


agnostic-apollo:
The [`v0.113`](https://github.com/termux/termux-app/releases/tag/v0.113) has already added an `About` page with links to termux repositories, not just `termux-packages`. Moreover, pull requests with hardcoded `com.termux` strings and paths are not accepted anymore. Check https://github.com/termux/termux-app#for-devs-and-contributors

For some reason [HelpActivity](https://github.com/termux/termux-app/blob/v0.113/app/src/main/java/com/termux/app/activities/HelpActivity.java) still has a hardcoded wiki link, will fix that.

Issue #1995, opened 2021-04-06
USB-C Yubikey device disables on-screen keyboard


cmosetick:
Hello and thank you for this excellent project.

I am a part-time security researcher and tester.
I have recently been doing some research on Android and found Termux to be helpful.
However, I have been road blocked by a USB issue with my testing device.
Please note that the Android device is NOT heavily customized.
So I am looking for other people that can reproduce and confirm the issue I describe below.
If several people are not able to reproduce this issue, then I am happy to consider it invalid.

**Problem description**
Inserting a Yubikey USB device into USB-C port on Android handset disables on-screen keyboard within Termux Application.
Device Tested: [Yubikey 5Ci](https://www.yubico.com/ch/product/yubikey-5ci)

NOTE - I do not see this issue with other applications while a Yubikey USB device is plugged into the same handset.
(i.e. the on-screen keyboard works normal and as expected in all other applications I have tested.)

NOTE: Inserting a regular USB-C mass storage device does not make the keyboard disappear!
Device Tested: A regular SanDisk USB-C flash disk formatted with a couple partitions. 

**Steps to reproduce**
- Plug in a Yubikey device into USB-C port of Android Handset.
- You will immediately see the regular on-screen keyboard within Termux App disappear.
- Tapping on the screen does not make on-screen keyboard re-appear.
- I have not found a way to make on-screen keyboard re-appear.

**Expected behavior**
- Android on-screen keyboard does not disappear.
OR
- There is a way to make on-screen keyboard come back and be usable even with Yubikey plugged into device.

**Additional information**

* Termux application version: `0.108` installed via F-Droid. (same bug experienced with previous Play Store version)
* Android OS version: `10`
* Linux Kernel version: `4.14.141-20190226`
* Device model: Samsung Galaxy

Please note that the Yubikey USB device works perfectly fine with @Yubico `yubioath-android` App on the same device.
(installed via [Play Store](https://play.google.com/store/apps/details?id=com.yubico.yubioath))

CC:
- https://github.com/Yubico/yubioath-android
- https://github.com/Yubico/yubico-java-client

Grimler91:
Thanks for the detailed report. 

The yubikey [identifies as a keyboard](https://support.yubico.com/hc/en-us/articles/360013790279-The-YubiKey-as-a-Keyboard) so that is why termux thinks you are using an external keyboard (and do not need the soft-keyboard). 

[...]

PR #2114, opened 2021-06-06
Fix issue where soft keyboard would not show in some cases

### LINKED ISSUE #2111: If keyboard toggle is enable/disable, you have to press it twice after keyboard is auto closed
**Problem description**
With the option introduced in e4e638bd318baf6ad344c898c3baa9093975e53f for #1098, I'm having some issues with the keyboard auto closing. When this happens, and I press the keyboard toggle, nothing happens until I press it a second time (because the keyboard is disabled on the first press even though it's closed). Personally, I don't think the keyboard should auto close when having this option set, it should always be shown when the keyboard enabled.

**Steps to reproduce**
1. Set `soft-keyboard-toggle-behaviour = enable/disable`.
2. Open Termux and have the keyboard open.
3. Lock the screen.
4. Unlock the screen.

Alternatively:
1. Set `soft-keyboard-toggle-behaviour = enable/disable`.
2. Open Termux and have the keyboard open.
3. Open a URL.
4. Press back in the browser to return to Termux.

**Expected behavior**
The keyboard should still be open in Termux after these steps. Currently, it's closed which causes the issue mentioned initially about having to press the keyboard toggle twice.

**Additional information**
When the `back-key` option is default/back, the keyboard is closed on back. This causes the same issue that you have to press the keyboard toggle twice. I think when the keyboard is set to enable/disable, the back key should either take you straight out of the app without closing the keyboard (like when you press back after the keyboard is closed), or alternatively disable the keyboard.

* Termux application version: 0.113
* Android OS version: 11
* Device model: OnePlus 9 Pro

### LINKED ISSUE #2112: If keyboard toggle is enable/disable, the toggle doesn't work properly after leaving and opening the app
**Problem description**
If you use the option to enable/disable the keyboard and back out of the app with the keyboard disabled, it can't be enabled with the keyboard toggle alone when entering the app again.

**Steps to reproduce**
1. Set `soft-keyboard-toggle-behaviour = enable/disable`.
2. Press the toggle to disable the keyboard.
3. Press back to leave (but not exit) the app.
4. Return to the app by pressing it in recent apps or the launcher.
5. Press the keyboard toggle to open the keyboard again.

**Expected behavior**
The keyboard should open when I press the toggle. Currently nothing happens. To open the keyboard I have to press the terminal view after pressing the toggle.

* Termux application version: 0.113
* Android OS version: 11
* Device model: OnePlus 9 Pro

agnostic-apollo:
1. If `soft-keyboard-toggle-behaviour=enable/disable` was set, then pressing keyboard toggle wouldn't show the keyboard after switching back from another app if keyboard was previously disabled by user.
2. If switching back from another app, like when opening url with context menu "Select URL" long press and returning to termux with back button, then soft keyboard wouldn't automatically open like it does on app startup.

Fixes #2111, Fixes #2112

@trygveaa Here it is, kindly test. Thanks.

### DIFF
--- a/app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java
+++ b/app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java
@@ -57,6 +57,9 @@ public class TermuxTerminalViewClient extends TermuxTerminalViewClientBase {
 
     private Runnable mShowSoftKeyboardRunnable;
 
+    private boolean mShowSoftKeyboardIgnoreOnce;
+    private boolean mShowSoftKeyboardWithDelayOnce;
+
     private static final String LOG_TAG = "TermuxTerminalViewClient";
 
     public TermuxTerminalViewClient(TermuxActivity activity, TermuxTerminalSessionClient termuxTerminalSessionClient) {
@@ -410,10 +413,19 @@ public class TermuxTerminalViewClient extends TermuxTerminalViewClientBase {
                 mActivity.getPreferences().setSoftKeyboardEnabled(false);
                 KeyboardUtils.disableSoftKeyboard(mActivity, mActivity.getTerminalView());
             } else {
+                // Show with a delay, otherwise pressing keyboard toggle won't show the keyboard after
+                // switching back from another app if keyboard was previously disabled by user.
+                // Also request focus, since it wouldn't have been requested at startup by
+                // setSoftKeyboardState if keyboard was disabled. #2112
                 Logger.logVerbose(LOG_TAG, "Enabling soft keyboard on toggle");
                 mActivity.getPreferences().setSoftKeyboardEnabled(true);
                 KeyboardUtils.clearDisableSoftKeyboardFlags(mActivity);
-                KeyboardUtils.showSoftKeyboard(mActivity, mActivity.getTerminalView());
+                if(mShowSoftKeyboardWithDelayOnce) {
+                    mShowSoftKeyboardWithDelayOnce = false;
+                    mActivity.getTerminalView().postDelayed(getShowSoftKeyboardRunnable(), 500);
+                    mActivity.getTerminalView().requestFocus();
+                } else
+                    KeyboardUtils.showSoftKeyboard(mActivity, mActivity.getTerminalView());
             }
         }
         // If soft keyboard toggle behaviour is show/hide
@@ -431,12 +443,20 @@ public class TermuxTerminalViewClient extends TermuxTerminalViewClientBase {
     }
 
     public void setSoftKeyboardState(boolean isStartup, boolean isReloadTermuxProperties) {
+        boolean noRequestFocus = false;
+
         // If soft keyboard is disabled by user for Termux (check function docs for Termux behaviour info)
         if (KeyboardUtils.shouldSoftKeyboardBeDisabled(mActivity,
-                mActivity.getPreferences().isSoftKeyboardEnabled(),
-                    mActivity.getPreferences().isSoftKeyboardEnabledOnlyIfNoHardware())) {
+            mActivity.getPreferences().isSoftKeyboardEnabled(),
+            mActivity.getPreferences().isSoftKeyboardEnabledOnlyIfNoHardware())) {
             Logger.logVerbose(LOG_TAG, "Maintaining disabled soft keyboard");
             KeyboardUtils.disableSoftKeyboard(mActivity, mActivity.getTerminalView());
+            noRequestFocus = true;
+            // Delay is only required if onCreate() is called like when Termux app is exited with
+            // double back press, not when Termux app is switched back from another app and keyboard
+            // toggle is pressed to enable keyboard
+            if (isStartup && mActivity.isOnResumeAfterOnCreate())
+                mShowSoftKeyboardWithDelayOnce = true;
         } else {
             // Set flag to automatically push up TerminalView when keyboard is opened instead of showing over it
             KeyboardUtils.setResizeTerminalViewForSoftKeyboardFlags(mActivity);
@@ -450,34 +470,55 @@ public class TermuxTerminalViewClient extends TermuxTerminalViewClientBase {
                 KeyboardUtils.hideSoftKeyboard(mActivity, mActivity.getTerminalView());
                 // Required to keep keyboard hidden when Termux app is switched back from another app
                 KeyboardUtils.setSoftKeyboardAlwaysHiddenFlags(mActivity);
-            } else {
-                // Do not force show soft keyboard if termux-reload-settings command was run with hardware keyboard
-                if (isReloadTermuxProperties)
-                    return;
-
-                if (mShowSoftKeyboardRunnable == null) {
-                    mShowSoftKeyboardRunnable = () -> {
-                        Logger.logVerbose(LOG_TAG, "Showing soft keyboard on focus change");
-                        KeyboardUtils.showSoftKeyboard(mActivity, mActivity.getTerminalView());
-                    };
-                }
-
-                mActivity.getTerminalView().setOnFocusChangeListener(new View.OnFocusChangeListener() {
-                    @Override
-                    public void onFocusChange(View view, boolean hasFocus) {
-                        // Force show soft keyboard if TerminalView or toolbar text input view has
-                        // focus and close it if they don't
-                        boolean textInputViewHasFocus = false;
-                        final EditText textInputView =  mActivity.findViewById(R.id.terminal_toolbar_text_input);
-                        if (textInputView != null) textInputViewHasFocus = textInputView.hasFocus();
-                        KeyboardUtils.setSoftKeyboardVisibility(mShowSoftKeyboardRunnable, mActivity, mActivity.getTerminalView(), hasFocus || textInputViewHasFocus);
-                    }
-                });
-
-                // Request focus for TerminalView
-                mActivity.getTerminalView().requestFocus();
+                noRequestFocus = true;
+                // Required to keep keyboard hidden on app startup
+                mShowSoftKeyboardIgnoreOnce = true;
             }
         }
+
+        mActivity.getTerminalView().setOnFocusChangeListener(new View.OnFocusChangeListener() {
+            @Override
+            public void onFocusChange(View view, boolean hasFocus) {
+                // Force show soft keyboard if TerminalView or toolbar text input view has
+                // focus and close it if they don't
+                boolean textInputViewHasFocus = false;
+                final EditText textInputView =  mActivity.findViewById(R.id.terminal_toolbar_text_input);
+                if (textInputView != null) textInputViewHasFocus = textInputView.hasFocus();
+
+                if (hasFocus || textInputViewHasFocus) {
+                    if (mShowSoftKeyboardIgnoreOnce) {
+                        mShowSoftKeyboardIgnoreOnce = false; return;
+                    }
+                    Logger.logVerbose(LOG_TAG, "Showing soft keyboard on focus change");
+                } else {
+                    Logger.logVerbose(LOG_TAG, "Hiding soft keyboard on focus change");
+                }
+
+                KeyboardUtils.setSoftKeyboardVisibility(getShowSoftKeyboardRunnable(), mActivity, mActivity.getTerminalView(), hasFocus || textInputViewHasFocus);
+            }
+        });
+
+        // Do not force show soft keyboard if termux-reload-settings command was run with hardware keyboard
+        // or soft keyboard is to be hidden or is disabled
+        if (!isReloadTermuxProperties && !noRequestFocus) {
+            // Request focus for TerminalView
+            // Also show the keyboard, since onFocusChange will not be called if TerminalView already
+            // had focus on startup to show the keyboard, like when opening url with context menu
+            // "Select URL" long press and returning to Termux app with back button. This
+            // will also show keyboard even if it was closed before opening url. #2111
+            Logger.logVerbose(LOG_TAG, "Requesting TerminalView focus and showing soft keyboard");
+            mActivity.getTerminalView().requestFocus();
+            mActivity.getTerminalView().postDelayed(getShowSoftKeyboardRunnable(), 300);
+        }
+    }
+
+    private Runnable getShowSoftKeyboardRunnable() {
+        if (mShowSoftKeyboardRunnable == null) {
+            mShowSoftKeyboardRunnable = () -> {
+                KeyboardUtils.showSoftKeyboard(mActivity, mActivity.getTerminalView());
+            };
+        }
+        return mShowSoftKeyboardRunnable;
     }
 
 
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -130,6 +130,11 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
     private boolean mIsVisible;
 
     /**
+     * If onResume() was called after onCreate().
+     */
+    private boolean isOnResumeAfterOnCreate = false;
+
+    /**
      * The {@link TermuxActivity} is in an invalid state and must not be run.
      */
     private boolean mIsInvalidState;
@@ -160,6 +165,7 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
     public void onCreate(Bundle savedInstanceState) {
 
         Logger.logDebug(LOG_TAG, "onCreate");
+        isOnResumeAfterOnCreate = true;
 
         // Check if a crash happened on last run of the app and show a
         // notification with the crash details if it did
@@ -251,6 +257,8 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
 
         if (mTermuxTerminalViewClient != null)
             mTermuxTerminalViewClient.onResume();
+
+        isOnResumeAfterOnCreate = false;
     }
 
     @Override
@@ -689,6 +697,10 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
 
     public boolean isVisible() {
         return mIsVisible;
+    }
+
+    public boolean isOnResumeAfterOnCreate() {
+        return isOnResumeAfterOnCreate;
     }
 
 


app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java
```diff
@@ -410,10 +412,15 @@ public void onToggleSoftKeyboardRequest() {
...
             } else {
+                // Show with a delay, otherwise pressing keyboard toggle won't show the keyboard after
+                // switching back from another app if keyboard was previously disabled by user.
+                // Also request focus, since it wouldn't have been requested at startup by
+                // setSoftKeyboardState if keyboard was disabled. #2112
                 Logger.logVerbose(LOG_TAG, "Enabling soft keyboard on toggle");
                 mActivity.getPreferences().setSoftKeyboardEnabled(true);
                 KeyboardUtils.clearDisableSoftKeyboardFlags(mActivity);
-                KeyboardUtils.showSoftKeyboard(mActivity, mActivity.getTerminalView());
+                mActivity.getTerminalView().postDelayed(getShowSoftKeyboardRunnable(), 300);
```
trygveaa, line 422: Is this delay necessary? Just requesting focus makes it works for me. I see no difference in behavior with or without the delay, except that toggling the keyboard now has a delay.

trygveaa (commented):
Great, now the keyboard opens after returning from a URL or locked screen. However, is there a reason that the keyboard should open even if it was hidden before? I think it would be better if the keyboard remained in the same state as it was previously when you return, so that if you hide the keyboard and change to another app, it's still hidden when you switch back to Termux.

Of course, if you use the disable toggle, it will be closed when you come back to Termux, but if it behaved the way described above, I would prefer using the toggle as show/hide instead of enable/disable.

There is also still one issue remaining with the enable/disable toggle. If you press back to hide the keyboard, you still have to press the toggle twice to open the keyboard.

agnostic-apollo:
> However, is there a reason that the keyboard should open even if it was hidden before? I think it would be better if the keyboard remained in the same state as it was previously when you return, so that if you hide the keyboard and change to another app, it's still hidden when you switch back to Termux.

There is [no reliable way](https://github.com/termux/termux-app/issues/1995#issuecomment-840992902) for termux to know whether the soft keyboard is open or closed, so I can't save the state before leaving termux. It only works with `enable/disable` because I'm using the disable  flags as a hack.

> There is also still one issue remaining with the enable/disable toggle. If you press back to hide the keyboard, you still have to press the toggle twice to open the keyboard.

That's expected behaviour and won't change. There is a difference between hide and disable. When you hide, you should tap the terminal view for it to show. When you disable, you should press keyboard toggle to show it again. When you press the keyboard toggle in `enable/disable` mode, as mentioned above, termux doesn't know if keyboard is hidden or shown, it only sets/clears disable flags. It works on twice press, because on first press, it sets the disable flags, second time clears the flags and shows the keyboard. Check `logcat | grep -i keyboard` for transitions with verbose log level. Use `show/hide` instead of your behaviour.

app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java
agnostic-apollo: Yes, as mentioned in the comment above it. It requires a delay after steps from #2112 on my device. It works with 100ms too, but may not work for [other slower devices](https://stackoverflow.com/a/29230447/14686958). I have no way to know the correct one unless multiple testers come and say it works for them, so better to set a looser delay, since releases are slow. And at least `300ms` may be required anyways. The `setSoftKeyboardVisibility()` called by `onFocusChange()` has it at `500ms`.

trygveaa:
> There is [no reliable way](https://github.com/termux/termux-app/issues/1995#issuecomment-840992902) for termux to know whether the soft keyboard is open or closed, so I can't save the state before leaving termux. It only works with `enable/disable` because I'm using the disable flags as a hack.

Oh, okay, too bad.

> That's expected behaviour and won't change. There is a difference between hide and disable. When you hide, you should tap the terminal view for it to show. When you disable, you should press keyboard toggle to show it again. When you press the keyboard toggle in `enable/disable` mode, as mentioned above, termux doesn't know if keyboard is hidden or shown, it only sets/clears disable flags. It works on twice press, because on first press, it sets the disable flags, second time clears the flags and shows the keyboard. Check `logcat | grep -i keyboard` for transitions with verbose log level. Use `show/hide` instead of your behaviour.

I think it would be more useful/practical if the toggle button always shows the keyboard when it's closed, so you don't have to know if it's disabled or just hidden. But the main reason I want this behavior is that I'm working on being able to open URLs by pressing on them directly, and I was thinking of making the tap action configurable so I don't accidentally open the keyboard if I miss a URL. But okay, if Termux can't know if the keyboard is open or not, I see that it wouldn't be possible.

Did you see my comment about the delay that doesn't seem to be necessary?

agnostic-apollo:
> But the main reason I want this behavior is that I'm working on being able to open URLs by pressing on them directly, and I was thinking of making the tap action configurable so I don't accidentally open the keyboard if I miss a URL.

I guess for that case, as you suggested, the back key can disable the keyboard as well if a `termux.properties` is `true`. Maybe that works.

>But okay, if Termux can't know if the keyboard is open or not, I see that it wouldn't be possible.

Yeah, but if someone can come up with a hack that works for everybody, then that can be included.

>Did you see my comment about the delay that doesn't seem to be necessary?

Already replied to it.

app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java
trygveaa: Okay, that's a bit surprising. I see that the delay is required when auto opening the keyboard after returning to Termux (the delay later down in this file), but here there would be some time between returning to Termux and pressing the button anyways (I don't think one would be able to press the button within a couple of hundred ms after returning to the app). I'm not doubting what you're saying, just wanted to make sure it's necessary in both cases.

The problem is not so much that there is a delay after returning to the app though, but that this also causes a delay when you haven't left the app. E.g. just press the toggle to disable and then press again to enable. This delay is not present in master, so I would assume it's not necessary in this case? Would it be possible to differentiate between these cases, so the delay is only used when you return to Termux, and not when you toggle the keyboard when Termux is active all the time?

trygveaa:
> I guess for that case, as you suggested, the back key can disable the keyboard as well if a `termux.properties` is `true`. Maybe that works.

Alright, maybe I'll add that when I make a PR for the click on URL functionality then.

> Yeah, but if someone can come up with a hack that works for everybody, then that can be included.

Fingers crossed, though from what you and the others have written in the issue you linked, it doesn't seem easy.

> Already replied to it.

Sorry, I didn't see it. I think GitHub showed the one I replied to automatically, but not the other until I refreshed the page. Or I was just blind.

app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java
agnostic-apollo: >, but here there would be some time between returning to Termux and pressing the button anyways

I was wondering about that too, I tried multiple configurations, but delays seems necessary.

>Would it be possible to differentiate between these cases

Yeah, a boolean can be added, just like `mShowSoftKeyboardIgnoreOnce`, delays seems to be required only after app changes.

trygveaa: > Yeah, a boolean can be added, just like mShowSoftKeyboardIgnoreOnce, delays seems to be required only after app changes.

Great, or alternatively we could call it both immediately and with a delay? That would fix the issue when toggling when Termux is active, and it wouldn't introduce a delay after app switch for devices that don't need it. Or would that be too hacky?

agnostic-apollo: In that case, if user hid the keyboard with back button before timer finishes, it would pop up again. I have added the boolean. Kindly test, thanks.

agnostic-apollo:
> Alright, maybe I'll add that when I make a PR for the click on URL functionality then.

Cool

> Fingers crossed, though from what you and the others have written in the issue you linked, it doesn't seem easy.

Probably not or at least not on all devices.

>Sorry, I didn't see it. I think GitHub showed the one I replied to automatically, but not the other until I refreshed the page. Or I was just blind.

Yeah, comments are not always refreshed. Sub comments probably don't either.

But if you do find out that you are actually blind, please let us know. We require our testers to not be blind to test UI changes. Thanks ;)

app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java
trygveaa: True, but since it's only 300 ms I wouldn't think it's an issue in practice.

I tested a bit more and I see that if I press back out of Termux and open it again (not exiting Termux, but so that onDestroy and onCreate is called), I have some delay even if I remove the delay from the code. However if I just switch to another app or home and back (so that only onStop, onStart and onResume is called) there is only the delay added in the code.

This makes me suspect the delay is only necessary if onDestroy/onCreate is called. Could you test if the delay is necessary on your device if you switch to another app and back to Termux (so that you trigger just onStop/onStart/onResume, not onDestroy/onCreate)?

RalfWerner:
The two KB [choices](https://github.com/termux/termux-app/issues/1995#issuecomment-841668368) in Termux Settings are set to other values ​​regardless of the selected one with each `KB` action. So they are _**not** needed_ because they do not do anything or switch permanently to other values.

The _double_ KB action is always repeatable with properties: `*on-startup=true`, `*behaiour=enable/disable`, switch KB on,  press Start-Menu Key (or sleep mode) and open termux-app again.
With the explanations above, I understood that this _can **not** be prevented._

app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java
agnostic-apollo: > I have some delay

Delay in what? Enabling keyboard with toggle if it was previously disabled?. `onCreate` will always have its own delay to set up stuff, so showing keyboard automatically or with toggle will have "some" delay anyways.

trygveaa: Delay between pressing the keyboard toggle and the keyboard appearing. That is after returning to the app when the keyboard is disabled, so not when it appears automatically.

These are the two cases:

With destroy:
1. Press the toggle to disable the keyboard.
2. Press back to leave (but not exit) the app.
3. Return to the app by pressing it in recent apps or the launcher.
4. Press the keyboard toggle to open the keyboard again.
5. The keyboard will appear.

Here there is some delay between 4. and 5. even if I remove the delay from the code.

Without destroy:
1. Press the toggle to disable the keyboard.
2. Press home to go to the home screen.
3. Return to the app by pressing it in recent apps or the launcher.
4. Press the keyboard toggle to open the keyboard again.
5. The keyboard will appear.

Here the only delay between 4. and 5. is the delay specified in the code. If I remove that delay the keyboard opens immediately.

I'm asking if the delay specified in the code is necessary in both cases for the keyboard to appear on your device, or if it is sufficient to have it in the first case.

agnostic-apollo: It seems its only required if `onCreate` is called. I have made the changes, but if its also required even when `onResume` is called without `onCreate` on startup on some device, then we all know who will be coming after.

agnostic-apollo:
>So they are not needed because they do not do anything or switch permanently to other values.

Only **1** `Soft Keyboard Enabled` toggle value changes outside of `Termux Settings` and **only** when `soft-keyboard-toggle-behaviour=enable/disable` is set, which its supposed to do. It does not change with `show/hide`. I don't know why you still haven't figured this out yet after all this time.

> With the explanations above, I understood that this can not be prevented.

With `hide-soft-keyboard-on-startu=true`, either double press toggle or touch on terminal view (if keyboard not disabled).

app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java
trygveaa: Perfect, thanks! Working great for me now.

trygveaa (approved):

app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java
agnostic-apollo: Great, welcome. Merging.

RalfWerner:
> I don't know why you still haven't figured this out yet after all this time.


You're right the `*=show/hide` Property I had only checked once. It led to _**"no chance to use OS-KB allways"**_. This is also the case in the currently merded version. I tried to solve the problem with `rm .termux/te*`, close all apps, reboot of the device and also reinstall the APK. Everything without success. Since I also start the `sshd` service at the Termux start, the last chance is using the PC and _**its KB**_ for editing the properties in a `ssh` Shell.

So far I have tried to understand the OS-KB behavior _without_ source analysis. I can only explain the condition described above by an init problem. Also by commenting:
```
#hide-soft-keyboard-on-startup=true
hide-soft-keyboard-on-startup=false # doesn't work when Termux Setting is "off"
#soft-keyboard-toggle-behaviour=show/hide
#soft-keyboard-toggle-behaviour=enable/disable
```
I can _not get back the OS-KB_. The only chance is to edit with the PC in the Properties `*=enable/disable` (line 3) But behind `*=show/hide` or set the first choice in the [7-step setting](https://github.com/termux/termux-app/issues/1995#issuecomment-843276188). Who does not know that, can easily despair or opens Issues.
The second choice I will check on another device (with BT-KB) and inform you again.

agnostic-apollo:
@RalfWerner Sorry, I don't understand what you are saying, so can't help. May get someone who can understand your local language and can write readable english to write the comment for you.

If someone else can understand the above comment and can explain, please do.

RalfWerner:
Thanks for your response @agnostic-apollo. I assume it's also to my question [here](https://gitter.im/termux/dev?at=60c39dbda5d5740d611cf628). I can understand the two (A: my and B: yours) commits above well when I translate in my local language. So I've prepared a Test with two sentences from above and translate them with google into German -> Spain -> French -> English. before:

A: I find the behavior extremely confusing. OS-KB is never displayed or always or after various KEYBOARD actions
B: I don't understand what you are saying, so can't help. If someone else can understand the above comment and can explain, please do.

after:
A: I find the extremely confusing behavior. OS-KB is never displayed or always or after different keyboard actions
B: I do not understand what you say, so I can not help. If someone else can understand and explain the previous comment, please do it.

The result is still understandable _for me_, even with the small differences.
If you did not understand the semantics, this can be assigned mistakenly to the language. A picture or movie sometimes says more than 1000 words even if it can be misinterpreted.

https://user-images.githubusercontent.com/45426380/121796568-4728bc00-cc1a-11eb-9c4c-0da1f3d3f50c.mp4

Here you see the behavior of Termux + X11 in Split mode on a real device. The effect is also repeatable to AVD and with other apps as X11. Since I know the effect, can repeat it and avoid, I will not suffice so often. Since you currently have no interest in troubleshooting is Ok and I'm waiting for others to annoy them and open Issues that you understand better.

agnostic-apollo:
Understanding one or two sentences here and there does not mean I understand what you are trying to say overall. And I don't have time to decipher it all.

> I find the extremely confusing behavior. OS-KB is never displayed or always or after different keyboard actions

Fill following template (No stories!)

- Device model: 
- Android Version: 
- Soft Keyboard App: 
- Hardware Keyboard State: connected/not connected
- Hardware Keyboard Type: usb/bluetooth
- Android Settings Language And Input `Show Soft Keyboard When Hardware Attached`: true/false
- Termux Settings `Soft Keyboard Enabled`: true/false
- `hide-soft-keyboard-on-startup`: 
- `soft-keyboard-toggle-behaviour`: 

If for above `Hardware Keyboard Type=bluetooth`, does it only happen with bluetooth keyboard or for usb keyboard as well.

>Here you see the behavior of Termux + X11 in Split mode on a real device. 

In this video the keyboard is showing so different from "OS-KB is never displayed". 

Anyways, did this start on termux `v0.114` or was it happening on previous versions too. If yes, then this is likely due to margin adjustment code e7dd0eee. It was expected that it would break on some edge case.

I need the logs to figure out what is happening. Set log level to `Verbose` and enable `Termux View Key Logging` toggle in termux settings, then run `logcat` command in termux. Trigger the bug, then post the log here with `Share transcript` in a markdown code block. No screenshots or videos! Thanks

RalfWerner:
> ... there does not mean I understand what you are trying to say overall. And I don't have time to decipher it all.

I also accepted that and that we cancel the common search for solutions now.
Checks I performed on various real devices and AVD with Android versions of 8-11. In the discovered errors, bugs or my lack of knowledge (I say concerns) I _have_ given what I used. Since you do not understand my descriptions, the video should be a reference to an open concerns. The 17 seconds video you should be able to repeat on all your _split-enabled_ devices with v114.
> ... In this video the keyboard is showing so different from "OS-KB is never displayed".

If it interests you: the video above shows my Huawei-Y7/Android 8, mirrored with _scrcpy_ and filmed with _apowersoft_. A shot is _not_ possible because the _flickers_ then can _not_ be seen. The _Extra keys_ of both splits are sometimes there, sometimes away and sometimes twice on unexpected positions. The OS KB is once there and finally away. Due to my screen touches, it would have appeared and disappear should be about 5-10 times. The topic BT/USB and PiP is _not treated_ in the video. It can be repeated on all AVD. So also on your and yourself can produce a few thousand _logcat_ lines.
> If for above Hardware Keyboard Type=bluetooth, does it only happen with bluetooth keyboard or for usb keyboard as well.

All checks with BT=Blue Tooth I have done with identical effects with USB. So BT-KB=USB-KB. You've used another term: _Hardware KB,_ with which the PC keyboard is meant using Android device mirrored (scrcpy, TeamViewer) on a PC monitor. This _have_ a different behavior.

Most concerns I have with my PiP capable device. You have no and no emulator found so we continue talking _different languages._

It is as if I wanted to explain an Android device to my mother 40 years ago. She would not understand me, even though she taught me speak. Today and with some time and patience it is no problem anymore. That she do not want to understand and learn the `bash` command Line with over 90, I can understand and accept.
In this sense, I will not make you nerves or spam you with my _**concerns.**_

PR #2193, opened 2021-07-28
Update workflows

suhan-paradkar:
Well.. this will ease up some jobs...
Here.. I've set up dependabot and set up nother workflow to update the gradle wrapper weekly...

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/.github/workflows/update-wrapper.yml
+++ b/.github/workflows/update-wrapper.yml
@@ -1,0 +1,16 @@
+name: Update Gradle Wrapper
+
+on:
+  schedule:
+    - cron: "30 6 * * SAT"
+
+jobs:
+  update-gradle-wrapper:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Update Gradle Wrapper
+        uses: gradle-update/update-gradle-wrapper-action@v1
+


agnostic-apollo:
Hey,
Similar workflows were sent before too. I don't think there is a specific need for these, `gradle` dependencies are usually bumped before releases in a single commit and I usually locally test the build before pushing. Same for when updating `gradle` version. And gradle wrapper scripts don't really need to be updated, at the very least that often that we add a unofficial 3rd party workflow for it.

https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper

>Projects will typically want to keep up with the times and upgrade their Gradle version to benefit from new features and improvements. One way to upgrade the Gradle version is manually change the distributionUrl property in the Wrapper’s gradle-wrapper.properties file. The better and recommended option is to run the wrapper task...
>Note that running the wrapper task once will update gradle-wrapper.properties only, but leave the wrapper itself in gradle-wrapper.jar untouched. This is usually fine as new versions of Gradle can be run even with ancient wrapper files. If you nevertheless want all the wrapper files to be completely up-to-date, you’ll need to run the wrapper task a second time.

However, if `dependabot` were to be added, then one would at least have to disable updates for [`commons-io:commons-io:2.5`](https://github.com/termux/termux-app/blob/v0.117/termux-shared/build.gradle#L21) with [`ignored_updates`](https://dependabot.com/docs/config-file/#ignored_updates).

suhan-paradkar:
Okay.. then I shall remove gradle update workflow and add disable updates for commons-io when I will be free...

agnostic-apollo:
Grouped PR are [still not supported](https://github.com/dependabot/dependabot-core/issues/1190) and won't be released anytime soon. This would create a separate pull request for each dependency update, even for `patch` (semvar) version updates. A whole lot of PR noise.

suhan-paradkar:
Well.. then I shall close this then....

agnostic-apollo:
Yeah, sorry, I don't think this is very useful for `termux-app`. It will likely create more work managing multiple pull requests and syncing with local repo. But if other maintainers think otherwise and want this added, I have no issue with `dependabot` workflow being merged.

PR #2237, opened 2021-08-23
Extra keys conversion to agnosticism

agnostic-apollo:
- The termux `extra-keys` have been moved to `termux-shared` library so that they can be imported and used by other apps for their own needs as long as they comply with `GPLv3` license.
- You can now long hold on `CTRL`, `ALT`, `SHIFT` and `FN` to lock those control keys. They will not be released when you press another key and will only be released by pressing the respective control key again.
- Add shift key support in extra keys and terminal with `SHIFT` or `SHFT`.
- Ensured `FN` extra key is read by the terminal.

@suhan-paradkar let me know if some design change needs to be made for https://github.com/suhan-paradkar/termux-wayland. Check commits and javadocs for more details, there should be enough info to get it working.

Someone on debug builds should "ideally" also test this before I merge, I have tested and seems to be working fine on my Android `7` device.


suhan-paradkar:
Thanks for that @agnostic-apollo I will be working on it soon and will let you know about that👍👍

suhan-paradkar:
Well.. for your information.. I will be working on it [here](https://github.com/suhan-paradkar/termux-wayland/pull/42)

agnostic-apollo:
You are welcome.

Jitpack build [was failing](https://jitpack.io/com/github/agnostic-apollo/termux-app/extra-keys-conversion-to-agnosticism-d1478fb6c3-1/build.log) which failed yours. Is fixed now.

suhan-paradkar:
Thanks 👍 lemme rerun the CI

suhan-paradkar:
Well.. is there any way an app can read termux-app's Sharedpreferences set in `./termux/termux.properties` ?

RalfWerner:
I've extend my Extra-Keys (EK) with `ALT,SHIFT,FN,` and checked on different AVD and my Android 9 device. The long-tap/hold of the first four Keys has worked as described (helpfull in `mc`). Here is a shot of the complete `*properties`:
grafik
The 11th Key is `KEYBOARD` and still working sometimes and sometimes not! The 11th `popup` is `DRAWER` but termux _**chrashes here**_ allways on all checked devices (not draw from left). _**Could you repeat that?**_
With _draw_ **on EK** from right to left the EK will removed and replaced by OS-KB-Settings (gboard) and _word input mode_
That is nice (floating/non-floating toggle is possible) but no more EK, _**Should that be so?**_

xeffyr:
What program you are using to test or work with keys `FN` and `SHIFT`?

xeffyr:
Used `vtshowkeys` from `vtutils` package.

I can be wrong, but it appears that `FN` simply do nothing. Here is what output `vtshowkeys` procuces on PC for combination `FN` + left arrow:
```
0033  27 0x1B Escape '^[' or '\e'
0133  91 0x5B ASCII '['
0110  72 0x48 ASCII 'H'
```

In Termux it produces *nothing*.

`SHIFT` appears to be working correctly.

RalfWerner:
> What program do you use to test or work with keys `FN`

Whether Key `FN` is really needed I know either. But it is included on many real keyboards and also does something.
[Here](https://github.com/termux/termux-app/issues/1995#issuecomment-843276188) is a video in which my BT-KB is coupled with different devices (also with `FN`) and the effects in Termux are confusing. For example, `FN+Right/Left` at AVD corresponds to the _startscreen function_ (`Home`) but not with Hard-KB and `scrcpy` or direct coupling.
If it is possible to merge the two EK from _Termux_ and **T:X11**, new possibilities of the X protocol result. Currently with **T:Wayland** only small sentence (eight keys) without config possible. If I use BT-KB so the keys above are `HOME/END` but also other buttons with `FN` like Num_Log, Screenshot, Search are recognized correctly. However, these buttons are not offered in most OS-KB (Hackers KB only).
My goal is to minimize the finger work and if FN is useful for this I do not know yet. I wish @suhan-paradkar  a lot of skill and success in the merge try. Incidentally, the paste button is also implemented correctly in @agnostic-apollo  artifact.

xeffyr:
> Here is a video in which my BT-KB is coupled with different devices (also with FN) and the effects in Termux are confusing. For example, FN+Right/Left at AVD corresponds to the startscreen function (Home)

Video appears to be about older Termux version.

I'm asking specifically about this pull request and only about behavior of software keyboard and extra keys row. `FN` + Left/Right arrows either does not produce expected input or actually does something else (in this case, what exactly?). Hardware keyboard that's another story. I have tried a USB keyboard for PC as well as bluetooth keyboard made for Nexus 7 - both produced *different* behavior on `FN` + Left/Right combination, but neither was one that I expect (when `FN` is pressed, left/right send home/end respectively). Related issue https://github.com/termux/termux-app/issues/731 (closed 3 years ago).

RalfWerner:
> Video appears to be about older Termux version

Right - the PR was closed yesterday and treated `KEYBOARD` Extra Key and BT/USB-KB together with OS-KB. I have no differences between BT/USB but discovered some between AVD and real device and the behavior of the latest version (below) is not different.

Since no one suffers from the _confusing behavior_ or has _not_ such a tablet device, it is still unchanged and probably remains so until other users suffer from it too.

agnostic-apollo:
>is there any way an app can read termux-app's Sharedpreferences set in ./termux/termux.properties ?

@suhan-paradkar That file stores `SharedProperties`, not `SharedPreferences`. If you are going to declared `sharedUserId=com.termux` in `Termux:Wayland`, then you will have direct access. If not, then you can get the file content in two ways. Either via `SAF` by letting user select file/directory and storing [`Uri` permanently](https://developer.android.com/training/data-storage/shared/documents-files#persist-permissions), but of course this will need UI interaction. Or possibly better background way would be to use `RUN_COMMAND` intent background mode. Use `PendingIndent` mechanism to read the file via `cat`, then store in your own data directory with a call to `FileUtils.writeStringToFile()`, then pass to `SharedProperties`. This way you can read any termux file if needed, even outside `$HOME`.

https://github.com/termux/termux-app/wiki/RUN_COMMAND-Intent

suhan-paradkar:
Ah.. thanks...

RalfWerner:
@agnostic-apollo [again](https://github.com/termux/termux-app/pull/2237#issuecomment-903672221) - Do you also get _**Termux crash**_ when tapping on _**`DRAWER` [Extra Key](https://github.com/termux/termux-app/pull/2237#issuecomment-904023260) (`PASTE` missing [here](https://wiki.termux.com/wiki/Touch_Keyboard#Extra_Keys_Row))?**_

agnostic-apollo:
The `termux-app` [sets a `FocusChangeListerner`](https://github.com/termux/termux-app/blob/v0.117/app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java#L531) for the `TerminalView` to automatically open keyboard when [it is selected](https://github.com/termux/termux-app/blob/v0.117/app/src/main/java/com/termux/app/terminal/io/TerminalToolbarViewPager.java#L106). The `LorieView` does not.

RalfWerner:
Thanks for the hint. @suhan-paradkar  had already noticed that and I hope he find a solution.
Commit above was about the _**fling**_ behave here. Without OS-KB, there is no use - _**why not direct open OS-KB after first EK remove?**_ Thereafter, with the first _tab/press_ clearly in which mode I am. Depending on the OS-KB, the button can be here corrected before _release_ or canceled when the finger is moved out of the OS-KB (typical behavior).

For some of my EK buttons, this is easily possible by switching in _**fling**_ after _press_, even if it is not completed (<50%). Back then does not work and after _popup_ no more. two Examples with `nano check.c`:
- `{macro:END,popup:HOME} After press on END (with CTRL/^ fixed), this can be broken with a small left/right finger movement. HOME does not work anymore but it could be redevised on END. If I'm at position 79 in a table with 125 lines (will be displayed with END) and does not want to start the beginning (HOME show line 1) I can cancel this but only before _popup_.
- `{key:LEFT,popup:BKSP}` LEFT has autorepeat so the finger movement has quickly to executed. In the case _up_ I can only dodge to _release_ on LEFT. This is usually no problem or you use autorepeat LEFT and move _finger up_ at the desired position.

PR #2458, opened 2021-12-08
Am integration into Termux

tareksander:
This PR includes:
- a UNIX socket interface for termux-shared (LocalFilesystemSocket.java)
- an interface to handle socket connections on a separate thread and a simple protocol to use CLI-like functions (parameters as a String array, stdout, stderr) (LocalSocketListener.java)
- a simpler implementation of `am` using the Android API (Am.java)
- Creating a socket at startup and handling connections using the new `am` implementation

The new `am` implementation lacks some features (some warnings can't be displayed because the Android API doesn't give them), but is about 10x faster, because there is no need to create a dalvik VM for `am` to run, the process is already there.  
Because the new implementation isn't a 1:1 copy of `am` it shouldn't replace `am`, but can be used as a replacement if tested, e.g. for termux-wake-lock, termux-notification... to get better speed.

More details can be found in [this PR](https://github.com/termux/termux-api-package/pull/147) that gave me the idea.


tareksander:
The test seems to fail because the native library for termux-shared cannot be found during the test. It works fine when running though.

agnostic-apollo:
I am currently busy with work and lot of other stuff so will look into this in next few days. This and other changes in `termux-api` require lot of testing as well so its time consuming stuff. This will have to merged after I have pushed my local changes to master anyways since I am far ahead of master and there are conflicts with this pull request, which would be easier to resolve here. Additional work in master is left as well.

The new design itself is good and  should be merge-able, just need time.

Grimler91:
@tareksander so with this, and the corresponding changes in [termux-api changes](https://github.com/termux/termux-api/pull/471), [termux-am-socket](https://github.com/termux/termux-am-socket) and the [termux-api-package changes](https://github.com/termux/termux-api-package/pull/147), we can drop [termux-am](https://github.com/termux/TermuxAm)?

Just want to check so that I understand it correctly (and if yes, then dependency of termux-tools on termux-am needs to be removed, or rather changed to termux-am-socket)

agnostic-apollo:
In my opinion `termux-am` should be kept at `$PREFIX/bin/am` for backward compatibility and in case `termux-am-socket` doesn't work. We can **add** the dependency of `termux-am-socket` for `termux-tools`.

tareksander:
> @tareksander so with this, and the corresponding changes in [termux-api changes](https://github.com/termux/termux-api/pull/471), [termux-am-socket](https://github.com/termux/termux-am-socket) and the [termux-api-package changes](https://github.com/termux/termux-api-package/pull/147), we can drop [termux-am](https://github.com/termux/TermuxAm)?

I made the changes to termux-api to make it faster, but then I realized that `am` is the main cause of the slow commands, and since other commands like `termux-wake-lock` etc. also use `am`, integrating `am` into Termux could also make them faster. The changes to termux-api aren't needed if this is merged, except maybe the service users can start with `am` to let Android keep the process around for longer, so there is no need to create a new Android process every now and then, which also slows it down.  
  
Replacing the use of `am` with `termux-am` has to be tested, but that can also be done after this is merged, there is no need to use it right away.

> In my opinion `termux-am` should be kept at `$PREFIX/bin/am` for backward compatibility and in case `termux-am-socket` doesn't work. We can **add** the dependency of `termux-am-socket` for `termux-tools`.

I'm also against removing `am` completely, because the implementation inside Termux has a bit less informational warnings and lacks the warnings for Activity starts, because the Android API provides no way to get these warnings.

Grimler91:
Alright, I have been confusing the two features here, we have:

Integration of termuxAm into termux-app:

* this PR makes termux-app listen for api commands sent through a new socket (currently in /data/data/com.termux/files/sockets/am-socket)
* [termux-am-socket](https://github.com/termux/termux-am-socket) is the userspace program that passes termux-api-package script commands to the socket
* the termux-api-package scripts then only need to be changed so that they use [termux-am instead of am](https://github.com/tareksander/termux-api-package/commit/c5f2567b0377b3b5ec240dd31ea9c51c4b5ca5fe)

We then also have the improvement of making termux-api use a unix socket instead of (termux-)am broadcast:

* https://github.com/termux/termux-api/pull/471 adds a service to termux-api that listens for commands at com.termux.api://listen 
* the corresponding [termux-api-package PR](https://github.com/termux/termux-api-package/pull/147) makes $PREFIX/libexec/termux-api first try to send commands to this socket, and if that fails falls back to (termux-)am

Please let me know if I am totally misunderstanding how something here works

tareksander:
Basically you got everything right, but this PR can not only be used with termux-api-package, but can also speed up the communication with the Termux app for things like termux-wake-lock, termux-wake-unlock ..., basically all communication that uses intents. That's why I would prefer this to be merged, as it is a more general solution.

Grimler91:
@tareksander aha, great!

I have been using a build with this PR, and the termux-api{,-package} PRs, merged for the last ~two weeks and haven't noticed any issues, so LGTM. I'll merge the other parts, and upload the termux-am-socket package. Leaving this PR for @agnostic-apollo to merge though, as it might conflict will the other pending changes

agnostic-apollo:
@tareksander Is there a specific reason why you have enforced a `1000s` custom deadline for `am` server? Do you think client will take too long or won't send data after connecting? And where did you decide or get the `100ms` timeout and `1000ms` deadline values, there are very low, under heavy system load, data transfer will definitely not be fast. In AOSP, its a few seconds at least.

`NativeCrashListener` has timeout at `10s`, `logd` at `32s`, bluetooth audio `a2dp` at `5s`, some have it at `1s`, but those are mostly system daemons which would have much higher priorities than app processes.

https://cs.android.com/android/platform/superproject/+/android-12.0.0_r32:frameworks/base/services/core/java/com/android/server/am/NativeCrashListener.java;l=55

https://cs.android.com/android/platform/superproject/+/android-12.0.0_r32:system/logging/logd/LogUtils.h;l=28

https://cs.android.com/android/_/android/platform/packages/modules/Bluetooth/+/f2047664a24b859a729608d807a689c7fc7b6803:system/audio_a2dp_hw/src/audio_a2dp_hw.cc;l=60

tareksander:
I wanted some sort of timeout so if something goes wrong, it doesn't block until the Termux process is restarted. I just picked some timeout with no particular reason, you can set it higher. 1s was more than enough for me, but that was on an emulator and with practically no load.

agnostic-apollo:
Okay. Yeah, will set it at a higher value.

agnostic-apollo:
Hey @tareksander and @Grimler91, so did some refactoring and some [variable name changes](https://www.reddit.com/r/ProgrammerHumor/comments/u1au9h/how_to_justify_foo_and_bar). Also pushed changes to https://github.com/termux/termux-am-socket

Read the commit messages for details, specially b2a64879, 30548360 and 4392e57b. Test is passing now too.

This should be tested and reviewed before its merged.

You can get apk from https://github.com/termux/termux-app/actions/runs/2181520612 and `termux-am-socket` `aarch64` is attached. I can push a new release for it too if needed. If someone wants a different arch, let me know.

I tested from within termux and root and it worked on  Android 7 and 11. From other app with `rwxrwxrwx` permissions for socket file and directory and selinux disabled, connection still failed. Tried `200` `RUN_COMMAND` intents in a for loop and they all succeeded, backlog `50` is not really used by linux as is. https://stackoverflow.com/questions/5111040/listen-ignores-the-backlog-argument

[termux-am-socket_1.02_aarch64.zip](https://github.com/termux/termux-app/files/8502682/termux-am-socket_1.02_aarch64.zip)

tareksander:
Great! Making the socket API more general is something I would have needed for the new plugin system anyways. I'll try and do better with variable naming in the future, this is the first time I really coded something others have to understand outside of class. I can't test anything new, my devices are also Android 7 and 11. I do have some Android emulators set up though, but those are x86.

agnostic-apollo:
>this is the first time I really coded something others have to understand outside of class.

Don't feel taken down by the refactor even a bit, it was seriously a great first version, specially if it was the first time after uni. As for the non-general design, that was an issue but writing better designs takes experience and time to learn, read other people's code and see how they do it, eventually you get better. One should also think a few steps ahead for how else the code may be used in future, so that you need fewer changes in future and don't waste time making lot of changes. Even current design may have issues, I guess time will tell.

>I'll try and do better with variable naming in the future

Yeah, variable names should be more descriptive. Deciding on variable names is like one of the hardest and time taking part.

> I can't test anything new, my devices are also Android 7 and 11

The testing is done for missed cases that I may not have tested or for device or build specific issues. I will do a basic test on android 12 and 13 as well in avd.

agnostic-apollo:
>After Installing latest termux-am-socket (and termux-api, and termux-api-package), and then installing app from this branch I get an error on socket creation which I can't seem to figure out the reason for:

Proguard was removing specific methods only used by JNI for `JNIResult` class, hence the issue. Fixed it with 45450eb. Would have caught this in my pre-release check, but good that its fixed now. Thanks for that! I tested release build on Android 13 with termux user and root and both are working. Also attached `x86_64` version above for avd testing.

The ` public JniResult(int retval, int errno, String errmsg, int intData)` method is missing.

https://github.com/termux/termux-app/blob/3054836049fc4b5ef38e3b3a472a0564df520346/termux-shared/src/main/java/com/termux/shared/jni/models/JniResult.java

https://github.com/termux/termux-app/blob/3054836049fc4b5ef38e3b3a472a0564df520346/termux-shared/src/main/cpp/local-socket.cpp#L175

```
public class JniResult {
    public String errmsg;
    public int errno;
    public int intData;
    public int retval;

    public JniResult(int i, int i2, String str) {
        this.retval = i;
        this.errno = i2;
        this.errmsg = str;
    }

    public JniResult(String str, Throwable th) {
        this(-1, 0, Logger.getMessageAndStackTraceString(str, th));
    }

    public static String getErrorString(JniResult jniResult) {
        return jniResult == null ? "null" : jniResult.getErrorString();
    }

    public String getErrorString() {
        StringBuilder sb = new StringBuilder();
        sb.append(Logger.getSingleLineLogStringEntry("Retval", Integer.valueOf(this.retval), "-"));
        if (this.errno != 0) {
            sb.append("\n");
            sb.append(Logger.getSingleLineLogStringEntry("Errno", Integer.valueOf(this.errno), "-"));
        }
        String str = this.errmsg;
        if (str != null && !str.isEmpty()) {
            sb.append("\n");
            sb.append(Logger.getMultiLineLogStringEntry("Errmsg", this.errmsg, "-"));
        }
        return sb.toString();
    }
}
```

And `termux-api-package` is not using `termux-am-socket` currently. Its using abstract namespace socket started by `termux-api`. That can be removed now.

https://github.com/termux/termux-api-package/blob/v0.57/termux-api.c#L25

https://github.com/termux/termux-api/blob/ba2836ba60083b54dcdb6a3a926777f3f0704829/app/src/main/java/com/termux/api/SocketListener.java#L23

Grimler91:
> > After Installing latest termux-am-socket (and termux-api, and termux-api-package), and then installing app from this branch I get an error on socket creation which I can't seem to figure out the reason for:
> 
> > `IS_DEBUGGABLE_BUILD`: `false`
> 
> Proguard was removing specific methods only used by JNI for `JNIResult` class, hence the issue. Fixed it with [45450eb](https://github.com/termux/termux-app/commit/45450eba29fcfab15ede4c451c60718f19bfefd2). Would have caught this in my pre-release check, but good that its fixed now. 

Thanks! Seems to work fine both on my aarch64 phone and an arm tablet as normal user

Running a termux-api command from a su shell seem to work as well, but not when run from a tsu shell (then it just hangs), due to some env var I guess.

> > `INCREMENTAL`: `eng.grimle.20210821.092214`
> 
> And someone builds their own lineageOS. :D I need to start doing that, but first need to get a 1TB SSD. :p

Yeah, so I have only myself to blame for all the bugs in the ROM ;) yeah, it takes an ridiculous amount of disk space, especially for later android versions..

agnostic-apollo:
>Thanks! Seems to work fine both on my aarch64 phone and an arm tablet as normal user

Great.

>tsu shell (then it just hangs), due to some env var I guess

I installed `tsu` on android 11 and it worked fine from within `tsu` shell and with `sudo termux-am`. Since `tsu` is so broken, try [this `sudo`](https://github.com/agnostic-apollo/sudo), its a very good script, from some random guy ;)

I guess pull request should be mergable, unless someone has an objection.

>Yeah, so I have only myself to blame for all the bugs in the ROM ;)

Your security patch is even older than mine! :p

> yeah, it takes an ridiculous amount of disk space, especially for later android versions..

Yeah, that's why I am gonna wait for the 1TB SSD, gonna need lot of space to build different test versions.

agnostic-apollo (approved):

PR #2550, opened 2022-01-29
TermuxFileReceiverActivity is using wrong path for "usr/bin/"

Worldwidebrine:
I also removed `android.intent.action.VIEW` because it's useless and annoying.
Everytime I open a file of new type, "Open with Termux" shows up, even APK.
We already have `android.intent.action.SEND` so I remove it.


xeffyr:
We intentionally use `$HOME/bin` for scripts. `$PREFIX` is not meant to be directly managed by user.

Removing the intent filter breaks the file receiver completely. While it could be "annoying" for you, its purpose to make Termux as generic file handler, so it could read any type of files.

Rejecting the PR.

PR #2810, opened 2022-05-29
Changed: RunCommandService Refactored

rubinstory:
Changed: RunCommandService Refactored

### DIFF
--- a/app/src/main/java/com/termux/app/RunCommandService.java
+++ b/app/src/main/java/com/termux/app/RunCommandService.java
@@ -36,6 +36,7 @@ import com.termux.shared.shell.command.ExecutionCommand.Runner;
 public class RunCommandService extends Service {
 
     private static final String LOG_TAG = "RunCommandService";
+    public static final int GRAY = 0xFF607D8B;
 
     class LocalBinder extends Binder {
         public final RunCommandService service = RunCommandService.this;
@@ -272,7 +273,7 @@ public class RunCommandService extends Service {
         builder.setSmallIcon(R.drawable.ic_service_notification);
 
         // Set background color for small notification icon
-        builder.setColor(0xFF607D8B);
+        builder.setColor(GRAY);
 
         return builder.build();
     }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/termux/app/RunCommandService.java
```diff
@@ -71,125 +72,63 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
-        // If EXTRA_RUNNER is passed, use that, otherwise check EXTRA_BACKGROUND and default to Runner.TERMINAL_SESSION
-        executionCommand.runner = IntentUtils.getStringExtraIfSet(intent, RUN_COMMAND_SERVICE.EXTRA_RUNNER,
-            (intent.getBooleanExtra(RUN_COMMAND_SERVICE.EXTRA_BACKGROUND, false) ? Runner.APP_SHELL.getName() : Runner.TERMINAL_SESSION.getName()));
-        if (Runner.runnerOf(executionCommand.runner) == null) {
-            errmsg = this.getString(R.string.error_run_command_service_invalid_execution_command_runner, executionCommand.runner);
-            executionCommand.setStateFailed(Errno.ERRNO_FAILED.getCode(), errmsg);
-            TermuxPluginUtils.processPluginExecutionCommandError(this, LOG_TAG, executionCommand, false);
-            return stopService();
-        }
+        Integer invalidRunner = checkRunner(intent, executionCommand);
```
agnostic-apollo, line 86: Could just return `false` from function and do

`if(!checkRunner()) return stopService()`

agnostic-apollo:
Also the `RunCommandService` is gonna be refactored and split soon for new API design, so above changes will be lost.

rubinstory:
@agnostic-apollo 
We didn't know that this file will be refactored and splitted, so we'll just close this PR.

agnostic-apollo:
Apart from `checkInvalidAction()` extraction, these were reasonable changes, although sequence of function definitions should have been the same as they were called. Split will indeed remove some stuff, specially intent extraction and creation, I currently haven't decided on final design, but already have a local implementation.

PR #2817, opened 2022-05-29
Changed: Refactor LocalClientSocket

rubinstory:
Changed: Refactor LocalClientSocket

### DIFF
--- a/termux-shared/src/main/java/com/termux/shared/models/ReportInfo.java
+++ b/termux-shared/src/main/java/com/termux/shared/models/ReportInfo.java
@@ -84,9 +84,19 @@ public class ReportInfo implements Serializable {
 
         if (reportInfo.addReportInfoHeaderToMarkdown) {
             markdownString.append("## Report Info\n\n");
+<<<<<<< HEAD
             markdownString.append("\n").append(MarkdownUtils.getSingleLineMarkdownStringEntry("User Action", reportInfo.userAction, "-"));
             markdownString.append("\n").append(MarkdownUtils.getSingleLineMarkdownStringEntry("Sender", reportInfo.sender, "-"));
             markdownString.append("\n").append(MarkdownUtils.getSingleLineMarkdownStringEntry("Report Timestamp", reportInfo.reportTimestamp, "-"));
+=======
+
+            for (Pair<String, Object> logVar: getLogVariableList(reportInfo)) {
+                String label = logVar.first;
+                Object object = logVar.second;
+                markdownString.append("\n").append(MarkdownUtils.getSingleLineMarkdownStringEntry(label, object, "-"));
+            }
+
+>>>>>>> 1677e24... Changed: Add explaining variables to `getLogString()` and `getMarkdownString()` in `ReportInfo.java`, `LocalClientSocket.java` and `LocalSocketRunConfig.java`.
             markdownString.append("\n##\n\n");
         }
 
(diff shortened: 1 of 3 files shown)


agnostic-apollo:
This wasn't properly rebased and you are pushing merge commits to all your pull requests.

Secondly, with this `getLogVariableList()` you are doing everywhere is requiring case statements to handle special cases and would fail for complex cases like [`ExecutionCommand`](https://github.com/termux/termux-app/blob/231ecff5f0837d1aa2831039b64afd07915e1cab/termux-shared/src/main/java/com/termux/shared/shell/command/ExecutionCommand.java#L378).

Instead, create specialized classes `LogStringBuilder` and `MarkdownStringBuilder` under `com.termux.shared.data` package that stores a `StringBuild` object. This class provides append methods for `Logger.getSingleLineLogStringEntry()` and `MarkdownUtils.getSingleLineMarkdownStringEntry()` and multiline variants. So something like

```
StringBuilder logString = new StringBuilder();
logString.append("\n").append(Logger.getSingleLineLogStringEntry("Path", mPath, "-"));
```
would be like

```
LogStringBuilder logString = new LogStringBuilder();
logString.appendSLEntry("Path", mPath);
```

The function will automatically add `\n` and use `-` as default, you can also add another function that allows default value to be passed. A simple `append()` method without any newline addition should also be added, also `appendTitle()` and `appendSubtitle()` functions.

Both the classes could actually implement an interface `ObjectStringBuilder` that contains all the functions, this could allow something like `JsonStringBuilder` and `YamlStringBuilder` to be added in future as well. Some simple implementations for objects of log and markdown strings could also be simplified into a single function by passing `ObjectStringBuilder` as parameter to the function like following...

```
getLogString() {
    getObjectString(new LogStringBuilder());
}

getLogString() {
    getObjectString(new MarkdownStringBuilder());
}

getObjectString(ObjectStringBuilder builder) {
    builder.appendSLEntry("Path", mPath);
}
```

You can create a single pull request for the log and markdown string replacements.

rubinstory:
@agnostic-apollo 
Thanks for reviewing. We'll  fix our commits with correctly rebasing and applying those Builder pattern you've said.

agnostic-apollo:
You are welcome. Let me know if you need any help or something is unclear.

rubinstory:
@agnostic-apollo 
Currently, I've almost finished adding `ObjectStringBuilder` and its child classes. But I can't understand the usage of `appendSubtitle()` function. Could you give me some examples?

agnostic-apollo:
Sometimes there are sub sections that have  

with nested keys

https://github.com/termux/termux-app/blob/231ecff5f0837d1aa2831039b64afd07915e1cab/termux-shared/src/main/java/com/termux/shared/net/socket/local/LocalSocketRunConfig.java#L215

https://github.com/termux/termux-app/blob/231ecff5f0837d1aa2831039b64afd07915e1cab/termux-shared/src/main/java/com/termux/shared/net/socket/local/LocalSocketManager.java#L395

or content

https://github.com/termux/termux-app/blob/231ecff5f0837d1aa2831039b64afd07915e1cab/termux-shared/src/main/java/com/termux/shared/shell/command/ExecutionCommand.java#L500

For markdown, heading size depends on `#` used and I close sections with `\n##\n\n` for horizontal bar.
For log strings, `:` is being used, and closing with `\n\n\n`.

For json and yaml, this could be implemented as keys instead, but that would require proper opening and closing.
For markdown, nesting could increasing number of `#`, like execution command does.

I haven't thought this too much, there is probably a way for all this. You can skip it for now. We can hard code it for markdown's case for now, since execution command doesn't log that stuff for log string and only markdown.

agnostic-apollo:
Hey @rubinstory, was there any progress on this? `ExecutionCommand` is becoming more and more of a mess with things being added and I am thinking on adding `ObjectStringBuilder` now. Even if its not complete, just send a patch file and I'll look into merging it locally. Otherwise, will just implement it myself, no worries. Thanks

PR #2851, opened 2022-06-20
Basic MIME type recognition in ContentProvider

TpmKranz:
This fixes apps that rely on [getType](https://developer.android.com/reference/android/content/ContentResolver#getType(android.net.Uri)), for example [WallpaperPicker2](https://cs.android.com/android/platform/superproject/+/master:packages/apps/WallpaperPicker2/src/com/android/wallpaper/asset/ContentUriAsset.java;l=158;bpv=1;bpt=1?q=isJpeg).


fornwall:
Thanks!

PR #2921, opened 2022-08-05
Plugins without shared UID

### LINKED ISSUE #2654: [Feature]: Plugins without shared UID
### Feature description

For security reasons (privilege separation) and because f-droid seems to be slow when it comes to setting up signing apps with the same key (see Termux:GUI) it could be useful to have a plugin that doesn't share the same UID. The main problem with that would be security: If the apps are different users, they cannot call not exported components of each other, and completely open components are a security risk.

My idea is: Create a new signature-level permission for Termux and have the plugins BroadcastReceiver or Service require that permission to get called. That way only Termux can call these components.

To establish a 2-way communication between a plugin and a program you could make Termux expose another unix socket to programs like for the `am` integration. The program sends Termux the component it would like to call and if successful Termux sends a file descriptor for a unix socket pair to the program and the other end is send in the Intent via ParcelFileDescriptor to the plugin. If more connections are needed, the program or plugin can then send more file descriptors over the connection they have already.

### Additional information

[Create a ParcelFileDescriptor socket pair.](https://developer.android.com/reference/android/os/ParcelFileDescriptor#createSocketPair())  

[Getting a FileDescriptor from a ParcelFileDescriptor.](https://developer.android.com/reference/android/os/ParcelFileDescriptor#getFileDescriptor())  

[Sending a FileDescriptor over a LocalSocket.](https://developer.android.com/reference/android/net/LocalSocket#setFileDescriptorsForSend(java.io.FileDescriptor[]))  

[Putting a Parcelable into an Intent.](https://developer.android.com/reference/android/content/Intent#putExtra(java.lang.String,%20android.os.Parcelable))

tareksander:
This adds a bound service plugins without a shared UID can use.  
closes #2654

How it works:  
- Plugins can connect to the service
- After connecting, they have to register themselves as a client by providing a Binder for callbacks (and for Termux to watch when the plugin process dies, so resources can be released)
- After that, the plugin can call the other methods of the service

As an alternative to providing a Binder, clients can also specify a bound service Termux will then bind to to call the callback methods. Plugins can use this prevent themselves from getting killed in the background without having to clutter the notification area with another foreground service notification.

Security model:  
- From the Termux side:
  - The bound service is protected with a new dangerous permission for plugins, so only apps the user granted the permission can use the service
  - All file methods check if the path is in the plugin directory, so no arbitrary file read/write can occur
  - The runTask method is additionally secured with the RUN_COMMAND permission
  - The service keeps track of which plugin started which task, and plugins can only send signals to their own tasks
- From the plugin side:
  - I made a [library](https://github.com/tareksander/plugin-template) that abstracts away the lower level communication with the service a bit and adds security checks
    - Before connecting to the service, it checks if the signature for the Termux app matches one of the google play, github or fdroid signatures.
    - The callback binder checks if the calling app is Termux on the first transaction, and also checks the Termux app signature
    - Callback services can be secured with a new signature permission Termux defines. This ties the signature into the Android permission system. After getting called, the plugin also has to verify the signature of the Termux app, to tie the permission to the actual signature. This is done in the callback binder by default.

I tested the security aspects, paths outside the plugin directory are rejected, absolute paths are interpreted with the plugin directory as the root. Limiting runTask to apps that have the RUN_COMMAND permission also works. Plugins refuse to connect to Termux if the signature doesn't match.

Features:
- Each plugin gets a directory with its package name under /data/data/com.termux/files/apps/plugins and some methods to manage it
  - Opening files for reading and/or writing
  - Creating and listening on socket files
- If the plugin has the RUN_COMMAND permission: A new way to run tasks, using pipes for arbitrarily long input and output
  - A way to send signals to the task
  - A callback for when the task exits/is killed

The features are deliberately minimal and designed to work with programs in Termux to use the sockets and files.

To discuss:
- Should there be more methods to manage the plugin directory?
  - rmdir/mkdir?
  - rm?


agnostic-apollo:
> @agnostic-apollo isn't even needed.

Things specially like this with security issues don't get merged without proper review and testing.

I did review it couple of weeks back and even with a quick look I did, it had quite a few issues, which I need to fix before this can be merged. I also have to push my local changes first, some of which are required by this pull, which I will do in next few days, mostly done with what I had to do, but lot of testing is required due to tonne of changes and some docs updates.

agnostic-apollo:
>I thought if it's bad, then it just gets rejected.

This pull request is not bad that it gets rejected/closed. The design should likely work and overall good work by @tareksander, but needs a deeper look. There are just issues that need to be fixed, initial implementations usually do have issues.

khimaros:
@agnostic-apollo @tareksander is this still happening?

was this fix committed in another PR?

according to https://github.com/termux/termux-gui/issues/4#issuecomment-1635968056 this should fix https://github.com/termux/termux-gui/issues/4

agnostic-apollo:
Read my comment above. And no, not merged elsewhere.

termux-shared/src/main/java/com/termux/shared/android/BinderUtils.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * @param c The context used to get the {@link android.content.pm.PackageManager}.
+     * @return The plugin directory of the calling package name under {@link TermuxConstants#TERMUX_APPS_DIR_PATH}, or null if {@link #getCallerPackageNameOrNull(Context)} returns null.
+     */
+    public static String getCallingPluginDir(Context c) {
+        String packageName = getCallerPackageNameOrNull(c);
+        if (packageName == null) {
+            return null;
+        }
+        return TermuxConstants.TERMUX_PLUGINS_DIR_PATH + "/" + packageName;
```
agnostic-apollo, line 61: This will cross `UNIX_PATH_MAX` (`108`) for long package names of termux forks or external apps. Check `TERMUX__APPS_DIR_MAX_LEN` and `TERMUX__ROOTFS_DIR_MAX_LEN` at following links. The sub directory hierarchy needs to change too.  I haven't given much thought into it, but it could be handled by using package name itself and seeing if limit will cross for socket file, and if it does, then generate a shorter random identifier. Whatever is chosen on first connection will be saved in shared preferences, and the selected app data directory can be queried by the plugin app.

- https://github.com/termux/termux-packages/wiki/Termux-file-system-layout#file-path-limits
- https://github.com/termux/termux-exec/pull/24#issuecomment-1873059748

tareksander: I don't think it's a major problem: `/data/data/com.termux/files/apps/plugins/very.long.termux.plugin.name.with.a.filesystem.socket/socket.sock` would be the limit, but maybe we could drop the `plugins` from the path to get a few more characters, package names are unique anyways. Or move the while thing up and even drop the `files`. And if it gets extreme you could only use one letter for the socket name instead of something more descriptive.

tareksander: When you generate short randomized names instead of the package names, one problem would be that the client binaries also need to get the name. You'd have to put a mapping file somewhere for them to read.

agnostic-apollo: > The `TERMUX__APPS_DIR_MAX_LEN` is chosen as `84` based on the `12`, `13`, `14`, `16`, `17` and `18` examples.

Check them, it will easily cross. And termux forks matter too, so `/data/data/very.long.termux.fork.name/` matters too. And `/data/user/10` for secondary users or `/mnt/expand` for adoptable users are far more easy to cross. I need to think about every possible use case and for any future changes, not just current termux one.

agnostic-apollo: >You'd have to put a mapping file somewhere for them to read.

You mean like `/data/data/com.termux/shared_prefs/plugin-apps-mappings.xml`? :p

I hadn't thought it from that case, but will have to think on it too, lots to think about and do.

agnostic-apollo: So dumbass me figured out an alternative to app package names (for `TERMUX__APPS_DIR_BY_PACKAGE`) that could be used that would never cross the max `11` digit limit for an identifier. What is another thing that all apps have unique to them, other than their package names, that could also be found easily by any native process (without hardcoding it or even without requiring apps to export their package names as env variables)?

tareksander: The uid?

agnostic-apollo: Yup. The max possible `uid` (`id -u`/[`ApplicationInfo.uid`](https://developer.android.com/reference/android/content/pm/ApplicationInfo#uid)) would be `100010xxx` for an app installed on user `1000`, so at max `9` characters would be used, so lower than the max `11` characters allowed for an app identifier. There will be two directories `TERMUX__APPS_DIR_BY_IDENTIFIER=/data/data/com.termux/termux/apps/i/<app_identifier>` and `TERMUX__APPS_DIR_BY_UID=/data/data/com.termux/termux/apps/u/<app_uid>` under which apps can create their files. Termux apps itself can use the `TERMUX__APPS_DIR_BY_IDENTIFIER` with the identifiers like `termux` and `termuxfloat`. Plugin apps can use the `TERMUX__APPS_DIR_BY_UID` directory, their app and native processes would already know their uid, so for "Opening files for reading and/or writing" and "Creating and listening on socket files" they would know the path, and termux app can validate the `ApplicationInfo.uid` against the uid path, as uids may change if app is reinstalled and termux shouldn't allow access to paths for older uid in case it now belongs to some other app. `TERMUX__APPS_DIR_BY_IDENTIFIER` could also be used by plugin apps, but termux should store the package names against their app identifier permanently in the shared preferences so that no other app tries to use an identifier of another app, a UI could be provided in Termux app settings to remove the saved values though. Apps that have `RUN_COMMAND` permission, they would be able to access any paths anyways, so shouldn't matter.

Do you see any problem with this, since I haven't fully checked your requirements?

---

The user names (`id -un`) for app processes are in the following formats:
- `u<user_id>_a<app_id>` for normal app processes, like `u0_a160`.
- `u<user_id>_i<app_id>` for isolated app processes, like `u0_i160`.

The uid (`id -u`) for apps are calculated as per:
`user_id * AID_USER_OFFSET + AID_APP_START + app_id`, where:
- `AID_USER_OFFSET=100000` (offset for uid ranges for each user)
- `AID_APP_START=10000` (first app user)
- `AID_APP_END=19999` (last app user)
- The `user_id` refers to the id for the user in which the app is
  installed. The default/primary user id is `0`. The ids for
  secondary users and profiles start at id `10`.
- The `app_id` is the unique id assigned to an app at install time
  by Android. If the same app is installed in secondary users, it
  will be assigned the same `app_id`, but `uid` will have a different
  `user_id` to distinguish them and so will the categories in the
  SeLinux contexts assigned to the processes and files of the apps
  as part of Multi-Category Security (MCS).
- https://cs.android.com/android/platform/superproject/+/android-14.0.0_r1:system/core/libcutils/include/private/android_filesystem_config.h;l=197
- https://cs.android.com/android/platform/superproject/+/android-14.0.0_r1:bionic/libc/bionic/grp_pwd.cpp;l=274
- https://github.com/agnostic-apollo/Android-Docs/blob/master/site/pages/en/projects/docs/os/selinux/security-context.md#categories

For example, for normal app processes with the `app_id=160`:
- `user_id=0: (0 * 100000 + 10000 + 160) -> 10160/u0_a160`
- `user_id=10: (10 * 100000 + 10000 + 160) -> 1010160/u10_a160`
- `user_id=256: (256 * 100000 + 10000 + 160) -> 25610160/u256_a160`
- `user_id=1000: (1000 * 100000 + 10000 + 160) -> 100010160/u1000_a160`

A `user_id` can have a max `1000` value, so should use max `4` characters.
https://cs.android.com/android/platform/superproject/+/android-14.0.0_r1:bionic/libc/bionic/grp_pwd.cpp;l=351
- https://cs.android.com/android/platform/superproject/+/android-14.0.0_r1:system/core/libcutils/multiuser.cpp;l=29

However, only `1-10` users are allowed to be created normally, based
on the `fw.max_users` property or `config_multiuserMaximumUsers`
config (`pm get-max-users`), so should only use 2 characters normally.
- https://source.android.com/docs/devices/admin/multi-user#applying_the_overlay
- https://cs.android.com/android/platform/superproject/+/android-14.0.0_r1:frameworks/base/core/res/res/values/config.xml;l=2802
- https://cs.android.com/android/platform/superproject/+/android-14.0.0_r1:frameworks/base/core/java/android/os/UserManager.java;l=5719

Rudxain: What about hashing the pack-name (identifier), then truncating it to 8nibbles, then re-coding it as B64 (6Bytes)? Of course, this would be inconvenient because (just like the randomized case) a file would be needed to disambiguate potential collisions.

Using uids seems like a better solution, because Android guarantees there's no such collision.

But then, why have 2 paths (pack-name and id)? This may be a "stupid question", because I'm writing this before reading the actual code

tareksander: Is there a way to may uids to package names in native code? Otherwise a mapping is still needed. An alternative would be to make the package name directories symlinks to the uid directories, so you can just use the link to look up the right directory.

agnostic-apollo: Why do you need to know the package name? Don't third party apps already know their package names? They can always export an environment variable for it too.

However, termux app will write to an `apps/i/termux/termux-apps.env` file for all the variables of the other termux apps whenever a package is installed/uninstalled/updated. Current way of exporting them when creating shell doesn't work for variables for version names, etc as app could get installed/uninstalled/updated after shell was created, so variables are invalid. So shell scripts will source the file when they want to read an app specific variable. I was thinking of adding variables for currently bound third apps too, so one could find the right variable for the uid of the app by comparing against `id -u`, and then get package name by generating a variable name with the same prefix as uid variable and reading it. A singular variable could also be provided with info for all the apps in some specific format.

tareksander: > Why do you need to know the package name? Don't third party apps already know their package names?

So programs inside Termux that use the plugin know where to look. Third party apps running commands is just one use case.

agnostic-apollo: Then they can do the reverse, find the variable for package name and then its uid and then generate the path, we can provide a helper script for it too. Or a single json variable with package names as keys and nested dict with values like path.

tareksander: I like the idea of a JSON file, it's more useful in case you don't use bash, so you can't just source the env file.

agnostic-apollo: That should be fine for plugins, not for termux apps itself, as parsing json with `jq` in an external call and is heavy and a dependency on the `jq` package, termux-api scripts and others require `ms` level efficiency, so sourcing is significantly faster. Could add info to both an env file and json file and clients can decide themselves.

agnostic-apollo:
I did plan on adding this for `v0.119.0`, but due to likely changes for maintainership of Termux in coming weeks, this (and others pulls) may either get delayed to a future version or may not be added in Termux app itself by me. Will post about the plans when they are more clear.

microbearsmb:
Just curious where things are at now - no pressure at all: is this PR still within the plans for the near future?

agnostic-apollo:
No update for what I said in last comment. When I will have pushed my local changes for next main beta, then will see what the situation is and if it will get implemented or not, will be a while for that, lot of other Termux things are being worked on currently.

tareksander:
No, just me fixing the build for the latest JDK bundled with Android Studio.

PR #2971, opened 2022-09-03
Clear LineWrap flag when clearing lines using blockSet.

MatanZ:
This commit resets a row's mLineWrapped flag, if the last column is cleared with spaces.

Without this the flag remains from a previous line after clearing, for example on escape sequence `\e[2J`.

To see the bug:
1. Open termux (with font large enough for some of the message lines to wrap).
2. run `clear ; ls /`
3. press `ctrl+alt+=` to increase font size.

Notice that the some of the ls lines are now not left aligned.


agnostic-apollo:
Can you provide before and after screenshots showing the issue?

MatanZ:
Start, 1920x1080 screen, terminal size: 27x40.

Note that the 5th and 7th line are wrapped:
Screenshot_1662220318

Results of ls before increasing fonts:
Screenshot_1662220325

Current master: After increasing fonts, 5th and 7th lines spaces wrapped to 6th and 8th lines respectively.
Screenshot_1662220356

With this commit (after the same process): display as expected.
Screenshot_1662221156

agnostic-apollo:
Yeah, I can confirm that there is an issue.

PR #2973, opened 2022-09-04
Add graphics in terminal support: - Sixel and iTerm2 protocols

### LINKED ISSUE #142: Feature Request: sixel graphics mode
https://en.wikipedia.org/wiki/Sixel is a raster protocol format part of VT200 and VT300 serial terminals
it recently started gaining traction on command line tools to output images directly to the terminal because it works trought ssh, works reasonably well, etc 
for an implementation detail to copy the encoder/decoder, one can look at https://github.com/saitoha/libsixel , the url documents also a series of terminals and programs that support the standard
I realize it's a low priority feature, but it'd allow rather nice semi graphical workflow under termux's terminal, currently  it's already possible to view sixel graphics from launching a ssh daemon and connecting externally with a sixel capable terminal emulator

### LINKED ISSUE #3002: [Feature]: Support for CSI 14
### Feature description

CSI 14 is an escape code, that returns the screen size in pixels.

This feature would allow the running application to get the screen size, it is used mainly in sixel supporting apps, that need to scale the image before printing it out. This would be needed if this #2973 would me merged.

Another alternative is TIOCGWINSZ syscall.

### Additional information

alacritty suports it:

https://github.com/alacritty/alacritty/commit/cdf5e51e742d67d1b96940a5096210725a1e51e8

MatanZ:
This commit implements graphics in the terminal, including two different protocols for placing images:
1. [Sixel](https://vt100.net/docs/vt3xx-gp/chapter14.html), and
2. [iTerm2](https://iterm2.com/documentation-images.html)

I ran this for a few days, and it seems to me that some people tried it from #142, and saw no crashes, so I hope it is reasonable safe.

- In TerminalEmulator, interpret sixel sequences, and send them to
  TerminalBuffer for constructing a bitmap.
- Sixel sequences may be longer than 8192 characters, so break them in
  natural places ($,-,#), rather than collecting all in the buffer.
- The bitmap is sliced to character cell sized slices, and each the
  the style attribute is used to store which bitmap slice is displayed
  in place of this character.
- In TerminalRenderer the style is interpreted, and drawn using
  drawBitmap, instead of drawText.

Support iTerm2 inline image protocol (OSC 1337):

- Using the same bitmap display infrastructure introduced for sixels.
- Collects the image data outside of the OSC buffer.
- Ignoring some parameters.

Small emulator changes:

- Also eat APC sequences, not echoing to screen.
- Fix `CSI 14 t` to give actual size
- Add `CSI 16 t`
- Add `4` (sixel) to device attributes

For those who tried this branch already, I force pushed to it  in order to remove changes that may not be acceptable.

The main known issue is that if graphics sequence is interrupted (in specific places), the emulator remains stuck waiting for the sequence to end, thus ignoring all input. This requires terminal reset from the menu.


aicynide:
You are a hero my brother

HardcodedCat:
It works! but gifs are being duplicated on the terminal when expanding and collapsing the keyboard

Regular images are being displayed fine

Screenshot_20220905-194040_Termux

MatanZ:
> It works! but gifs are being duplicated on the terminal when expanding and collapsing the keyboard

This is not a bug. img2sixel displays animations by sending each frame as a different sixel with a command to send the cursor to position (1,1) between them (CSI H). This causes each frame to overwrite the previous one.

When you remove the keyboard, the screen gets taller, and termux handles this by scrolling down, so the previous frame which was at position (1,1), is now lower, so it is not covered by the following frames.

HardcodedCat:
oh, so everything's working fine then

HardcodedCat:
### Here's a small guide for anyone else who wants to test this PR

> For displaying images and gifs using `Sixel`, do `pkg install libsixel` and use `img2sixel image.png`

> For displaying images and gifs using `iTerm2`, download the [imgcat](https://iterm2.com/utilities/imgcat) script, and use it with the command `./imgcat image.png`

HardcodedCat:
> Has issues with big images. For example when using `imgcat` without limiting width Termux app may get crashed due to out-of-memory exception:
> 
> ```
> 09-06 19:10:37.740 10702 10702 E Termux  : java.lang.OutOfMemoryError: Failed to allocate a 521628640 byte allocation with 74375880 free bytes and 232MB until OOM, target footprint 99167840, growth limit 268435456
> 09-06 19:10:37.740 10702 10702 E Termux  :      at com.termux.terminal.TerminalBuffer.resizeBitmap(TerminalBuffer.java:48)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at com.termux.terminal.TerminalBuffer.addImage(TerminalBuffer.java:199)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at com.termux.terminal.TerminalEmulator.doOscSetTextParameters(TerminalEmulator.java:2304)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at com.termux.terminal.TerminalEmulator.doOsc(TerminalEmulator.java:2043)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at com.termux.terminal.TerminalEmulator.processCodePoint(TerminalEmulator.java:577)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at com.termux.terminal.TerminalEmulator.processByte(TerminalEmulator.java:553)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at com.termux.terminal.TerminalEmulator.append(TerminalEmulator.java:502)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at com.termux.terminal.TerminalSession$MainThreadHandler.handleMessage(TerminalSession.java:345)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at android.os.Handler.dispatchMessage(Handler.java:106)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at android.os.Looper.loopOnce(Looper.java:201)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at android.os.Looper.loop(Looper.java:288)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at android.app.ActivityThread.main(ActivityThread.java:7898)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at java.lang.reflect.Method.invoke(Native Method)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
> 09-06 19:10:37.740 10702 10702 E Termux  :      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
> ```
> 
> Test image (7779x4191): https://user-images.githubusercontent.com/107305601/188685258-87eb0704-12c5-4b43-a47a-a1deae99e208.jpg
> 
> Device config: Pixel 5, Android 13, 8 GiB RAM.

Image displays fine on [termux-monet](https://github.com/HardcodedCat/termux-monet/actions/runs/2996331381) with `android:largeHeap="true"` flag. Try adding this flag to your AndroidManifest.xml, it should fix oom exception. largeHeap flag does let applications to use more RAM.

Screenshot_20220906-133931_Termux

cogburnd02:
I believe if [MatanZ](https://github.com/MatanZ) gets this imported and [issue 142](https://github.com/termux/termux-app/issues/142) gets solved, then MatanZ gets my [$100 bounty on Bountysource for that issue.](https://app.bountysource.com/issues/36956941-feature-request-sixel-graphics-mode) It will be well worth it!! :-D

HardcodedCat:
> Thanks, but anyway there is some proper workaround needed to avoid OOM. Larger heap indeed can solve it for a particular case. Not all devices have lots of memory and part is always used by other apps including system, max heap size also vary between device models.
> 
> Unfortunately I can't propose the best way to implement that, but I think refusing to display supersized bitmaps is better than crashing all sessions. For example if dimensions are too big, user can get a toast message describing the issue.
> 
> Command line programs should not be able to crash terminal or cause resource overload by just sending some control sequence (no matter whether it is sixel or something else).

I don't think that's necessary, since most of command line programs always will be able to crash the application when overloaded or abused.

That would only limit how the user should use his terminal, creating a "fake error" that doesn't even exist for some users, like the people who has enough RAM for displaying those images.

Those people who can display the images wouldn't be able to.

But if you guys decide that's the best thing to do, i purpose adding a limit only for devices with less than 4GB RAM.

MatanZ:
> ### Here's a small guide for anyone else who wants to test this PR
> 
> > For displaying images and gifs using `Sixel`, do `pkg install libsixel` and use `img2sixel image.png`
> 
> > For displaying images using `iTerm2`, download the [imgcat](https://iterm2.com/utilities/imgcat) script, and use it with the command `./imgcat image.png`

Just so I can get more testing from the nice people who test my code:

- There are other sixel encoders, for example ImageMagick. Simple example: `convert image.jpg sixel:-`
- gnuplot can also output sixel, but I am not sure if it uses its own encoder, or one of the above. Use `set terminal sixelgd`.
- imgcat has some parameters to control scaling of the image, but you don't even need it. iterm2 protocol is as simple as: `echo -en '\e]1337;File=inline=1;keepAspectRatio=0;height=70%:' ; base64 -w 0 /sdcard/z1.jpg ; echo -e '\e\\'` . You can have width parameter in addition to height.

HardcodedCat:
Tested with imagemagick, and it's displaying fine on my side

MatanZ:
> Thanks, but anyway there is some proper workaround needed to avoid OOM. Larger heap indeed can solve it for a particular case. Not all devices have lots of memory and part is always used by other apps including system, max heap size also vary between device models.
> 
> Unfortunately I can't propose the best way to implement that, but I think refusing to display supersized bitmaps is better than crashing all sessions. For example if dimensions are too big, user can get a toast message describing the issue.
> 
> Command line programs should not be able to crash terminal or cause resource overload by just sending some control sequence (no matter whether it is sixel or something else).

I hope this solves the problem. I put a try{}catch() block around each bitmap operation that allocates memory. This should be similar to what  you suggest - ignoring large images, without hard coding a definition of large.

I believe it is better without a toast notification. Usually  when a terminal cannot (or does not want to) perform a certain operation it just ignores it, without notifying the user. Maybe a line in the log.

agnostic-apollo:
@MatanZ Please use dedicated class for long sixel logic in `TerminalEmulator` and `TerminalBuffer`.

Trying to catch `OutOfMemoryError` and freeing resources "should" work. For normal images, one can display bitmap at lowed res, not sure if something like that can be done with sixel, instead of showing nothing at all.

Checking current memory before even trying to display large image could be done too.

https://developer.android.com/reference/android/app/ActivityManager#getMemoryInfo(android.app.ActivityManager.MemoryInfo)

https://developer.android.com/topic/performance/graphics/load-bitmap#read-bitmap

>Maybe a line in the log.

This is the way. No toasts. Being done in other places too.

rihaq:
@MatanZ brother kitty image protocol when?

MatanZ:
> @MatanZ brother kitty image protocol when?

The way graphics display is implemented, it is not possible to implement kitty protocol. It is possible to implement very simplified ("put image here"), but image and placement management, as well as combining text and images is impossible.

MatanZ:
> @MatanZ Please use dedicated class for long sixel logic in `TerminalEmulator` and `TerminalBuffer`.

Refactored some code to two new classes. I don't see anything in TerminalEmulator that belongs in another class, or which may become clearer.

> Trying to catch `OutOfMemoryError` and freeing resources "should" work. For normal images, one can display bitmap at lowed res, not sure if something like that can be done with sixel, instead of showing nothing at all.
> 
> Checking current memory before even trying to display large image could be done too.
> 
> https://developer.android.com/reference/android/app/ActivityManager#getMemoryInfo(android.app.ActivityManager.MemoryInfo)
> https://developer.android.com/topic/performance/graphics/load-bitmap#read-bitmap

I am not sure this is a good idea. I would have to add assumptions about how much memory is used by the android for various bimap manipulations (decode, scale, resize). I'll settle for trying anything the user asks for, and catching the errors. I did not crash termux with this, with all the above, and various other large images.

agnostic-apollo:
Thanks. Will take another look later myself during merging.

You can also let `OutOfMemoryError` happen and then try to display lower res once and see if `OutOfMemoryError` triggers or not and abort if it did. This way would increase chances for user to get to see something than nothing at all. Of course that would add code, so just a suggestion.

MatanZ:
I cannot see those videos. And can you please include the actual files used, and the commands that cause the problem?

rihaq:
@sylirre Please enable libpng and libjpeg
I think -Dlibpng=enabled
-Dlinjpeg=enabled , right?

rihaq:
@sylirre maybe gdk-pixbuf conflicting with them, not possible to enable both png/jpeg and pixbuf, any 1 should be enabled right?

@sylirre mpv (--vo=sixel) supporting playing video in terminal using sixel but not suxel support not enabled in our termux package https://github.com/termux/termux-packages/blob/master/packages/mpv/build.sh please submit patch with `--enable-sixel` build flag 

Thank you and take care

rihaq:
I think w3m also needed to be built with sixel support

rihaq:
@sylirre there is another package in termux-repo which is not working. Its photon-rss.

leapofazzam123:
- ✅ Works
- ℹ️ Has problems/implementation issues
- ❌ Doesn't work

Sixel coverage:

- ✅ `img2sixel`

- ✅ ImageMagick `convert`

- ✅ Chafa

- ✅ lsix

- ✅ neofetch


Conclusion: All applications that uses and/or supports Sixel works properly 

iTerm2 coverage:
- ✅ Basic `echo` method

- ✅ iTerm2 `imgcat`

- ❌ Chafa
When using `chafa --format iterm2`, Termux just straight up crashes
- ℹ️ neofetch
The image for some reason is in the bottom which shouldn't be

After all of the information is fully displayed, the image for some reason suddenly vanished


Conclusion: Not all applications that uses and/or supports iTerm2 works properly

HardcodedCat:
- Chafa iTerm2 (`chafa --format iterm2`) crash report:

## Report Info


**User Action**: `crash report`  
**Sender**: `TermuxActivity`  
**Report Timestamp**: `2022-09-17 11:30:33.505 UTC`  
##

## Crash Details

**Crash Thread**: `Thread[main,5,main]`  
**Crash Timestamp**: `2022-09-17 11:30:29.822 UTC`  

**Crash Message**:
```
Attempt to invoke virtual method 'int android.graphics.Bitmap.getWidth()' on a null object reference
```


### Stacktrace

```
java.lang.NullPointerException: Attempt to invoke virtual method 'int android.graphics.Bitmap.getWidth()' on a null object reference
	at android.graphics.Bitmap.createScaledBitmap(Bitmap.java:798)
	at com.termux.terminal.TerminalBitmap.<init>(TerminalBitmap.java:93)
	at com.termux.terminal.TerminalBuffer.addImage(TerminalBuffer.java:593)
	at com.termux.terminal.TerminalEmulator.doOscSetTextParameters(TerminalEmulator.java:2307)
	at com.termux.terminal.TerminalEmulator.doOsc(TerminalEmulator.java:2046)
	at com.termux.terminal.TerminalEmulator.processCodePoint(TerminalEmulator.java:577)
	at com.termux.terminal.TerminalEmulator.processByte(TerminalEmulator.java:553)
	at com.termux.terminal.TerminalEmulator.append(TerminalEmulator.java:502)
	at com.termux.terminal.TerminalSession$MainThreadHandler.handleMessage(TerminalSession.java:345)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loopOnce(Looper.java:201)
	at android.os.Looper.loop(Looper.java:288)
	at android.app.ActivityThread.main(ActivityThread.java:7875)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)

```
##

MatanZ:
> * Chafa iTerm2 (`chafa --format iterm2`) crash report:

The crash is fixed. But the underlying issue is unsolved. Chafa sends a TIFF image, but BitmapFactory does not support this, so the image cannot be displayed. 

Since chafa sends an uncompressed TIFF, it should not be too hard to convert it to BMP. But I think it is better to say that Termux supports iterm2 protocol, but only for image formats supported by BitmapFactory.

MatanZ:
neofetch iterm2 should work now. neofetch includes linefeeds in its base64 data, while the single line example does not. This is the `-w 0` option. Now they are ignored.

rihaq:
> neofetch iterm2 should work now. neofetch includes linefeeds in its base64 data, while the single line example does not. This is the `-w 0` option. Now they are ignored.

Chafa?

leapofazzam123:
> > neofetch iterm2 should work now. neofetch includes linefeeds in its base64 data, while the single line example does not. This is the `-w 0` option. Now they are ignored.
> 
> Chafa?

https://github.com/termux/termux-app/pull/2973#issuecomment-1250072679

leapofazzam123:
Chafa now shows nothing instead of crashing the entire Termux application, we just need a way to display uncompressed TIFF image (or converting it to BMP for easy parsing) and Chafa will output something.

Meanwhile, neofetch display the image in the correct position and didn't get mysteriously vanished to the shadow realm when the information is fully displayed.

HardcodedCat:
Is this supposed to happen or is it a bug?

Screenshot_20220918-091214_Termux

When displaying images using Sixel or iTerm2, on neofetch, imgcat, etc
On the right side, some crosses always appear.
But these crosses doesn't appear when using chafa without sixel and iterm2

MatanZ:
> Is this supposed to happen or is it a bug?

It is a bug. should be fixed now.

rihaq:
apk please ( of latest commit)

HardcodedCat:
Nice!
And also, i've noticed that when zoomed in, the right side of the image gets cropped

# Zoomed in
Screenshot_20220918-111513_Termux

# Zoomed out
Screenshot_20220918-111539_Termux

Anonymous2716:
I get outofmemory crash error just for keept open a few text editor and if try to change the orientation of display , or change size.

HardcodedCat:
@Anonymous2716 what's your device specifications and OS version?
Could you please also provide a crash report?

Anonymous2716:
@HardcodedCat  

```
## Crash Details

**Crash Thread**: `Thread[main,5,main]`
**Crash Timestamp**: `2022-09-20 19:34:09.390 UTC`

**Crash Message**:
```
OutOfMemoryError thrown while trying to throw OutOfMemoryError; no stack trace available
```


### Stacktrace

```
java.lang.OutOfMemoryError: OutOfMemoryError thrown while trying to throw OutOfMemoryError; no stack trace available

```
##


## Termux App Info

**APP_NAME**: `Termux`
**PACKAGE_NAME**: `com.termux`
**VERSION_NAME**: `0.118.0`
**VERSION_CODE**: `118`
**TARGET_SDK**: `28`
**IS_DEBUGGABLE_BUILD**: `false`
**APK_RELEASE**: `F-Droid`
**SIGNING_CERTIFICATE_SHA256_DIGEST**: `228FB2CFE90831C1499EC3CCAF61E96E8E1CE70766B9474672CE427334D41C42`
##


## Device Info

### Software

**OS_VERSION**: `4.4.146`
**SDK_INT**: `28`
**RELEASE**: `9`


**IS_DEBUGGABLE**: `0`
**IS_TREBLE_ENABLED**: `true`
**TYPE**: `user`
**TAGS**: `release-keys`

### Hardware


**BOARD**: `MK008`
**HARDWARE**: `mt6757`

**SUPPORTED_ABIS**: `arm64-v8a, armeabi-v7a, armeabi`
##
```

4GB of RAM

rihaq:
Is anything need to change to make photon-rss working in termux? https://sr.ht/~ghost08/photon
Termux package - https://github.com/termux/termux-packages/blob/master/packages/photon-rss/build.sh

The developer says

```
Photon needs the terminal window resolution for scaling the images. It tries TIOCGWINSZ syscall and if it doesn't succeed then it tries to ask the terminal with CSI 14. When that also doesn't go well, then it just prints out this error message and quits.

The solution would be for termux to implement CSI14, or that photon would just ignore this error and play along without images.
```

Grimler91:
dmagnetic is another program(/game) we have that support sixels out of the box. To test, install dmagnetic, download a game with the `magnetic-scrolls` utility, and then run `dMagnetic -vmode sixel name-of-game`:

dmagnetic

Works well, no issues as far as I can see.

bropines:
Screenshot_2022-11-08-03-13-49-926_com termux

Either I'm stupid, or for some reason nothing works for me

TomJo2000:
Make sure you are running one of the latest GitHub CI builds, Sixel support isn't in the stable release yet.

rihaq:
@MatanZ Please commit new new features

XOXO045 (commented):
**terminal-view/src/main/java/com/termux/view/TerminalRenderer.java
**

AnonymouX47:
First of all, thank you all for your work on this! 🙏🏾 

> Small emulator changes:
> 
> - Also eat APC sequences, not echoing to screen.
> - Fix CSI 14 t to give actual size

Thanks for these... been working around these for a quite a while.

In addition, setting the `ws_xpixel` and `ws_ypixel` members of the `winsize` struct used by the `TIOCGWINSZ` ioctl will be very useful as this method is much faster and more reliable than the `CSI 14 t` query.

Most importantly, I have some concerns with the last point here (*from the OP*):

> Support iTerm2 inline image protocol (OSC 1337):
>
> - Using the same bitmap display infrastructure introduced for sixels.
> - Collects the image data outside of the OSC buffer.
> - Ignoring some parameters.

It would be greatly helpful to know what exact parameters are ignored (without having to figure it out by oneself).

I suggest that all unimplemented protocol features, protocol extensions and inconsistencies with other implementations (particularly the *reference* implementation by **iTerm2** itself) be clearly stated and documented for the sake of developers writing software to use this protocol.

One of the major headaches in developing cross-terminal software is inconsistency in implementation of protocols/specifications... and it's sincerely quite a pain.

Talking about inconsistencies... for instance,
- what happens to images when the screen is cleared?
- how do images interact with text and vice versa?

Note that I'm not saying the answers to questions like these should be documented. Instead, if the answers differ from those of the reference implementation, then those differences should be documented... **or better still, fixed**!

This also kinda applies to the sixel implementation (maybe see [this](https://github.com/hackerb9/vt340test) for reference/benchmarks/tests).

Once I'm able to get a working build of Termux with this branch and some time, I'll test as much as I can and give any necessary feedback.

Thanks once again 🙏🏾

MatanZ:
> In addition, setting the `ws_xpixel` and `ws_ypixel` members of the `winsize` struct used by the `TIOCGWINSZ` ioctl will be very useful as this method is much faster and more reliable than the `CSI 14 t` query.

This is in #3098.
 
> Most importantly, I have some concerns with the **last** point here (_from the OP_):
> 
> > Support iTerm2 inline image protocol (OSC 1337):
> > 
> > * Using the same bitmap display infrastructure introduced for sixels.
> > * Collects the image data outside of the OSC buffer.
> > * Ignoring some parameters.
> 
> It would be greatly helpful to know what exact parameters are ignored (without having to figure it out by oneself).

Checking the code, it seems that the only ignore parameter is `size`, which should be meaningless for well behaved programs.
 
>     * what happens to images when the screen is cleared?

The images are also cleared.
 
>     * how do images interact with text and vice versa?

Each cell may display either a part of an image or a character. Each cell displays the last thing (text or image) printed to it. I believe this is the standard behaviour for sixel and iterm2 protocols for all terminal emulators. kitty protocol has more advanced behaviour, and cannot really be supported with the current graphics implementation.
> 
> Note that I'm not saying the answers to questions like these should be documented. Instead, if the answers differ from those of the reference implementation, then those differences should be documented... **or better still, fixed**!
> 
> This also kinda applies to the sixel implementation (maybe see [this](https://github.com/hackerb9/vt340test) for reference/benchmarks/tests).

In general, I am more interested in getting current applications working than in bug to bug compatibility with 40 years old hardware. But even there I will sometimes choose a more reasonable behaviour.

For example - xterm limits sixel images to 1024 and screen width (the smaller of those). I limit to 8192 and not to screen width. This breaks lsix.

> 
> Once I'm able to get a working build of Termux with this branch and some time, I'll test as much as I can and give any necessary feedback.

If you want to run termux with this PR (and some other), without building yourself, there is https://github.com/HardcodedCat/termux-monet

AnonymouX47:
Oh, great!

> If you want to run termux with this PR (and some other), without building yourself, there is https://github.com/HardcodedCat/termux-monet

Will try that out. Thanks.

ikcikoR:
Is this being worked on? Can't wait for this feature but I'm seeing there has been no activity for over half a year

TomJo2000:
> Is this being worked on? Can't wait for this feature but I'm seeing there has been no activity for over half a year

The feature is finished as far as I am aware and pending inclusion in a future release.
If that's not soon enough you can grab a testing APK from a CI run.

Suletta-Majo:
i want. test apk or old sixel worked version ! what and called CI where is this?

Kreijstal:
idc if this get merged, where can I get an apk ??? o.o 
me want

Kreijstal:
> @Kreijstal: https://github.com/KitsunedFox/termux-monet

I uninstalled termux so fast I forgot to backup my configuration.
Doesn't matter sixel, was worth it.

AnonymouX47:
For what it's worth, I think the APK signatures are compatible. Over here, I can install one over the other (and vice versa) without loosing any data... though, I'm not sure how good an idea it is.

ikcikoR:
> For what it's worth, I think the APK signatures are compatible. Over here, I can install one over the other (and vice versa) without loosing any data... though, I'm not sure how good an idea it is.

Not if you got your Termux from f-droid sadly

TomJo2000:
Out of scope for this PR, please do not reply to dormant threads unnecessarily.
It notifies everybody involved in the thread.

SastroXXX (commented):
Update

Mainnet7m1 (commented):
Start

agnostic-apollo:
Hi @MatanZ, thanks for implementing sixels and iTerm images. Sorry, it took this long to look over this, didn't get time (not that I had time right now).

I have added my changes in 948ffa41 and 430ae28a primarily, but have kept your old ones in a7f2872c for now after rebasing against `master` and fixing conflicts. I will squash both sixel commits before merging pull, leaving it now to track old and new changes. While I have reviewed/fixed/improved changes for bitmaps/sixel/iterm, I have not fully looked at the rendering/scaling bitmap parts, but they seem to be working fine, will look into them in future when I have more time and issues arise.

Users should thoroughly test the terminal for both sixel and iterm images, and also other text editors, etc in case something got broken. I have done by own testing for the entire protocols, but not a variety of images (primarily tested attached interstellar image). Once testing has been done and @MatanZ reviews the changes, I will merge the pull and make a new Termux app version release for F-Droid users.

**You can grab APK build for current push from https://github.com/termux/termux-app/actions/runs/24187595506?pr=2973**

Commit message for 430ae28a is below.

---

Added: Add terminal support for Bitmaps, Sixel (`DCS q`) and iTerm Image (`OSC 1337`)

Support for displaying bitmaps inside the terminal has been added via `TerminalBitmap` which can be created from image `byte[]` or sixel bitmap. The bitmaps are sliced to character cell sized slices. The `TerminalBuffer` stores a map for bitmap number to the `TerminalBitmap` loaded in the terminal. The bitmap number and coordinates are encoded in the `long` `TerminalRow.mStyle` for the `TerminalRow` character of a column by `TerminalBitmap#buildOrThrow()` by getting encoded value from `TextStyle.encodeTerminalBitmap()`. The `TerminalRenderer.render()` then checks during rendering terminal output whether a character at a row/coloumn index is a bitmap instead of text by calling `TextStyle.isTerminalBitmap()`, then draws it using `Canvas.drawBitmap()` instead of `Canvas.drawText()`.

Sixel images can be created with Sixel Device Control String command sent via `DCS q s..s ST` or `DCS P1; P2; P3; q s..s ST`.

- The `TerminalEmulator` interprets sixel sequences, and sends them to `TerminalBuffer` for constructing a `TerminalSixel`. Once the sixel command has been completely processed, a `TerminalBitmap` is created from the `TerminalSixel`. If an error occurred during processing (like OOM), then remaining sixel command is completely read, but is ignored and no sixel is drawn (done by setting `mTerminalSixel` to `null` so that `TerminalBuffer.sixelReadData()` ignores further commands).
- Since a sixel sequence can be very long to render a full image and can have length greater than `TERMINAL_CONTROL_ARGS__MAX_LENGTH` (`16384`), the entire sequence is not stored in the `mTerminalControlArgs` buffer before being processed as it will result in an overflow error, instead as soon as length crosses `TERMINAL_CONTROL_ARGS__MAX_LENGTH / 2` and a complete sixel sub command (`#`, `!`, or `"`) has been received, it is immediately processed, and then further commands are read after emptying buffer.
- If "rough" horizontal and vertical size of image is received at start of sixel data string with a `Raster Attributes` command, like done by `img2sixel` command, then sixel commands args buffer capacity (`mTerminalControlArgs`) is increased and sixel bitmap in `TerminalSixel` is resized at start, instead of having to keep resizing buffer/bitmap as more sixel data is received, which has a performance hit due to memory reallocations and copying.
- The `4` (sixel) value has been added to `CSI` `Primary Device Attributes` terminal response.

The `img2sixel` command can be used to display sixel images after installing with `libsixel` package with `pkg install libsixel`, like with `img2sixel --width=1000px image.jpg`.

To manually send an escape sequence, check the `digiater.nl` link below, but it is too cumbersome to create images large enough to be easy viewable in the terminal.

See Also:
- https://vt100.net/docs/vt3xx-gp/chapter14.html
- https://en.wikipedia.org/wiki/Sixel
- https://www.digiater.nl/openvms/decus/vax90b1/krypton-nasa/all-about-sixels.text

iTerm images can be created with `1337` Operating System Control command.

- Both `File=` and `MultipartFile=` (chunk based) protocols are supported. The `inline` parameter should be `1` to display inline images in the terminal. Downloading images to Downloads folder with the value `0` will be ignored as that is not supported.
- The escape sequences/image data cannot be greater than `TERMINAL_CONTROL_ARGS__MAX_LENGTH` (`16384`) bytes if sent via (`File=`) protocol, otherwise it will be ignored with an overflow error. For larger images, send images via `MultipartFile=` protocol in chunks with `FilePart=`, the `imgcat` utility uses that with 200-byte chunks if `--legacy` flag is not passed.
- The `TerminalEmulator` interprets iTerm images sequences and creates an `ITermImage` to process parameters and store the base64 encoded image sent. Once all the data has been received, which can be over multiple `OSC` commands for `MultipartFile=` protocol, the encoded image is decoded to a `byte[]`, which is then passed to `TerminalBuffer`, which creates a `TerminalBitmap` for the image.

The `imgcat` utility can be used for sending images, like with `imgcat --width 1000px image.jpg` (`MultipartFile=`) or `imgcat --width 1000px --legacy image.jpg` (`File=`).

To manually send an escape sequence, run `echo -en '\e]1337;File=inline=1;keepAspectRatio=0;width=1000px;:' ; base64 -w 0 ./image.jpg ; echo -e '\e\\'` (`File=`).

See Also:
- https://iterm2.com/documentation-images.html
- https://iterm2.com/utilities/imgcat
- https://github.com/gnachman/iTerm2-shell-integration/blob/d1d4012068c3c6761d5676c28ed73e0e2df2b715/utilities/imgcat

---

### 284x177

interstellar

bitmap-tests

### 1920x1200

interstellar1

bitmap-large-tests

Kreijstal:
finally I can get termux again

robertkirkman:
If someone who already has Termux installed wants to test this without necessarily having to back up, erase and then later reinstall the regular Termux, they can use this `two.termux` with this PR applied that I built.

- [sixel `two.termux`](https://github.com/robertkirkman/termux-generator/actions/runs/24218531388)

I also built a `com.termux` one with similar settings, but have not linked it because agnostic-apollo already linked one above so if you don't already have a `com.termux` installed you should use their link.

In order to avoid speculation from anyone who notices that the build I linked also currently contains this other PR,

- https://github.com/termux/termux-app/pull/4961

I would like to acknowledge that and also emphasize that **the termux-generator links I send are _not_ for replacing or competing with the official Termux as a fork with more PRs merged**, so that detail should not be interpreted that way. I send them because I want PRs that "**request testers**" to be accessible to anyone who only has 1 device and for whom uninstalling and reinstalling the regular Termux constantly while testing PRs is frustrating or inconvenient.

robertkirkman:
I have tested this in comparison to the `konsole` package running in Termux:X11, and for the most part it looks great,

Screenshot_20260410-021158_Termux_X11


However, I do seem to have found a sixel image test case that `konsole` currently seems to be able to handle better than this PR.

When I use `img2sixel arrow.jpg` on this large-resolution `arrow.jpg` I found, this PR doesn't seem to be able to display any portion of it, while `konsole` displays the portion that it's able to fit into its window.

Screenshot_20260410-021502_Termux_X11

arrow


however, if I reduce the resolution of `arrow.jpg` to something more reasonable using an image editor, I'm able to display it in both `konsole` and this PR:

Screenshot_20260410-022526_two termux


arrow_small

That seems somewhat reasonable; there has to be a size limit at some point, and maybe the size limit of `konsole` is just greater than that of this PR.

agnostic-apollo:
Likely this is engaging.

https://github.com/termux/termux-app/pull/2973/changes#diff-f729a9bbec25dbca63392265f8852d3ec8d4f0afa848fccf3b2db837c457ab98R29

robertkirkman:
I see, I think 2048 is a reasonable limit for sixel.

I've tested `imgcat` (iterm2) now as well, and interestingly, I've noticed that `imgcat` is able to display the same image at full resolution with no problems,

Screenshot_20260410-035115_two termux

however, `imgcat` is also faster. It takes much less time to display the large image than `img2sixel`, even when `img2sixel` is limited to a working size with `--width=2000`.

If iterm2 is just more efficient than sixel, it's probably best for the size limit of sixel to be smaller than the size limit of iterm2 (as it currently is)

robertkirkman (approved):

agnostic-apollo:
There are some discussions at following links about size limits, I choose 2048x2048 (which uses 16MB memory) as max sixel bitmap dimensions as android devices are generally more stricter with RAM allocation for apps than linux distros and will trigger an OOM, especially with all the other data and processes running in termux. Only issue seems to be that sixel command should still have been read completely without being printed to screen, and no image displayed, just like iterm.

https://github.com/contour-terminal/contour/discussions/680
https://github.com/contour-terminal/contour/issues/656#issuecomment-1118390610

agnostic-apollo:
iterm (or possibly imgcat) is around 20 times faster than sixel, I did calculations and rendering same image with iterm was around 7-10ms on average, compared to around 200ms or more for sixel.

robertkirkman:
> Only issue seems to be that sixel command should still have been read completely without being printed to screen, and no image displayed, just like iterm.

It displays no output for a small range of sizes around 2049-2100, then after that it starts displaying this large text artifact in the terminal, but shows similar messages in `logcat` both cases:

image

agnostic-apollo:
Can you post logcat errors for that?

agnostic-apollo:
> The sixel repeat command Pn value 4500 is greater than max repeat value 2048 at index 3410 of sixel input.

Ah, a repeat value of 4500 is considered as an invalid sixel sequence and further input is not read, and that gets printed to screen. Will have to think on whether to continue reading or keep as is, probably former as 4500 is a valid integer sent by client, we just dont accept it. Other sixel command limits would need to be looked too.

MatanZ:
> Hi @MatanZ, thanks for implementing sixels and iTerm images. Sorry, it took this long to look over this, didn't get time (not that I had time right now).

> iTerm images can be created with `1337` Operating System Control command.
> 
>     * Both `File=` and `MultipartFile=` (chunk based) protocols are supported. The `inline` parameter should be `1` to display inline images in the terminal. Downloading images to Downloads folder with the value `0` will be ignored as that is not supported.
> 
>     * The escape sequences/image data cannot be greater than `TERMINAL_CONTROL_ARGS__MAX_LENGTH` (`16384`) bytes if sent via (`File=`) protocol, otherwise it will be ignored with an overflow error. For larger images, send images via `MultipartFile=` protocol in chunks with `FilePart=`, the `imgcat` utility uses that with 200-byte chunks if `--legacy` flag is not passed.

I understand supporting `MultipartFile=`. It was not in iterm2 when I wrote this patch. But intentionaly crippling `File=` seems to me pointless. There are a lot of older tools that only support 'File=' method. They work with my patch, but not with yours. Furthermore, limiting chunks to 16K, when iterm2 uses 1MB, with no way to query the terminal for this value, mean that programs need to test terminal type and have a database of terminal capabilities, which is not a reasonable approach.

I suggest removing the arbitrary 16K limit for both `File=` and `MultipartFile=` methods.

MatanZ:
> There are some discussions at following links about size limits, I choose 2048x2048 (which uses 16MB memory) as max sixel bitmap dimensions as android devices are generally more stricter with RAM allocation for apps than linux distros and will trigger an OOM, especially with all the other data and processes running in termux. Only issue seems to be that sixel command should still have been read completely without being printed to screen, and no image displayed, just like iterm.
> 
> [contour-terminal/contour#680](https://github.com/contour-terminal/contour/discussions/680) [contour-terminal/contour#656 (comment)](https://github.com/contour-terminal/contour/issues/656#issuecomment-1118390610)

Agan, I prefer to avoid unnecessary arbitrary limits. Termux does run on screens wider than 2048 pixels, and programs might want to use all that. Images need not be square, so one might want to draw a 2300x50 header for the page.

If you want to reduce the chance of OOM, maybe limit actual area, instead of each dimension separately.

Not that there are no problems with supporting large sixels. [lsix](https://github.com/hackerb9/lsix) for example, uses maximum sixel size rather than actual screen width to decide its display width.

robertkirkman:
> I prefer to avoid unnecessary arbitrary limits. Termux does run on screens wider than 2048 pixels, and programs might want to use all that.

Do you happen to know whether Konsole and other terminal emulators are really fully unlimited, or do they just have a size limit that's really big?

MatanZ:
There actually need to be limits, for reporting in  XTSMGRAPHICS escape sequence. 

In foot it is 10000 pixels. In konsole, 16384.

robertkirkman:
Maybe if the limit were set to something around 8192 in Termux, that would be an acceptable compromise? That is somewhat in between the limit agnostic-apollo wanted, and the limits of other terminals. Estimating based on Agnostic-apollo's previous calculation, I think that would put the maximum memory use somewhere in between 100 and 200 MB, which isn't actually much even on Android. (assuming only attempting to display one large image at a time)

agnostic-apollo:
@MatanZ and @robertkirkman agreed that the limits were low, I have increased them with 5743c5de. I looked into what limits Android itself was using for different components, and I have increased them to support `7680x4320 (8K UHD)` images depending on Android version and device, more info in `TerminalBitmap.MAX_BITMAP_SIZE` variable docs at https://github.com/MatanZ/termux-app/blob/5743c5dee233ddec4e1dc6fc1ebaa752e7105b49/terminal-emulator/src/main/java/com/termux/terminal/TerminalBitmap.java#L42-L123. If an OOM happens, image rendering will just be ignored too anyways.

> a repeat value of 4500 is considered as an invalid sixel sequence and further input is not read, and that gets printed to screen.

This has been fixed too. Repeat value cannot be greater than 8192 now, or rendering will be ignored.

**You can grab APK build for current push from https://github.com/termux/termux-app/actions/runs/24365026780?pr=2973**

robertkirkman:
- [`two.termux` build of the current version of this PR](https://github.com/robertkirkman/termux-generator/actions/runs/24372243816)

robertkirkman:
> > a repeat value of 4500 is considered as an invalid sixel sequence and further input is not read, and that gets printed to screen.

> This has been fixed too. Repeat value cannot be greater than 8192 now, or rendering will be ignored.

I have tested this with Samsung Galaxy S8+ SM-G955F, and I can confirm that it is now able to display the large (4500x4500) arrow image! It takes a few seconds to load, but I know that the delay is caused by inefficiency inherent to the sixel protocol, since Konsole also takes a long time to display the same sixel on the same device.

Just to stress test this, I have also tested this on Samsung Galaxy S III SPH-L710 (released 2012, 2GB RAM). Here are results of that test:

It has no problem displaying the small image, and loads it in under 1 second.

Screenshot_20260414-011007

However, attempting to display the large arrow image (4500x4500) results in the activity freezing for about 7 minutes,

Screenshot_20260414-005755

after that screen, it progresses if I just wait, but after the command completes, it creates a large space where the image should be displayed, without displaying the image itself:

Screenshot_20260414-010717

Screenshot_20260414-010751

`logcat` shows this:

Screenshot_20260414-010835

I think this is reasonable to expect. Since sixel is shown by preexisting implementation Konsole to be intensive and inefficient with large resolution images by nature, very old devices should not be expected to be capable of displaying very large sixel images.

agnostic-apollo:
Thanks for testing robert. Yeah, loading large images on older devices is not recommended and is bound to cause issues. Termux isn't responding errors are expected too since bitmaps are created and rendered on main UI thread, high end devices could have issues too with very large images if it takes too long.

j4james:
If you really wanted to, you could theoretically support large images, even on an older device, but you'd need to implement sixel in the same way as the original hardware terminals. These were devices from the 1980s, with a _tiny_ amount of memory, but they could handle infinitely large images (albeit slowly). It's just that they treated sixel as a streaming protocol, rather than an image protocol, so the content was displayed as it was received, rather than trying to buffer the whole thing into memory, and only rendering it at the end.

When you receive a row of sixels that is 4500 pixels wide, and your display is only 800 pixels wide, you just render the first 800 columns and ignore the rest - you don't buffer all 4500 columns, most of which will never be seen. And as new sixel rows are added to the image at the bottom of the screen, you're going to have older parts of the image scrolling off the top. So you don't _have_ to retain those older rows if you're low on memory, in the same way that you don't necessarily keep every row of text in the scrollback.

Not suggesting you have to implement it this way though. Just something to consider if you're keen to improve support for low-memory devices.

agnostic-apollo:
> And as new sixel rows are added to the image at the bottom of the screen, you're going to have older parts of the image scrolling off the top. So you don't have to retain those older rows if you're low on memory, in the same way that you don't necessarily keep every row of text in the scrollback.

But modern terminals including termux have a scrollback history. If entire image is not stored, you cannot scroll back to older rows. Termux supports upto 50000 rows, and defaults to 2000 rows, which is large enough for many many sequential images, a 1920 height image uses only 45lines on a 2856 height display. Older devices also dont have a display resolution high enough that it would matter having to support 4K+ images. Users should just remain within their means.

j4james:
> But modern terminals including termux have a scrollback history. If entire image is not stored, you cannot scroll back to older rows.

I'm not saying you should _never_ store any of the image in the scrollback - just that you can choose to drop parts of an image that have scrolled out of view if you're low on memory. Having part of a large image missing from the scrollback is surely better than not seeing any image at all?

But feel free to ignore me. It was just an idea.

agnostic-apollo:
Its likely more easily doable to free bitmaps that are at start of scrollback history if an OOM occurs when reading a new bitmap to be added to end of scrollback history, instead of freeing parts of an image. Bitmaps that have been completely scrolled out of scrollback history are already removed, note that scrollback history is larger than the current terminal view of the screen.

Your idea is actually good, especially for older terminals, but is more likely to complicate things for termux's case. And your optimization wouldn't work for iTerm images anyways as entire base64 encoded image would need to be read before its converted to a bitmap. If sixels is an issue in future, one could create a new bitmap for each row of a sixel input, and bitmap lines scrolled out could be freed, and older bitmap lines at start of scrollback history could be freed too if an OOM occurs.

robertkirkman:
@MatanZ I'm sorry for bothering you here again, I just wanted to let you know that my impression of the status is that this PR will not be merged unless you express assent to the current compromise-based version, or alternatively object and request more changes which we could then attempt to implement, if you already understand this and are still reviewing the changes then please disregard my message.

agnostic-apollo:
Yeah, waiting for @MatanZ response to changes and approval.

MatanZ:
For me, not having arbitrary size `File=` in iterm2 protocol makes this (much) less useful than my branch. As such, I did not compile or tried to run this code.

If this is merged into termux I might later try to (re-)add support for this feature atop agnostic-apollo's work. Until then, I am happy with my branch.

agnostic-apollo:
> For me, not having arbitrary size File= in iterm2 protocol makes this (much) less useful than my branch.

Are you planning to display more than 8k resolution images? Even android won't render them anyways considering internal texture size limits, which my PR automatically adjusts too anyways if a device manufacturer has defined `ro.hwui.max_texture_allocation_size` system property. My limits are consistent with device manufacturer and Android limits.

MatanZ:
Sorry for not being clear enough. This is not about image dimensions, but about the limit of 16K characters for the escape sequence (which was not in my code).

All the tools I am using employ the `File=` method with arbitrary size. The other terminal I mostly use, konsole, only supports `File=` (though there is a merge request for `MultipartFile=`). There is also no simple way for an application to tell which of those two are supported.

agnostic-apollo:
Limit has already been increased from 16K to 100-150MB or more.

https://github.com/MatanZ/termux-app/blob/5743c5dee233ddec4e1dc6fc1ebaa752e7105b49/terminal-emulator/src/main/java/com/termux/terminal/TerminalEmulator.java#L2700

robertkirkman:
@MatanZ would you happen to have any examples we could test of iTerm2 images that don't work with the current version in this PR? That way we could test them and find a way to change the code to work with them.

Kabouik:
> @MatanZ and @robertkirkman agreed that the limits were low, I have increased them with [5743c5d](https://github.com/termux/termux-app/commit/5743c5dee233ddec4e1dc6fc1ebaa752e7105b49). I looked into what limits Android itself was using for different components, and I have increased them to support `7680x4320 (8K UHD)` images depending on Android version and device, more info in `TerminalBitmap.MAX_BITMAP_SIZE` variable docs at https://github.com/MatanZ/termux-app/blob/5743c5dee233ddec4e1dc6fc1ebaa752e7105b49/terminal-emulator/src/main/java/com/termux/terminal/TerminalBitmap.java#L42-L123. If an OOM happens, image rendering will just be ignored too anyways.
> 
> > a repeat value of 4500 is considered as an invalid sixel sequence and further input is not read, and that gets printed to screen.
> 
> This has been fixed too. Repeat value cannot be greater than 8192 now, or rendering will be ignored.
> 
> **You can grab APK build for current push from https://github.com/termux/termux-app/actions/runs/24365026780?pr=2973**

I have installed the build from the artifacts you posted, then the extra apks from robertkirkman's post below, but I still can't get `chafa` working. I ran `pkg update` and `pkg upgrade`, installed `chafa libsixel`, but no images. 

`chafa -f sixels path/to/screenshot.png` just shows a bunch of characters (more than the terminal buffer can remember). `chafa -f iterm path/to/screenshot.png` shows nothing. `./imgcat image.png` using the script posted somewhere above as well as `img24ixel image.png` work just fine. 

Any idea what may be wrong with `chafa`? Someone reported here it was working.

agnostic-apollo:
Thanks for the report @Kabouik.

For `chafa -f sixels`: The sixel DCS command `P3` parameter must end with a `;` and then be followed by `q` as per spec. Chafa does not send the semicolon, so it wasn't recognized as a valid sixel command. Have fixed it with 506183f18f56fac93b6c6b7f98db502ace1934ae since other clients might not be sending it either.

https://vt100.net/docs/vt3xx-gp/chapter14.html#S1S14.2.1

For `chafa -f iterm`: Android throws a `Failed to create image decoder with message 'invalid input'` error in logcat when trying to create an image from the decoded base64 image data sent by `chafa` with [`BitmapFactory.decodeByteArray()`](https://developer.android.com/reference/android/graphics/BitmapFactory). That's likely because android does not support TIFF format used by `chafa`. To add support for that would require some external library to decode it, which is out of scope at this time.

- https://github.com/hpjansson/chafa/blob/ea18ac4fa094213e92ec31b687efc8054d2876b6/chafa/internal/chafa-iterm2-renderer.c#L276
- https://developer.android.com/media/platform/supported-formats#image-formats

**You can grab APK build for current push from https://github.com/termux/termux-app/actions/runs/25403833943?pr=2973**

j4james:
> The sixel DCS command `P3` parameter must end with a `;` and then be followed by `q` as per spec.

@agnostic-apollo The example in the VT330/VT340 manual, while technically valid, is not an indication of the required format for Sixel sequences. And I suspect it's probably a typo, because the VT340 itself doesn't produce Sixel output like that when generating a hardcopy.

According to the DEC STD 070 reference (§ 3.5.4.1 and 3.5.3.1), DCS parameters should be parsed in the same way as CSI parameters. A semicolon is a parameter separator, not a parameter terminator; zero-length parameter value represents a default value; and default parameter values can be omitted from the end of the parameter list.

So something like `\eP1;2;3;q` is equivalent to `\eP1;2;3;0q`, which is also the same as `\eP1;2;3q`. And note that `\ePq` is also a perfectly valid sequence introducer, and would be the equivalent of `\eP0q`, `\eP0;0q`, or `\eP0;0;0q`.

robertkirkman:
- [`two.termux` build of the current version](https://github.com/robertkirkman/termux-generator/actions/runs/25404946548)

agnostic-apollo:
@j4james Well, not sure if its a typo, even wikipedia says "Sixel mode is entered by sending the sequence `ESC Pp1;p2;p3;q`". Hard to know if real world applications were actually following it. I do know that the last parameter normally doesn't end with a semicolon, even for sixel control function commands, but I am sure we can agree that parsing is easier if every parameter were to end with a semicolon. The sixel DCS command start sequence itself is also a stupid design, the optional parameters should have come after the `q` instead of them coming before and parsers having to look ahead and then going back, the other DCS commands start with a `$q` or `+q`, then follow other data, could be some legacy design port issue.

https://en.wikipedia.org/wiki/Sixel

> would be the equivalent of `\eP0q`, `\eP0;0q`, or `\eP0;0;0q`.

I wasn't allowing missing parameters before, but have fixed it with 2151d08496b184980bdf89351213c7617a54e0d3

**You can grab APK build for current push from https://github.com/termux/termux-app/actions/runs/25445369130?pr=2973**

Kabouik:
Awesome, thanks @agnostic-apollo, I can confirm `chafa` and my TUI streaming app now show covers correctly (though with some flicker and slowness that I don't observe on desktop Linux):

Screenshot_20260506-201543

agnostic-apollo:
Welcome. An android device or at least termux is likely going to be slower than desktop, but check if resolution of video being passed is not higher than display size, otherwise share the commands and video you are trying to view and I can try to test. Is video streaming app using sixel or iterm?

Kabouik:
> Welcome. An android device or at least termux is likely going to be slower than desktop, but check if resolution of video being passed is not higher than display size, otherwise share the commands and video you are trying to view and I can try to test. Is video streaming app using sixel or iterm?

It's just music album covers which are all 640×640. However, I don't have time to dive into it right now, but I believe there are many redraws occurring when the TUI is navigated, and that may exacerbate any slowness compared to desktop. It should be using `chafa -f sixels`, but I need to check if that logic is actually respected because the flickering I'm observing should not be there if that was really the case, since I don't observe it on PC with the same dependencies.

agnostic-apollo:
Okay, let me know if you figure it out. iterm is 20x faster than sixel though.

I'll wait a couple of days more for any more tests by users, before I merge and make a release.

robertkirkman:
- [`two.termux` build of the current version]( https://github.com/robertkirkman/termux-generator/actions/runs/25454187335)

Kabouik:
> Okay, let me know if you figure it out. iterm is 20x faster than sixel though.
> 
> I'll wait a couple of days more for any more tests by users, before I merge and make a release.

An official Termux release with sixel support? Awesome. 

However I just tested `chafa -f iterm /path/to/jpg` and it shows nothing at all. You suggested using iterm because it's faster, so does it work on your end?

robertkirkman:
I can confirm that,

for me, `chafa -f sixel` (chafa sixel), `img2sixel` (1st-party-ish sixel tool), and `imgcat` (1st-party-ish iterm2 tool) all work, but `chafa -f iterm` does not seem to work.

`imgcat` comes from:

```bash
curl https://iterm2.com/utilities/imgcat > $PREFIX/bin/imgcat
chmod +x $PREFIX/bin/imgcat
```

Screenshot_20260507-113857_Termux

agnostic-apollo:
As I already said chafa will not work with iterm as it sends the image in tiff format regardless of source image format, and tiff is not supported by android platform libraries for decoding. To support tiff in termux end would require adding libtiff or some other external native/java library in the app, adding native libraries is complex for an Android app and I avoid adding dependencies unless necessary. And it may not be needed anyways as chafa maintainer says he plans to shift to PNG images, which will automatically work in termux when chafa shifts, you can track that in https://github.com/hpjansson/chafa/issues/333.

agnostic-apollo:
Instructions are also available at end of https://github.com/termux/termux-app/pull/2973#issuecomment-4213795809 for how to use iterm protocol with just simple echo/base64 commands without any other external tool, like imgcat/chafa.

MatanZ:
chafa uses uncompressed tiff. It would be quite easy to decode without any external library.

agnostic-apollo:
I haven't looked at how tiff format works, and how chafa is using it. And not sure whether a simpler decoding will work for other clients.

I also don't have time to work or even research into it currently, I'm busy with termux rewrite, if you want work on it yourself right now, let me know and I'll hold off merging the pull. Otherwise can wait for chafa to switch to png or add basic tiff support in future and merge for now.

MatanZ:
I believe it is better to merge (and release) as soon as possible. chafa works in sixel mode, and other tools work also in iterm2 mode.

It is always possible to add this later if desired.

robertkirkman:
The link and explanation above suggests that a future version of chafa plans to switch to png for iterm2, which would solve the problem without termux intervention, I guess? in my opinion that makes it not a high priority unless many other iterm2 apps also require tiff support.

agnostic-apollo:
Ok, merge it is then.

WALElionel:
Wifizone je suis très intéressé

robertkirkman:
On April 17th the status was that approval of the current version was required from MatanZ, and on May 7th MatanZ gave approval of this version.

In between those dates, on May 6th, an ETA of 2 days was given, and if the merging preparation work is still ongoing and more time is needed then that is fine, I just thought that I should mention that time passed since an ETA was given.

agnostic-apollo:
Sorry, pull is ready to be merged, other than squashing commits and updating commit messages. I have been busy with finishing some of the termux-app rewrite stuff and couldn't shift attention from it to do a release as I have a limited context window size. Will get to it soon.

robertkirkman:
Thanks for the update, based on the context I had just assumed that you were prepared to and wanted to merge this PR before finishing the major rewrite project, but if it would be easier and less stressful for you to postpone it until afterward I wouldn't mind that.

agnostic-apollo:
I didn't mean the entire rewrite, I meant a part of it, have had to think on some complex issues and couldn't be distracted.

tmo793:
> **You can grab APK build for current push from https://github.com/termux/termux-app/actions/runs/25445369130?pr=2973**

The artifacts at that link are "expired" now, unable to download. Could I pls request a rebuild/refresh?

Same issue here:
> - [`two.termux` build of the current version]( https://github.com/robertkirkman/termux-generator/actions/runs/25454187335)

I note there are newer two.termux builds, not sure if they include the sixel PRs. I couldn't get the newer ones to work on my phone (bootstrap fail - file not found something-something-aarch64.xz or similar)

I'm new to GitHub actions/workflows, I did my best to find an APK build including this PR but couldn't figure it out - feel free to lecture me on the basics if I've missed something obvious

agnostic-apollo:
Thanks a lot @MatanZ for your contribution, and everyone who helped to test, especially @robertkirkman.

I was supposed to make a non-critical change about increasing some limit, but I have forgotten what that is and idiot me don't seem to have added a note for it anywhere, will fix it later if my brain remembers.

@tmo793 You can grab the `master` branch build from https://github.com/termux/termux-app/actions/runs/34651962568

PR #3043, opened 2022-10-12
Add LastPendingIntentRequestCode support to TermuxAPIAppSharedPreferences

EduardDurech:
For use with https://github.com/termux/termux-api/pull/556 and https://github.com/termux/termux-api-package/pull/161

See https://github.com/termux/termux-api/issues/550 and https://github.com/termux/termux-api/issues/246


agnostic-apollo:
Enabled `commitToFile`. Also ideally pull requests shouldn't be sent from the `master` branch, create separate branche for each pull, useful if multiple pull requests need to be sent and you won't have to force push `master` branch on your repo after merging and can just delete pull branch.

agnostic-apollo:
Also ideally constant should have been added to `TermuxPreferenceConstants` javadoc changelog due to some old convention I keep on using. I didn't do it currently cause it would have broken multiple commits on my local repo.

EduardDurech:
> Also ideally pull requests shouldn't be sent from the `master` branch, create separate branche for each pull, useful if multiple pull requests need to be sent and you won't have to force push `master` branch on your repo after merging and can just delete pull branch.

Okay, will keep in mind, thanks!

PR #3083, opened 2022-11-08
Reformat code

ghost:


agnostic-apollo:
Hard word wrapping and needlessly splitting on newlines is not needed and just decreases readability.

PR #3138, opened 2022-12-09
Added monochrome tag to ic-launcher(_round).xml to support dynamic icons

tiredElephant:
I just added the monochrome tag to change the icon with material you. I just used the normal forground because the icon is monochrome already.


tiredElephant:
Just saw it had already been done, sorry.

PR #3184, opened 2023-01-12
Add documentation from wiki.termux.com

Neo-Oli:
I did some weird magic using php and pandoc and converted most Wiki pages to Markdown. It seems to work alright with termux.github.io. While some of the info is extremely outdated and some conversion errors may still remain, I think this is a good base to build upon for a new documentaton.


agnostic-apollo:
Thanks a lot for converting to markdown though! This will be really helpful.

However, entire wiki can't be merged in termux-app, every termux repo will have its own docs directory and termux.github.io will combine its own main docs and all the others like its currently doing for termux-app.

Secondly, a flat directory structure is not going to be used. There should be multi level sub directories for each type of doc, which will also be reflected in url path. So will have to think on right directory hierarchy.

So there is still lot of work to be done before wiki can be split and merged even if content itself is fixed. Wiki only contains few pages related to termux-app docs, which I have mostly replaced locally anyways. A lot belongs to termux-packages docs and the rest for main termux.github.io docs, like index and community pages, etc.

Will respond about other stuff if any later.

Thanks again! 🎉

Neo-Oli:
@agnostic-apollo I see. Perhaps a rough outline should be created first in the individual repositories. Folder Structure and example pages mostly. Then we can start to sort in content from this PR into the specific repositories. A lot of content will obviously have to be discarded completely as it is outdated irrelevant or of poor quality. This would also somewhat outline, what we actually want to document, ie. complete documentation of all available packages or just Termux-Specific special stuff.

agnostic-apollo:
>Perhaps a rough outline should be created first in the individual repositories.

Yeah, that would be great! But until the next release cycle happens, I can't pick any more work. We can do this afterwards, unless someone else wants to take initiative and write the packages docs, in which case I guess I will try taking time out to guide.

> A lot of content will obviously have to be discarded completely as it is outdated irrelevant or of poor quality.

Yeah, agreed. Lot of stuff needs to be completely rewritten since current wiki doesn't meet the standards for how docs should be for many things.

Moreover, that are also certain rules that must be followed while writing the docs related to markdown formatting, so that the docs work both on the site and as well as in-app markdown viewer I was writing a while back. They both use different libraries and they both have differences from markdown shown on github, so there are compatibility issues that must be taken care of. I fixed some of them related to headings for the site locally and have decided on some rules for other stuff, will document them later.

Moreover, the docs must not contain any hardcoded `com.termux` or even `/data/data/com.termux/...` paths. Docs will have placeholders which will be dynamically replaced when site gets built or by the in-app  viewer. Have added support for that as well in the site, but need to add some stuff.

The directory hierarchy will also require using relative paths when linking pages. Asset files would also have to be in sub directories themselves.

Basically, lot of things needs to be worked on and will take time.

lukestanley:
@agnostic-apollo 
Hello, what about an imperfect solution for now that's less work? 
"Long time listener, first time caller here."
After looking at source code, I was hoping to document the awesome termux-notification-list command (which has no page) so I made an account on the wiki, and found I couldn't add a page yet, I thought, no worries, I'll see where I can request access to that or somewhere to put article content, like a related command page, nope, can't do that I have to edit a talk page to request access first, okay, so I tried to do that, then found out that the talk page itself was locked!
As a wiki maintainer in a previous life, I totally get the problems of spam and access.
An imperfect solution for now, using a GitHub repo for content that was on the wiki is probably a good solution with less friction since PR's can be done. Wiki means quick after all. Content could be moved or copied to individual repos later?
I don't want to tread on anyone's toes of this wonderful project, but surely there is a more scrappy but better solution than the current broken wiki situation that has less burden? :)

agnostic-apollo:
We were granting access to specific people who were requesting it but it was pulled for every non-termux dev a few days ago.

https://matrix.to/#/!rrHhKvLYtBVDOISphu:gitter.im/$JnxmDjWV6MAjRS7yQR2vFzW6d0M5ZAZGkcBVWlAZJnY?via=gitter.im&via=matrix.org&via=aria-net.org

https://matrix.to/#/!rrHhKvLYtBVDOISphu:gitter.im/$Gv1_eWg6JvBKZFdJC186bDCOaS78veYdGFNoACskQU0?via=gitter.im&via=matrix.org&via=aria-net.org

You can submit a modified wiki page in a termux-api repo issue and the termux devs can add it themselves in the wiki. That should work for now.

PR #3241, opened 2023-02-08
ci(attach_debug_apks_to_release): delete release on failure to build/upload

thunder-coding:
As discussed in https://app.element.io/#/room/#termux_dev:gitter.im

PR #3253, opened 2023-02-16
apply changes to attach_debug_apks_to_release as requested

thunder-coding:
See #3241

Once this PR is approved, a new release of upload-release-actions will be created, and the last commit will be reworded.

Test release:
https://github.com/thunder-coding/termux-app/releases/tag/v0.118.6

### DIFF
--- a/.github/workflows/attach_debug_apks_to_release.yml
+++ b/.github/workflows/attach_debug_apks_to_release.yml
@@ -39,7 +39,10 @@ jobs:
           echo "Building APKs for 'APK_VERSION_TAG' release"
           export TERMUX_APK_VERSION_TAG="$APK_VERSION_TAG" # Used by app/build.gradle
           export TERMUX_PACKAGE_VARIANT="${{ env.PACKAGE_VARIANT }}" # Used by app/build.gradle
-          ./gradlew assembleDebug
+          if ! ./gradlew assembleDebug; then
+            echo "Build failed for '$APK_VERSION_TAG' release."
+            exit 1
+          fi
 
           echo "Validating APKs"
           for abi in universal arm64-v8a armeabi-v7a x86_64 x86; do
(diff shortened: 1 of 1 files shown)


agnostic-apollo:
Thanks for the changes.

However, read the specs again linked via https://github.com/termux/termux-app/pull/3241#discussion_r1106019790

There must be a space followed by a space for text mode or `*` for binary mode between hash and filename. The `sha265sum` and `md5sum` defaults to text mode, so there are two spaces. You are adding a tab instead for some reason.

https://www.gnu.org/software/coreutils/manual/html_node/md5sum-invocation.html

https://www.gnu.org/software/coreutils/manual/html_node/sha2-utilities#sha2-utilities

https://github.com/termux/upload-release-action/commit/5f5653d0d1e24a1278a29b96a2029eab12a2f51f

I don't know what mode node crypto internally uses compared with gnu utils, which would have to be checked too. Apparently, in text mode, `\r\n` may get translated to `\n`.

https://unix.stackexchange.com/questions/127959/md5sum-command-binary-and-text-mode/127961#127961

https://stackoverflow.com/questions/17988090/what-are-the-differences-between-md5-binary-mode-and-text-mode

https://stackoverflow.com/questions/70360822/python-binary-mode-vs-encoding-and-its-sha256

Hopefully for following, specially with globs, there are duplicate file name checks since different (sub)directories could have files with same basenames.

https://github.com/termux/upload-release-action/commit/422a9a2f161f0e86bd33f8739d9dfe612bda3ae0

agnostic-apollo:
Also, follow commit guidelines.

https://github.com/termux/termux-app#For-Maintainers-and-Contributors

Also, despite the fixes to `upload-release-action` for other termux repos, I still haven't decided if adding an external action dependency without any advantages should be done, specially something [written in frickin' node/javascript](https://www.reddit.com/r/ProgrammerHumor/comments/114se66/104_vulnerabilities_91_moderate_9_high_4_critical/j8xrt60) ;)

Maybe other termux devs can give their opinion. Technically, termux-packages uses it for packages and bootstrap, so any effects would get translated.

PR #3285, opened 2023-03-07
New setting: draw-bold-text-with-bright-colors

markuspeloquin:
The new setting is draw-bold-text-with-bright-colors, and it is disabled by default, as it always should have been. This is also the name Alacritty uses. Similarly, it is disabled by default.

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxService.java
+++ b/app/src/main/java/com/termux/app/TermuxService.java
@@ -605,6 +605,10 @@ public final class TermuxService extends Service implements AppShell.AppShellCli
             return null;
         }
 
+        newTermuxSession.getTerminalSession().setBoldWithBright(
+            mProperties.shouldDrawBoldTextWithBrightColors()
+        );
+
         mShellManager.mTermuxSessions.add(newTermuxSession);
 
         // Remove the execution command from the pending plugin execution commands list since it has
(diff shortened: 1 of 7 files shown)


markuspeloquin:
Disclaimer: I do not know how to get this on my Android. I was merely able to build it cleanly after updating `minSdkVersion` to 24. I tried `./gradlew assemble` on the APK but it wouldn't install...

agnostic-apollo:
I'll have to read up on this.

>It deviates from existent behaviour.

Existent behavior for termux? New property already defaults to false and will change termux behavior.

https://github.com/termux/termux-app/blob/b1b36a36b23c66fb3c03a1e7538429078d9f2863/termux-shared/src/main/java/com/termux/shared/termux/settings/properties/TermuxPropertyConstants.java#L442-L458

thunder-coding:
> Existent behavior for termux? New property already defaults to false and will change termux behavior.

Sorry, my bad. I assumed that Termux didn't force bold text to print with bright colors. Had never ran over this issue myself.

I do believe changing the default behaviour would be nice thing to do. Would like to know your reasoning behind keeping the current behaviour.

agnostic-apollo:
I have no reasoning currently and will need to read up on its history. I haven't noticed the issue myself either.

markuspeloquin:
Here, you can see how Linux treats bright vs bold. If the color is bright, it converts it to non-bright and switches on 'bold':
https://github.com/torvalds/linux/blob/e8d018dd0257f744ca50a729e3d042cf2ec9da65/drivers/tty/vt/vt.c#L1761-L1765

But 'bold', 'italic', and 'underline' are merely implemented as colors. Here, you can see that when bold is enabled, the color (b0BBB0FFF for 3-bit background and 3-bit foreground) gets 8 added to the foreground:

luisalb667:
Hello Can you tell me how I can find commands for Linux and termux to have access to the security camera

PR #3336, opened 2023-04-09
Enable clipboard

john-peterson:
I am not entering any passwords I am not a child I think I can handle the full keyboard experience like a clipboard. Something few people know about .... Alright it's just me always in good spirit. Thanks for everything guys


agnostic-apollo:
`TYPE_TEXT_VARIATION_VISIBLE_PASSWORD` is required for some keyboards to works properly as far as I remember. If you want full keyboard experience, use text input view.

https://wiki.termux.com/wiki/Touch_Keyboard#Text_Input_View

PR #3337, opened 2023-04-09
Separate the build targets INCOMPLETE 

john-peterson:
I think I can ask for some minor apartheid. As usual everything in this world surprise me. The bottom line is I need to have a separate build target. This is incomplete if somebody has the energy or computer.   ... I am currently on my phone all summer. .... Horrible allergies ... The debug build still call some of the other the main window... You understand.

There has to be a free build target that is not blocked by the signature. I'm not going to uninstall the entire thing or whatever it doesn't make any sense

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -93,6 +93,9 @@ android {
 
         debug {
             signingConfig signingConfigs.debug
+            applicationIdSuffix ".debug"
+            manifestPlaceholders.TERMUX_PACKAGE_NAME = "com.termux.debug"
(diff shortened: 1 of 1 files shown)


agnostic-apollo:
Termux packages are compiled for `com.termux` package and  changing package name would require recompiling all packages from source. If you wish that, check https://github.com/termux/termux-app#Forking

Moreover, termux github builds use debug builds, which require same package name.

https://github.com/termux/termux-app#github

PR #3344, opened 2023-04-15
CI: Trigger builds on jitpack for termux repository only

Biswa96:


agnostic-apollo:
Forks by default should have their libraries published. If they don't want it, then can just disable the workflow if needed from the actions page.

https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow#disabling-a-workflow

PR #3345, opened 2023-04-16
Changed: Bump Robolectric to 4.10

utzcoz:
See https://github.com/robolectric/robolectric/releases/tag/robolectric-4.10.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -142,7 +142,7 @@ android {
 
 dependencies {
     testImplementation "junit:junit:4.13.2"
-    testImplementation "org.robolectric:robolectric:4.9.2"
+    testImplementation "org.robolectric:robolectric:4.10"
     coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:1.1.5"
 }
 


utzcoz:
Friendly ping @agnostic-apollo for reviewing.

agnostic-apollo:
Thanks robolectric-dependabot! Our one test dependent on it appreciates it. ;)

utzcoz:
@agnostic-apollo Thanks for reviewing. Oh, you found out who I am. :)

agnostic-apollo:
You are very welcome. lolz, yeah, I was curious why you were sending just pulls for robolectric updates. ;)

agnostic-apollo:
Ah, cool. More tests are planned for future so robolectric would hopefully be of more use and would then require keeping up to date with updates even more and so would probably require our robolectric-dependabot.

That's great. Hopefully, termux has been treating you nice.

agnostic-apollo:
Yeah, will look into writing more tests when `termux-shared` library is split. Currently, I only write on-device tests for critical or security related stuff like filesystem, unfortunately have limited time to write tests. Thanks for the offer, will surely keep it in mind :)

PR #3359, opened 2023-04-24
Update build tool versions

john-peterson:
It's hard as it is to do this on the phone

Now I'm seeing information in the packages repository that I can use my own fork ID it is specified in the bootstrap that is that contains the base installation somewhere in there or from there the package name is taken. I was puzzled before why or how I would do that so now I've seen that in another repository

I have not yet seen a single word about how to build this application. Thanks for sharing guys. 

There exist an interesting idea to use a common docker image to build inside. Baby I can build the main application also inside a doctor image from the scripts I don't know that. And now I'm giving up because text to speech is not working anymore why did the corrections disappear years ago I could correct all homonyms. Google...

Anyway now I got my APK and I can maybe try to fix my broken style package

I have the control myself to not destroy something if an application does not fall of the system theme. Days are bright nights are dark. How hard is that to understand. Every operating system has had this setting for years. And half of all the idiotic apps I use use either black or white and that's not fall of the system by default. Not even the windows terminal or vs code follows the system by default. .... Give me strength... So anyway I have to fix a few annoying flies I have to shoot down if you annoying helicopters I gave up at TTS doesn't work I'm listening to sort now TTS. Zork the old game I'm sick of everything

This is what I have to fix

Nights dark days white. Follow the SYSTEM THEME
second just use the normal keyboard. And disable spelling in the freetyping input. And make it multi line
And swipe between ALL KEYBOARDS

Can you imagine how annoying it is to have me as a software producer or vice versa. I have to fix a few obvious things and I have to use my phone because I'm on an expedition

And I have to fix Squircle and Amaze to use the Termux file interface. 99% is brilliant guys. I'm just in such enormous pain alll the time from the neural implants b   everything will be okay. I can develop normally on my phone

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:4.2.2"
+        classpath "com.android.tools.build:gradle:7.4.2"
     }
 }
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,7 +17,7 @@ android.useAndroidX=true
 
 minSdkVersion=21
 targetSdkVersion=28
-ndkVersion=22.1.7171670
+ndkVersion=23.1.7779620
 compileSdkVersion=30
 
 markwonVersion=4.6.2
--- a/terminal-emulator/build.gradle
+++ b/terminal-emulator/build.gradle
@@ -56,7 +56,6 @@ dependencies {
 
 task sourceJar(type: Jar) {
     from android.sourceSets.main.java.srcDirs
-    classifier "sources"
 }
 
 afterEvaluate {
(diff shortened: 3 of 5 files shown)


john-peterson:
Right I forgot to say this is the first time I managed to actually destroy a package in my 4-year-old phone. My botched attempt to use my own app ID. Broke the style application there's still a dangling installation that the system can't fix so nothing can be installed I can't install the style application. Which I shouldn't even use inside the main application that should be one dark and one white theme. ONE DARK AND ONE WHITE God made the days bright and the nights dark that's why we have two different themes for our applications

john-peterson:
I managed to put something in /data/data/com.termux.debug. That caused the system crash during uninstallation and left the dangling file that Android can't fix. And I have no access to deleted manually unless I root it which I don't want because I don't have any fast booth or any ADB here where any laptop. I'm in the middle of the goddamn wilderness to get away from the stupid apes no fantasy it's just my thing. 

But but it looks like you actually made it possible for me to use my own fork ID so I will try that sometime thanks of course I appreciate everything thanks guys

agnostic-apollo:
Already updated locally.

PR #3295, opened 2023-03-13
[doc](readme)modified the links in file

nikam14:
improved some links to Hyperlinks.

PR #3368, opened 2023-05-01
Changed(README.md): Improved some links to Hyperlinks

2096779623:
https://github.com/termux/termux-app/pull/3295#issuecomment-1529174808


agnostic-apollo:
Looks good. Thanks. Builds have succeeded on re-run.

But first letter should normally be capital in commit message description after type, like "Improved". This is different from termux-packages style. Packages devs committing here keep using lower case. When changelog is generated with [`create-conventional-changelog`](https://github.com/termux/create-conventional-changelog), then I would have to manually fix each commit message with wrong case or format for consistency. Check examples at https://github.com/termux/termux-app#For-Maintainers-and-Contributors and history https://github.com/termux/termux-app/commits/master.

agnostic-apollo (approved):

PR #3399, opened 2023-05-15
a font for display qr code and New Session menu

morningos:
- Termux font can't show qrcode, so I add a font for display qrcode.
- Due to the impact of full-screen gestures, it is difficult to open the sidebar by sliding, so the operation of creating a new session has been added to the menu.


agnostic-apollo:
>Termux font can't show qrcode, so I add a font for display qrcode.

If you want to use a custom font, then copy it to `~/.termux/font.ttf` or use Termux:Styling app to test other fonts.

Replacing default monospace system font, which may have symbols for user locale/country does not seem like a good idea. You could get the FiraCode font added to Termux:Styling if needed with proper attribution if license if not an issue.

https://github.com/tonsky/FiraCode

https://github.com/termux/termux-styling

https://wiki.termux.com/wiki/Termux:Styling

>Due to the impact of full-screen gestures, it is difficult to open the sidebar by sliding, so the operation of creating a new session has been added to the menu.

If "Add session" option is added to menu because opening drawer is hard, how will you switch to different sessions without opening drawer? Instead just use shortcuts or try changing `terminal-margin-horizontal` property to `0` or higher values. You can also add shortcut values to extra keys macro after also setting the `shortcut.*` properties if I remember correctly.

https://wiki.termux.com/wiki/Terminal_Settings#Handling_terminal_sessions

https://github.com/termux/termux-app/commit/63504f0a

https://wiki.termux.com/wiki/Touch_Keyboard#Extra_Keys_Row

And if sending pull requests in future, create separate commits for different things and follow commit message guidelines in README.

https://github.com/termux/termux-app#For-Maintainers-and-Contributors

PR #3612, opened 2023-10-01
ci: Update actions/checkout to v4

Sushrut1101:


agnostic-apollo:
Already submitted in #3574

PR #3622, opened 2023-10-03
Make it possible to target sdk 34

fornwall:
Here are changes which makes it possible to set `targetSdk=34`, and still start up a successful terminal session on an Android 34 device.

We don't bump the targetSdk yet, as that would require the https://github.com/termux/termux-exec/pull/24 to be present for Android 10+ devices to successfully execute files in the session. But this is one step closer!


app/src/main/java/com/termux/app/TermuxActivity.java
```diff
@@ -932,7 +936,11 @@ private void registerTermuxActivityBroadcastReceiver() {
         intentFilter.addAction(TERMUX_ACTIVITY.ACTION_RELOAD_STYLE);
         intentFilter.addAction(TERMUX_ACTIVITY.ACTION_REQUEST_PERMISSIONS);
 
-        registerReceiver(mTermuxActivityBroadcastReceiver, intentFilter);
+        if (Build.VERSION.SDK_INT >= 28 ) {
```
tuhanayim, line 939:
```suggestion
        if (Build.VERSION.SDK_INT >= 28) {
```

fornwall: Thanks!

PR #3634, opened 2023-10-06
Update build.gradle

Aayush1357:
Update build.gradle dependency tools.build:gradle from 4.2.2 to 7.0.4


agnostic-apollo:
Already changed locally and other changes need to be made as well when bumping android gradle major versions.

PR #3692, opened 2023-11-08
Update november2023

alyssonwolfpoet:
upgrade using the upgrade wizard ide
where I updated simple things like dependencies and grandle
it was tested and compiled and worked correctly
But to update to an even more up-to-date version of Grandle, it will be necessary to make some changes that may complicate the update.


fornwall:
Thanks! Closing this in favour of https://github.com/termux/termux-app/pull/3619.

PR #3705, opened 2023-11-16
added support to remove the termux window from the recent apps menu on EOF

RonakWithCode:
android:excludeFromRecents="true": Ensures that the activity is excluded from the list of recent apps when it finishes.


app/src/main/AndroidManifest.xml
```diff
@@ -58,6 +58,7 @@
             android:launchMode="singleTask"
             android:resizeableActivity="true"
             android:theme="@style/Theme.TermuxActivity.DayNight.NoActionBar"
+            android:excludeFromRecents="true"
```
TomJo2000, line 61: This should only apply after all sessions have exited.
Is that the current behavior with this change?

PR #3708, opened 2023-11-17
added support to remove the termux window from the recent apps menu o…

RonakWithCode:
…n EOF

android:excludeFromRecents="true": Ensures that the activity is excluded from the list of recent apps when it finishes.


fornwall:
Duplicate of #3705.

PR #3712, opened 2023-11-19
Build termux locally with `act`

OscarSaharoy:
Hi, sorry if this is a strange contribution, I am not an android developer!! I am working on some termux edits for myself but thought this might be useful for the project. This is tested on an arm64 mac so sorry if it doesn't work for you. I'm not expecting this to be merged but maybe its interesting, thanks for taking a look :)

I wanted to build the app locally but couldn't figure out how to. I found this project called [`act`](https://github.com/nektos/act) that lets you run github actions locally, so we can use the github actions workflows to build the app. You can do this by following these steps:

1. [install `act`](https://github.com/nektos/act#installation-through-package-managers) and docker (I think that is all the requirements, there may be a few others that I can't remember)
2. checkout this branch, which adds the `localbuild` directory
3. [create a github token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens), this is needed to run `act`. Not sure why its needed to be honest. But you can use a fine-grained token to limit its scope a bit at least, this how my one looks: Screenshot 2023-11-19 at 16 26 19
5. add your token to a new file at `localbuild/.env`. The format for this file should follow `localbuild/.env.example`, so should just contain one line something like `GITHUB_TOKEN=github_pat_ABC123...`
6. create the build environment by running `./localbuild/create-termux-build-env.sh` which builds a docker image called `termux-build-env`. Unfortunately due to the sdk packages its a huge image, around 10GB
7. run the build with `./localbuild/run-github-actions-build.sh` - the build outputs will be placed in `./build/artifacts`
Screenshot 2023-11-19 at 16 29 37

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,4 +40,3 @@ Thumbs.db
 
 # secrets
 .env
-
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,7 @@ local.properties
 .swp
 ehthumbs.db
 Thumbs.db
+
+# secrets
+.env
+
--- a/localbuild/.env.example
+++ b/localbuild/.env.example
@@ -1,0 +1,1 @@
+GITHUB_TOKEN=
--- a/localbuild/Dockerfile
+++ b/localbuild/Dockerfile
@@ -1,0 +1,16 @@
+from eclipse-temurin:11.0.21_9-jdk
+
+ARG ANDROID_SDK_ROOT="/usr/lib/android-sdk"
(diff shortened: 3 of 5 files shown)


tareksander:
I always just build Termux with the Android Studio IDE, it also has a Mac version.

agnostic-apollo:
I agree with @tareksander, android studio is the standard way to build apks. If you want command line, use `gradle` to build, which is what both android studio and the action use.

PR #3720, opened 2023-12-02
Add support for apps supporting PROCESS_TEXT intents.

caarmen:
In the context menu that appears when a user selects text, add a new submenu which includes third-party activities which handle the `PROCESS_TEXT` intent. This will allow processing the selected text in other apps.

Some links about this type of interaction with other apps:
* Manifest config required to work on Android 30+: https://developer.android.com/training/package-visibility/use-cases
* `PROCESS_TEXT` intent action: https://developer.android.com/reference/android/content/Intent#ACTION_PROCESS_TEXT)
* Blog post by Ian Lake on how to make an app appear in the system text selection menu: https://medium.com/androiddevelopers/custom-text-selection-actions-with-action-process-text-191f792d2999

Since the text selection menu is customized in termux, if we want the text selection menu to include this functionality, we have to add it ourselves (this commit :) ).

Here's a demo of selecting text, and opening it in Google Translate, then in Wikipedia:

[process_text.webm](https://github.com/termux/termux-app/assets/1731388/e6bc5993-a421-49f2-b868-41cfee0afe68)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -233,4 +233,14 @@
 
     </application>
 
+    <!-- Starting with api level 30, we need to declare intent actions inside a queries element, 
+    to be able to find activities which support a given intent
+    https://developer.android.com/training/package-visibility/use-cases#custom-text-selection-actions -->
+    <queries>
+        <intent>
+          <action android:name="android.intent.action.PROCESS_TEXT" />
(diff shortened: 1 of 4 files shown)


agnostic-apollo:
Thanks for the pull, this looks good and mergable in essence. Will have to test it and will merge with the others before next release.

Issue #3738, opened 2023-12-16
Cursor colour turns grayish black on turning on dark mode.


KanhaKanhaiya:
### Problem description

The cursor turns grayish black which is almost invisible on turning on Dark Mode from device settings in mobile. It does not changes if I uncomment `use-black-ui = true`.

### Steps to reproduce the behavior.

In a device with Xiaomi MIUI :
• Turn on dark mode from Settings.
• Open Termux.
• Observe the cursor colour.

### What is the expected behavior?

The cursor should remain white.

### System information

* Termux application version: 0.118.0
* Android OS version: 10
* Device model: Redmi

KirkSuD:
Try to [turn off dark mode on a per app basis in the device settings](https://github.com/termux/termux-app/issues/3036#issuecomment-1273430688).
My device is also MIUI, and dark mode makes a lot of colors wrong, ...in a weird way.
Here's some comparison: [256 colors with](https://files.catbox.moe/2pno4u.jpg) & [without dark mode](https://files.catbox.moe/g5o6vz.jpg), [HSV with](https://files.catbox.moe/w4eudb.jpg) & [without dark mode](https://files.catbox.moe/v5ig9n.jpg).
(In dark mode, black text are white, some black colors are not black, some blue colors are not blue.)
I tried [Termux](https://github.com/termux/termux-styling)[:Styling](https://wiki.termux.com/wiki/Termux:Styling) & `colors.properties`, then I found [Termux is already using](https://github.com/termux/termux-app/blob/master/terminal-emulator/src/main/java/com/termux/terminal/TerminalColorScheme.java) [the standard colors](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit).

Issue #2653, opened 2022-03-11
[Bug]:  0.118.0 No cursor when using black text on white background (regression: worked in 0.117.0)


Jamie-Landeg-Jones:
### Problem description

Whilst it was fine in 0.117.0, the cursor is no longer visible with 0.118.0 when using a black text on white background theme.

### Steps to reproduce the behavior.

install 0.118.0 and set the foreground and background colours appropriately, or install one of the black on white themes

### What is the expected behavior?

A visible cursor!

### System information

* Termux application version: 0.118.0
* Android OS version: 9
* Device model: Chromebox Reference

Jamie-Landeg-Jones:
I was just using black background colour and white foreground in colors.properties but the same thing happens with any white foreground, for instance, the "black on white" style.

https://user-images.githubusercontent.com/483894/157988092-6debd73f-9f37-445f-8d71-7cea04a7508f.mp4


Attached is a quick demonstration - at the end i'm pressing cursor-left - you can see how the text on the command line
responds

I just had a further look around, and this looks suspect: https://github.com/termux/termux-app/pull/2228/commits/4a74618f17a65bcafba67b7bf85273390dfa64cb

I tried adding this to colors.properties but it didn't do anything:
 was just using black background colour and white foreground in colors.properties but the same thing happens with any white foreground, for instance, the "black on white" style.

Attached is a quick demonstration - at the end i'm pressing cursor-left - you can see how the text on the command line
responds

I just had a further look around, and this looks suspect: https://github.com/termux/termux-app/pull/2228/commits/4a74618f17a65bcafba67b7bf85273390dfa64cb

I tried adding this to colors.properties but it didn't do anything:

cursor: #ffa9aaa9

I can't compile the source at the moment, so I wasn't able to try regressing the above patch.

It woould seem to be useful if the default cursor colour was dark when the background colour is close to white. Or maybe an option to set the cursor [colour?]

Let me know if you need anything else.

Thanks, Jamie
cursor: #ffa9aaa9

I can't compile the source at the moment, so I wasn't able to try regressing the above patch.

It woould seem to be useful if the default cursor colour was dark when the background colour is close to white. Or maybe an option to set the cursor [colour?]

Let me know if you need anything else.

Thanks, Jamie

agnostic-apollo:
> cursor: #ffA9AAA9

[...]

PR #3779, opened 2024-01-09
fix the problem Cursor colour turns grayish black on turning on dark …#3738

VyankateshA1:
…mode. #3738


agnostic-apollo:
Agreed with sylirre. May be related to https://github.com/termux/termux-app/issues/2653 which isn't available in 0.118.0

Roblox23W12 (commented):
Request changes

Issue #3709, opened 2023-11-17
[Feature]: Persist Sessions Across Restarts - Save / Restore


Splizard:
### Feature description

It doesn't feel good to create a few sessions, name them for organisational purposes and then restart the app (or android restarts it) and they just go away. I understand anything running in those sessions would be terminated - and that's fine but I would like the history and session names to be somewhat preserved.

Since android goes to sleep often, it feels like an essential thing to address.

### Additional information

if you've ever used the terminal on a Mac for example you'll notice when you restart the computer with restore sessions on, it will include scrollable history and any tabs/windows will be preserved. I think the working directory is also preserved, which is helpful because it can be awkward to get back to the right directory in Termux rather than a "Open Terminal Here" often available on desktop/laptop platforms.

agnostic-apollo:
If you use `exit` command or `Ctrl + D` to close the session, then your history should be saved by your shell.

If you use notification `Exit` button or android kills the app, then it won't be. Related https://github.com/termux/termux-app/issues/1988#issuecomment-812684788

Mac just stores the scrollback history, not process state.

https://apple.stackexchange.com/a/278381

Splizard:
> Mac just stores the scrollback history, not process state.
Yes and this is very convenient. 

Even disregarding the history though, persisting session names and working directories across all types of restarts is very valuable, as effort towards organising these sessions for different purposes is not lost. 

Imagine if web browsers on Android just lose all of your tabs when they are restarted/killed. This would be incredibly frustrating. It doesn't really matter what signal you send to them, they'll happily open up the same locations where you left off.

romanovj:
you can write your own scripts for termux, bash and tmux/screen

there are many examples https://github.com/alessandro-g89/bash-sessions

twaik:
Not possible due to the nature of linux programs.
Unlike Mac or Android it is not possible to save or restore process state of running program. Even in the case if you dump process memory and then restore it nothing will work because we can not restore opened file descriptors, sockets and linked processes. 
As said above the only thing you can do is to close terminals with `exit` command or `Ctrl + D` to save shell history and write some scripts, shortcuts or aliases to open needed programs faster.

Splizard:
Hi @twaik,

Nowhere did I mention saving or restoring process state. Only the working directory if possible and even better if you could see some of the history by scrolling to remember what you were doing.

The point is, if I create 5 different sessions in the Termux sidebar and name them all. I want these named tabs to persist so I can use them as markers for different activities.

[...]

PR #3821, opened 2024-02-04
Add ACTION_SERVICE_STOP intent to only stop a single AppShell

lvogt:
This adds the possibility to basically stop a running AppShell that was created by sending a ACTION_SERVICE_EXECUTE intent.

This is used by my ongoing implementation of a cron-like service into TermuxAPI (PR to be published)

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxService.java
+++ b/app/src/main/java/com/termux/app/TermuxService.java
@@ -8,6 +8,7 @@ import android.app.Service;
 import android.content.Context;
 import android.content.Intent;
 import android.content.res.Resources;
+import android.net.Uri;
 import android.net.wifi.WifiManager;
 import android.os.Binder;
 import android.os.Build;
@@ -153,6 +154,10 @@ public final class TermuxService extends Service implements AppShell.AppShellCli
                 case TERMUX_SERVICE.ACTION_SERVICE_EXECUTE:
                     Logger.logDebug(LOG_TAG, "ACTION_SERVICE_EXECUTE intent received");
                     actionServiceExecute(intent);
+                    break;
+                case TERMUX_SERVICE.ACTION_SERVICE_STOP:
+                    Logger.logDebug(LOG_TAG, "ACTION_SERVICE_STOP intent received");
+                    actionServiceStop(intent);
                     break;
                 default:
                     Logger.logError(LOG_TAG, "Invalid action: \"" + action + "\"");
@@ -352,6 +357,23 @@ public final class TermuxService extends Service implements AppShell.AppShellCli
             updateNotification();
 
         Logger.logDebug(LOG_TAG, "WakeLocks released successfully");
+    }
+
+    private void actionServiceStop(Intent intent) {
+        if (intent == null) {
+            Logger.logError(LOG_TAG, "Ignoring null intent to actionServiceStop");
+            return;
+        }
+
+        int gracePeriod = IntentUtils.getIntegerExtraIfSet(intent, TERMUX_SERVICE.EXTRA_TERMINATE_GRACE_PERIOD, 5000);
+
+        Uri executableUri = intent.getData();
+        String executable = UriUtils.getUriFilePathWithFragment(executableUri);
+        String shellName = ShellUtils.getExecutableBasename(executable);
+        AppShell appShell = getTermuxTaskForShellName(shellName);
+        if (appShell != null) {
+            appShell.terminateIfExecuting(getApplicationContext(), gracePeriod, true);
+        }
     }
 
     /** Process {@link TERMUX_SERVICE#ACTION_SERVICE_EXECUTE} intent to execute a shell command in
--- a/termux-shared/src/main/java/com/termux/shared/termux/TermuxConstants.java
+++ b/termux-shared/src/main/java/com/termux/shared/termux/TermuxConstants.java
@@ -993,6 +993,9 @@ public final class TermuxConstants {
             /** Intent action to execute command with TERMUX_SERVICE */
             public static final String ACTION_SERVICE_EXECUTE = TERMUX_PACKAGE_NAME + ".service_execute"; // Default: "com.termux.service_execute"
 
+            /** Intent action to execute command with TERMUX_SERVICE */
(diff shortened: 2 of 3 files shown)


app/src/main/java/com/termux/app/TermuxService.java
```diff
@@ -354,6 +359,23 @@ private void actionReleaseWakeLock(boolean updateNotification) {
...
+        if (intent == null) {
+            Logger.logError(LOG_TAG, "Ignoring null intent to actionServiceStop");
+            return;
+        }
+
+        int gracePeriod = IntentUtils.getIntegerExtraIfSet(intent, TERMUX_SERVICE.EXTRA_TERMINATE_GRACE_PERIOD, 5000);
+
+        Uri executableUri = intent.getData();
+        String executable = UriUtils.getUriFilePathWithFragment(executableUri);
+        String shellName = ShellUtils.getExecutableBasename(executable);
```
agnostic-apollo, line 372: There would be conflicts if same executable was run from someplace else, and that function only returns first one. You should generate at least a six digit alphanumeric `[A-Z0-9]` random string and store that in cron tab in addition to the id and when sending execute or stop intent, pass that as `TERMUX_SERVICE.EXTRA_SHELL_NAME`, so that the executable for the job is killed, and not any other.

https://github.com/termux/termux-app/commit/d287734abae95107e73c763ccbd807315ca9a9c0

https://github.com/termux/termux-app/issues/3709

lvogt: Good point.
Is there a character limit on the shell name?
Or the other way around: Could I just use a UUID?

agnostic-apollo: There is no limit currently, but there should be, ideally `255`, equal to `NAME_NAME`.

UUID would be too long and harder to see in app shells UI list and logs, just use a six character alphanumeric `[a-zA-Z0-9]` string, like `mktemp` template uses for unique files. Following should work, make sure `<script_name>_<public_id>_<private_id>` doesn't already exist in cron tab, otherwise regenerate `private_id`. The `public_id` here is the incrementing number you are currently using.

```java
char[] allowedCharsArray = ("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789").toCharArray();
char[] private_id = new char[6];
Random random = new SecureRandom();
for (int i = 0; i < private_id.length; i++) {
    private_id[i] = allowedCharsArray[random.nextInt(allowedCharsArray.length)];
}
```

https://man7.org/linux/man-pages/man3/mktemp.3.html

https://bugs.python.org/issue12015

lvogt: I see. Thanks for the example.

I changed the method to only work if the shell name is explicitly set.

termux-shared/src/main/java/com/termux/shared/termux/TermuxConstants.java
```diff
@@ -993,6 +993,9 @@ public static final class TERMUX_SERVICE {
             /** Intent action to execute command with TERMUX_SERVICE */
             public static final String ACTION_SERVICE_EXECUTE = TERMUX_PACKAGE_NAME + ".service_execute"; // Default: "com.termux.service_execute"
 
+            /** Intent action to execute command with TERMUX_SERVICE */
+            public static final String ACTION_SERVICE_STOP = TERMUX_PACKAGE_NAME + ".service_execution_stop"; // Default: "com.termux.service_execute"
```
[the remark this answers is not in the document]
lvogt: Fixed.

termux-shared/src/main/java/com/termux/shared/termux/TermuxConstants.java
```diff
@@ -1046,6 +1049,9 @@ public static final class TERMUX_SERVICE {
              * be created in {@link #EXTRA_RESULT_DIRECTORY} if {@link #EXTRA_RESULT_SINGLE_FILE} is
              * {@code false} for the TERMUX_SERVICE.ACTION_SERVICE_EXECUTE intent */
             public static final String EXTRA_RESULT_FILES_SUFFIX = TERMUX_PACKAGE_NAME + ".execute.result_files_suffix"; // Default: "com.termux.execute.result_files_suffix"
+            /** Intent {@code long} extra for graceperiod between SIGTERM and SIGKILL
+             * for the TERMUX_SERVICE.ACTION_SERVICE_STOP intent */
+            public static final String EXTRA_TERMINATE_GRACE_PERIOD = TERMUX_PACKAGE_NAME + ".execute.stop.delay"; // Default: "com.termux.execute.stop.delay"
```
agnostic-apollo, line 1054: This should be `stop_delay`

app/src/main/java/com/termux/app/TermuxService.java
agnostic-apollo: Welcome. Makes sense. I think it would be better to call `getTermuxTaskForShellName()` in a `while` loop until it starts returning `null`, so that all shells with same name get killed, since currently only first one would get killed, which may not be the one caller intended to kill. To prevent duplicates, callers should start shells with a unique shell name, like the cron API would do.

termux-shared/src/main/java/com/termux/shared/termux/TermuxConstants.java
lvogt: Applied

app/src/main/java/com/termux/app/TermuxService.java
lvogt: Good point - I had not considered multiple shells with the same name.

termux-shared/src/main/java/com/termux/shared/termux/TermuxConstants.java
agnostic-apollo: The extra name was  also different. Both should be same, naming them `sigkill_delay_on_stop` seems better, with respective case.

agnostic-apollo: Will have to rename `gracePeriod` and `gracePeriodMsec` as well

Issue #3839, opened 2024-02-16
[Bug]: Large amount of invisible unicode characters causes freeze and crash


Kimapr:
### Problem description

Printing a specific file to terminal freezes and crashes Termux. The file is valid UTF-8 but contains large amounts of various invisible Unicode characters, including zero-width spaces, word joiners, variant selectors, etc.

### Steps to reproduce the behavior.

1. Download https://kimapr.net/lappy/matroska.js
2. Print it to the terminal using `cat` or similar

### What is the expected behavior?

Termux does not freeze or crash

### System information

* Termux application version: 0.118.0
* Android OS version: 11
* Device model: TECNO Spark 8C KG5j

tstein:
Sent #3840 with a fix for the crash, but there's still something asymptotically disastrous going on here. I gave it 15+ minutes trying to print the bad line before ^C.

PR #3840, opened 2024-02-16
Fixed: Widen TerminalRow's index so very long lines don't overflow it.

tstein:
This prevents the crash in #3839, but not by enough to call it done. Termux gets stuck on the very bad line and is responsive, but laggy enough to trigger ANRs. It's possible to ^C in this state and get your terminal back with no apparent damage.

### DIFF
--- a/terminal-emulator/src/main/java/com/termux/terminal/TerminalRow.java
+++ b/terminal-emulator/src/main/java/com/termux/terminal/TerminalRow.java
@@ -11,11 +11,37 @@ public final class TerminalRow {
 
     private static final float SPARE_CAPACITY_FACTOR = 1.5f;
 
+    /**
+     * Max combining characters that can exist in a column, that are separate from the base character
+     * itself. Any additional combining characters will be ignored and not added to the column.
+     *
+     * There does not seem to be limit in unicode standard for max number of combination characters
+     * that can be combined but such characters are primarily under 10.
+     *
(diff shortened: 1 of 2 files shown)


agnostic-apollo:
So I looked into it, it does fix the crash, but like you said terminal still becomes laggy, the code for `TerminalRow.setChar()` is complex and would take time to understand how it actually works and if even converting `short` to `int` is the right move logically. Fornwall made changes in the initial commit to the jakpal's terminal commit.

https://github.com/jackpal/Android-Terminal-Emulator/commit/9a47042620bec87617f0b4f5d50568535668fe26

https://github.com/jackpal/Android-Terminal-Emulator/pull/338

https://github.com/termux/termux-app/commit/a18ee58f7a4f56cab6ba4a8cacaa70c4c76379b3#diff-f84d215b18106c037e01986a3968fa54b74691174a78fcc99493f745d3805be5

agnostic-apollo:
>but I don't think I'll figure out the full fix in the near future. :)

What do you mean, start working on it ASAP! Finish what you started! Chop chop!

I'll take another look on upper bounds and why short was used before, and if it was just a mistake. If it's reasonable, then will merge, I agree, lag is better than crash.

agnostic-apollo:
Check fcd7a34e for fix and details, hopefully it doesn't break anything. The build in https://github.com/termux/termux-app/actions/runs/8378462868 should be tested before this is merged.

cc @Kimapr

PR #3870, opened 2024-03-03
termux-app is not booted and not properly working in background on Android 10 (on Xiaomi Redmi 7A)

AntonSamokat:
For details see:
https://stackoverflow.com/a/19856267/1455694
https://stackoverflow.com/a/63250729/1455694
https://www.reddit.com/r/tasker/comments/d7whyj/android_10_and_auto_starting_apps/
https://stackoverflow.com/q/64642362/1455694

Not sure about real need for:
            android:enabled="true"
            android:exported="true"
in app/src/main/AndroidManifest.xml

Related pull request for termux-boot:
termux-boot is not working on Xiaomi Redmi 7A on Android 10
https://github.com/termux/termux-boot/pull/195

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -196,7 +196,8 @@
 
         <service
             android:name=".app.TermuxService"
-            android:exported="false" />
+            android:exported="true"
+            android:enabled="true" />
 
         <service
             android:name=".app.RunCommandService"
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -13,6 +13,7 @@ import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.IBinder;
+import android.provider.Settings;
 import android.view.ContextMenu;
 import android.view.ContextMenu.ContextMenuInfo;
 import android.view.Gravity;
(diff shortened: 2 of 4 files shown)


rpla14024 (commented):
Good

app/src/main/AndroidManifest.xml
```diff
@@ -196,7 +196,8 @@
 
         <service
             android:name=".app.TermuxService"
-            android:exported="false" />
+            android:exported="true"
```
agnostic-apollo, line 199: Enabling export of a service, activity, receiver, provider allows other apps to call them without any permission. This will allow any random app to run commands in termux without user permission, and even with root if termux has root permissions, i.e you are creating a privilege escalation vulnerability.

app/src/main/AndroidManifest.xml
```diff
@@ -196,7 +196,8 @@
...
+            android:enabled="true" />
```
agnostic-apollo, line 200: Services are enabled by default.

https://developer.android.com/guide/topics/manifest/service-element#enabled

app/src/main/java/com/termux/app/TermuxActivity.java
```diff
@@ -198,6 +199,20 @@ public final class TermuxActivity extends AppCompatActivity implements ServiceCo
     @Override
     public void onCreate(Bundle savedInstanceState) {
         Logger.logDebug(LOG_TAG, "onCreate");
+
+        // For details see:
```
agnostic-apollo, line 203: TermuxService also requests the permission if its required to start foreground terminal session from background. Permissions must not be requested unless needed.

https://github.com/termux/termux-app/blob/8e3a8980a849046adbd4156741e1d84047ee1df6/app/src/main/java/com/termux/app/TermuxService.java#L717

app/src/main/java/com/termux/app/TermuxActivity.java
```diff
@@ -257,9 +272,17 @@ public void onCreate(Bundle savedInstanceState) {
         FileReceiverActivity.updateFileReceiverActivityComponentsState(this);
 
         try {
+            Intent serviceIntent;
```
agnostic-apollo, line 275: This is one issue that has already been solved locally and fix will be available in next version.

agnostic-apollo:
There were various issues with why termux-boot did not work properly and it will not be used from next termux release as boot logic has been rewritten into termux-app itself.

PR #3894, opened 2024-03-24
Create Python

Momohsankoh:
Good


TomJo2000:
Packages can be installed by the user after installation.
If you would like to include additional packages in a customized version of the bootstrap archive,
please see the documentation for doing so at:
https://github.com/termux/termux-packages/wiki/For-maintainers#bootstraps

However `python`, `clang` and `git` are out of scope for the official bootstrap archives and will more than likely not be added to them.

PR #3908, opened 2024-04-05
Adds immersive mode for Samsung Dex

shmatt:
Hides the top bar when using full screen mode in Samsung Dex


afonsoft:
Nice

gutierri:
What would this immersive mode be? Does it hide some components (such as shortcuts footer [CTRL, Pg Up...]) from the interface only when in full screen mode?

shmatt:
> What would this immersive mode be? Does it hide some components (such as shortcuts footer [CTRL, Pg Up...]) from the interface only when in full screen mode?

Samsung Dex usually displays a titlw bar across the top of an app (similar to any desktop OS) showing minimise, maximise and close buttons.

This PR hides that bar.  A similar change has been implemented in Termux-X11

gutierri:
> > What would this immersive mode be? Does it hide some components (such as shortcuts footer [CTRL, Pg Up...]) from the interface only when in full screen mode?
> 
> Samsung Dex usually displays a titlw bar across the top of an app (similar to any desktop OS) showing minimise, maximise and close buttons.
> 
> This PR hides that bar. A similar change has been implemented in Termux-X11

Yes, there is indeed something similar to this in X11 (I use it daily), but there it is possible to enable it through the settings. It would be interesting to have something like that here too.

From what I understand, it would be a fullscreen mode (F11).

Thanks for the clarifications!

shmatt:
> Yes, there is indeed something similar to this in X11 (I use it daily), but there it is possible to enable it through the settings. It would be interesting to have something like that here too.
> 
> From what I understand, it would be a fullscreen mode (F11).
> 
> Thanks for the clarifications!

Termux has an immersive flag already, which does a similar thing in normal android mode (hides the system bar)

This piggy back on that same flag, as it is the same intention but in different environments

Navisystem77 (commented):
Ok

Pawaris2732:
[](url)

robertkirkman:
I can confirm this works, it just has a lot of merge conflicts because of how there is an old version of #2973  mixed in with the immersive mode. Is that important for this, or should this be separated to be a PR just containing the samsung dex immersive mode by itself?

shmatt:
> I can confirm this works, it just has a lot of merge conflicts because of how there is an old version of #2973 mixed in with the immersive mode. Is that important for this, or should this be separated to be a PR just containing the samsung dex immersive mode by itself?

No, not important. I added sixel for my own use but must have unintentionally combined the work.

shmatt:
@robertkirkman I have opened a new PR excluding the sixel merge, and up-to-date with master

https://github.com/termux/termux-app/pull/4265

shmatt:
Closed in favour of https://github.com/termux/termux-app/pull/4265

Issue #3935, opened 2024-04-30
[Bug]: termux-file-editor running after sharing links with unknown/not included uri schemes


macearl:
### Problem description

When sharing URLs which are not part of the [WEB_URL pattern from AOSP](https://github.com/aosp-mirror/platform_frameworks_support/blob/a9ac247af2afd4115c3eb6d16c05bc92737d6305/compat/src/main/java/androidx/core/util/PatternsCompat.java#L211) or Magnet links (added through https://github.com/termux/termux-app/issues/1339 ).

The file editor gets opened instead of the url opener. 

I would suggest adding at least these URI schemes as well:
- gopher 
- irc
- irc6
- ircs
- jabber
- mailto
- nfs
- nzblnk
- rtmp
- sftp
- smb

### Steps to reproduce the behavior.

Example for SFTP:

1. Install the midnight commander (`pkg install mc`)
2.  Share a SFTP link with termux

### What is the expected behavior?

The url opener script should be executed or an error message, that the url opener does not exist should be shown instead the file opener / editor is shown

### System information

(not really relevant) 
* Termux application version: 0.118.0
* Android OS version: 14
* Device model: Xiaomi Poco F1

PR #3936, opened 2024-04-30
Add additional uris to FileReceiverActivity.java

macearl:
Added
- gopher
- irc
- irc6
- ircs
- jabber
- mailto
- nfs
- nzblnk
- rtmp
- sftp
- smb

Uris

Fixes: https://github.com/termux/termux-app/issues/3935

### DIFF
--- a/app/src/main/java/com/termux/app/api/file/FileReceiverActivity.java
+++ b/app/src/main/java/com/termux/app/api/file/FileReceiverActivity.java
@@ -56,7 +56,10 @@ public class FileReceiverActivity extends AppCompatActivity {
 
     static boolean isSharedTextAnUrl(String sharedText) {
         return Patterns.WEB_URL.matcher(sharedText).matches()
-            || Pattern.matches("magnet:\\?xt=urn:btih:.*?", sharedText);
+            || Pattern.matches("magnet:\\?xt=urn:btih:.*?", sharedText)
+            || Pattern.matches("nzblnk:\\?((t|h|g|p)=(.*)&?)+", sharedText)
+            || Pattern.matches("(jabber|mailto|xmpp):.+", sharedText)
+            || Pattern.matches("(gopher|irc(6|s)?|nfs|rtmp|sftp|smb)://.+", sharedText);
     }
 
     @Override


agnostic-apollo:
There a lot of other protocols that are supported by https://github.com/chedim/termux-launcher/blob/17fc6b66e55b4fb53c58a439ad35cc2aa410d0e8/termux-shared/src/main/java/com/termux/shared/termux/data/TermuxUrlUtils.java for terminal url extractor that also need to be supported. Just adding the above ones is not enough and regexes used are not compliant with rfc approved characters/formats, like `Pattern.WEB_URL` handles them.

I have locally ported the `Pattern` class regexes and combined the protocols in `TermuxUrlUtils` and fixed the regexes so both `termux-url-opener` and terminal url extractor have consistent experience. Changes should be available in next pre release when I push changes. There are conflicts as `FileReceiverActivity` class does not exist anymore in my local changes and has been rewritten, so cannot push to `master` currently.

- https://cs.android.com/android/platform/superproject/+/android-14.0.0_r1:frameworks/base/core/java/android/util/Patterns.java;l=326

agnostic-apollo:
Thanks for the report and attempt though.

macearl:
Thanks, I was not aware of the changes in your fork.
I have 2 additional questions however (if that's alright):
- Am I correct in assuming these changes will eventually be merged into the upstream `termux/termux-app` project or will they only be done in the forked project?
- if I'm reading the code right and the old class got removed entirely this removes the support for magnet links right? I think they should still be supported, the format is also a bit different and probably doesn't fit that neatly into the others: https://www.iana.org/assignments/uri-schemes/prov/magnet I would also like to see nzblnk included ( https://nzblnk.info/ ) which would be useful for the existing sabnzbd package, once it is supported it could also be useful to add a package for https://github.com/Tensai75/nzb-monkey-go (pretty easy to install through go right now so not strictly necessary).

agnostic-apollo:
>if that's alright

Of course it is.

>Am I correct in assuming these changes will eventually be merged into the upstream termux/termux-app project 

Yes, my local changes will be pushed to  termux/termux-app for the next app version.

> if I'm reading the code right and the old class got removed entirely this removes the support for magnet links right? I think they should still be supported

If you are referring to the `TermuxUrlUtils` class link, yes, it does not support `magnet`, but my local changes do support `magnet` and `nzblnk`. The new supported protocols are combination of the one in `TermuxUrlUtils`, `Pattern.WEB_URL` and your pull. Both `TermuxUrlUtils` and `FileReceiverActivity` class have been removed in my local changes with new implementation.

Currently, `FileReceiverActivity` for `magnet` and new protocols in your pull are using `.*` and `.+` in url regex, which is wrong and breaks matching, as it matches any character until newline, and not just the ones supported by rfc, for example `magnet:?xt=urn:btih:hash https://example.com` would match as a single url and space character and later would get matched too.

`magnet` protocol also supports other URNs other than `bith`.

- https://en.wikipedia.org/wiki/Magnet_URI_scheme#Exact_Topic_(xt)
- https://en.wikipedia.org/wiki/Uniform_Resource_Name

`mailto` protocol supports other formats too other than a single email.

- https://en.wikipedia.org/wiki/Mailto

`xmpp` (and possibly `jabber`) URIs may contain `://` instead of just `:`.

- https://www.cisco.com/c/en/us/td/docs/voice_ip_comm/jabber/11_5/CJAB_BK_D00D8CBD_00_deployment-installation-guide-cisco-jabber115/CJAB_BK_D00D8CBD_00_deployment-installation-guide-cisco-jabber115_chapter_010011.html#JABW_TP_PF21EDAF_00

PR #4005, opened 2024-06-09
0.118: bump bootstraps to v2024.06.09-r1

Grimler91:
Installation and pkg/apt usage, and general use, works fine on both arm and aarch64

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -195,11 +195,11 @@ clean {
 
 task downloadBootstraps() {
     doLast {
-        def version = "2022.01.07-r1"
-        downloadBootstrap("aarch64", "0fe6d0159d12fcb8baf7750ce9072b9b36f742662b02ad4da145ab85873614cd", version)
-        downloadBootstrap("arm",     "0a6014e2ec3b7079524fee3caabd02be05bcb4add3c6cd9e5ad98408b428c717", version)
-        downloadBootstrap("i686",    "c55369a9af1316dc3d99457aa23cce64b29c1e60e375159352c0d4b9cfed4ac6", version)
-        downloadBootstrap("x86_64",  "e93a7c66d15edb7d1b5ceca906255c85ead35a8b6a52f93524e117cfb07e6778", version)
+        def version = "2024.06.09-r1" + "+apt-android-7"
+        downloadBootstrap("aarch64", "23390696860d0375729b65cc11a13fed1a04a480071bdcdd7e307ab3f1cff419", version)
+        downloadBootstrap("arm",     "f85c846fd3718f4fc54f3aebf28f93394b2a05e33c9d104437a06dc631037f94", version)
+        downloadBootstrap("i686",    "f4008384b64d2e47c2cd4d49d2229744f76af81e404fab0066ab4679503dc6de", version)
+        downloadBootstrap("x86_64",  "78eebd6b9719a1092d1a874c2de425399601cc88a9e91e6d010e6bfd6c54d8f1", version)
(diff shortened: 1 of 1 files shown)


Grimler91:
Running tests from https://github.com/termux/termux-exec/pull/24 succeeds if I remove --ld-preload arg and just install and use libtermux-exec.so from $PREFIX/lib/ instead, that is I apply:

```diff
diff --git a/Makefile b/Makefile
index afc3d63..cf5769e 100644
--- a/Makefile
+++ b/Makefile
@@ -178,7 +178,7 @@ deb: all
 
 test: all
        $(MAKE) TERMUX_EXEC__INSTALL_PREFIX=install install
-       tests/termux-exec-tests --ld-preload="install/lib/libtermux-exec.so" --tests-path="install/libexec/installed-tests/termux-exec" -vvv all
+       tests/termux-exec-tests --tests-path="install/libexec/installed-tests/termux-exec" -vvv all
 
 test-unit: all
        $(MAKE) TERMUX_EXEC__INSTALL_PREFIX=install install
@@ -186,7 +186,7 @@ test-unit: all
 
 test-runtime: all
        $(MAKE) TERMUX_EXEC__INSTALL_PREFIX=install install
-       tests/termux-exec-tests --ld-preload="install/lib/libtermux-exec.so" --tests-path="install/libexec/installed-tests/termux-exec" -vvv runtime
+       tests/termux-exec-tests --tests-path="install/libexec/installed-tests/termux-exec" -vvv runtime
 
 
 

```

haven't tested with termux-api and termux-am, but otherwise all tests succeed on both arm and aarch64
[arm.log](https://github.com/user-attachments/files/15754146/arm.log)
[aarch64.log](https://github.com/user-attachments/files/15754147/aarch64.log)

fornwall:
Ok to merge and tag a `0.118.1` release?

Would be nice to get it out to fix the issue of people being prompted for updating away from F-Droid to Google Play. See also https://github.com/termux/termux-app/issues/4012#issuecomment-2164762685 - a systematic version code scheme can come later perhaps.

Grimler91:
I think @agnostic-apollo mentioned that it would be good to cherry-pick some additional fixes/improvements from master before tagging a release, for example workflows so that apk's are built and uploaded to the new release

agnostic-apollo:
Workflow is broken in master branch too, I fixed it in my local `master` branch a while back and have also fixed it in my new local release branch.

agnostic-apollo:
@Grimler91 Found the reason for `make test` failing after exporting `TERMUX_EXEC__LOG_LEVEL=4`.

```
Termux.exec: Failed to access executable path '/data/data/com.termux/files/usr/libexec/installed-tests/termux-exec/files/exec/print-args-binary': No such file or directory 
```

Even though the test files path was passed to `termux-exec-tests` with `--tests-path` to the `install` sub directory under source directory, the  `runtime-binary-tests.c` was using hardcoded (macro) tests paths under `$PREFIX/libexexec/*` and then failed to find the paths. I have fixed it by making it check the `TERMUX_EXEC__TESTS_PATH` env variable to see if its set (with fallback to macro) and dynamically generating the test file paths.

Another error was that `LD_PRELOAD` being passed with `--ld-preload` wasn't converted to an absolute path, which then failed due to linking error for later `bash` tests.

twaik:
Probably it will be a good idea to update release with newest bootstrap with same version but bumped versionCode, without making any new commits to release code. So both f-droid and github releases will have newest bootstrap even if the apk code was not changed.

PR #4072, opened 2024-07-30
Refactored Identifiers

congyuluo:
Refactored certain identifiers into more appropriate names according to their usage in the code. This is to improve code clarity.


agnostic-apollo:
Sorry. Seems like change for the sake of change.

PR #4106, opened 2024-08-16
Make Shift-PgUp and Shift-PgDn scroll by pages rather than lines

joshtriplett:
In other terminals, such as gnome-terminal, Shift-PgUp and Shift-PgDn
scroll the screen by a full page, rather than a single line. Adjust
termux to match.


fornwall (approved):
Makes sense - thanks a lot!

PR #4108, opened 2024-08-17
add horisontal swipes

john-peterson:
# test

## mouse 0, alt 0

bash, more

## mouse 0, alt 1

less: designed for manual keyboard input the right jump is too big when swiping 

## mouse 1, alt 0

r2 -c "aa;s main;VV" a.out: need patch https://github.com/john-peterson/radare2/tree/scroll

## mouse 1, alt 1

links: works but really needs patch to never overflow terminal width 

vim: unessecery and unimplemented 

mcedit help menu: unimplemented. pressing left crashes to bash that now print escape codes in plain text when swiping


Jbub5:
I liked your idea, so I made it possible to move the cursor horizontally by swiping a finger across the screen.
During the implementation, I encountered the problem of triggering on vertical scrolling, which leads to auto-scrolling to the entered text, which creates problems because it is impossible to swipe up or down perfectly. As a solution to the problem with horizontal scrolling, setting a flag is used to block auto-scrolling to the entered text. This solution puts a heavy load on CPU, and need to find a better way.
https://github.com/Jbub5/termux-app/commit/8b9a6e274ef2d38af15e6355e199dea4d9e02884

PR #4165, opened 2024-09-23
fix: Mark view as important for AutoFill after requesting an AutoFill

AChep:
Currently the view is still marked as not important even after requesting an autofill.

### DIFF
--- a/terminal-view/src/main/java/com/termux/view/TerminalView.java
+++ b/terminal-view/src/main/java/com/termux/view/TerminalView.java
@@ -101,6 +101,20 @@ public final class TerminalView extends View {
     private int mAutoFillType = AUTOFILL_TYPE_NONE;
 
     /**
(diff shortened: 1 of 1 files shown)


AChep:
Causes https://github.com/AChep/keyguard-app/issues/594

agnostic-apollo:
Thanks for the fix. Makes sense, in the initial commit for AutoFill long ago, the view was marked as not important with a hardcoded value as it wasn't important at all times, hence I kept the logic in my last changes in c2d57f2, but now that we are dynamically setting autofill config, this is the way to go. I assume the keyguard-app wouldn't have been working even before my last changes. 

https://github.com/termux/termux-app/commit/8c80efb904b023cfafab23083df8d5881f201dd3#diff-e8f482031a706124776cc4b4e81898127e6dd8dda09f1ffffa25eba7ff711f0aR23

p.s Please use commit message format guidelines in future.

https://github.com/termux/termux-app#commit-messages-guidelines

PR #4193, opened 2024-10-11
Allow build with empty bootstraps

MatanZ:
To build without bootstrap image (useful apk is only used for upgrading existing installations), use
`export TERMUX_NO_BOOTSTRAP=1`
before running gradle.


Sharique55:
Thanks for this feature

PR #4196, opened 2024-10-13
Create IG

Vyruzxx:
termux-setup-storage
pkg upgrade && update
pkg install git
pkg install curl
pkg install openssh opnessl-tool-y
git clone https://github.com/xd20111/inshackle ls

### DIFF
--- a/IG
+++ b/IG
@@ -1,0 +1,6 @@
+termux-setup-storage
+pkg upgrade && update
+pkg install git
+pkg install curl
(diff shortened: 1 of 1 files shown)


Vyruzxx:
termux-setup-storage
pkg upgrade && update
pkg install git
pkg install curl
pkg install openssh opnessl-tool-y
git clone https://github.com/elizabeth_ibarra21/inshackle ls

PR #4250, opened 2024-11-10
Added display of extra key "SPACE" as "␣" (U+2423).

cymok:
Added display of extra key "SPACE" as "␣" (U+2423). 
Referenced from https://en.wikipedia.org/wiki/Space_bar.


Sharique55:
Ok

PR #4412, opened 2025-02-28
chore: fastlane full_description.txt: reword to simplify translation

yurtpage:
Use allowed HTML tags and use a list.
Remove mention of Subversion, it's epoch gone.
"Enjoy Bash" replaced with just "Command line shells Bash".

This will simplify auto-translations


benja2998 (commented):
A lot of it doesn't make sense like "Command line shells the Bash and Zsh".

PR #4417, opened 2025-03-01
Fixed: fully consume unknown CSI sequences containing non-numeric parameter byte

### LINKED ISSUE #4338: [Bug]: may fail to ignore escape sequences for kitty progressive enhancements
### Problem description

termux seems to fail to parse certain CSI escape sequences, and ends up echoing them.
This is unfortunate because it prevents programs like text editors and shells from using the [kitty keyboard protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/)

This was originally reported in the [fish-shell matrix channel](https://matrix.to/#/!YLTeaulxSDauOOxBoR:matrix.org/$Ra4slTX2o_pIRUiiJY_qsJzxLUTf294Fv5hL_lzI0sE?via=gitter.im&via=matrix.org)
I have not had a chance to reproduce this but it should be easy to do so.

`printf '\x1b[=5u`

This supposedly echoes `5u`

Nothing should be echoed, since the string is a [CSI-prefixed escape sequence](https://en.wikipedia.org/wiki/ANSI_escape_code#Control_Sequence_Introducer_commands)

### System information

not sure

krobelus:
- [x] To-do: this is untested, would appreciate your help on that (I don't have Android Studio or a device set up).

Standard ECMA-48: Control Functions for Coded Character Sets specifies
the format of CSI commands.  Wikipedia has a concise description:
https://en.wikipedia.org/wiki/ANSI_escape_code#Control_Sequence_Introducer_commands

Do this for at least for sequences like "CSI = 5 u" that
contain non-numeric parameter bytes.

This fixes a problem with fish shell 4.0.0 which (for better or worse)
uses that sequence.

This patch introduces a slight inconsistency: for the above example,
"unknownSequence('u')" will be called.  The resulting log message may
be confusing, because we *do* support "CSI u". We should fix this to
log the entire unknown sequence. I can try doing that.

In future we should also fully consume all other unknown CSI commands.
A contrived example would be "CSI ! ! a".  If desired, I'm happy to
fix those as well, as I have some context already (but I don't think
that would block this patch).

Closes #4338

### DIFF
--- a/terminal-emulator/src/main/java/com/termux/terminal/TerminalEmulator.java
+++ b/terminal-emulator/src/main/java/com/termux/terminal/TerminalEmulator.java
@@ -83,6 +83,10 @@ public final class TerminalEmulator {
     private static final int ESC_APC = 20;
     /** Escape processing: "ESC _" or Application Program Command (APC), followed by Escape. */
     private static final int ESC_APC_ESCAPE = 21;
+    /** Escape processing: ESC [ <parameter bytes> */
+    private static final int ESC_CSI_UNSUPPORTED_PARAMETER_BYTE = 22;
+    /** Escape processing: ESC [ <parameter bytes> <intermediate bytes> */
+    private static final int ESC_CSI_UNSUPPORTED_INTERMEDIATE_BYTE = 23;
 
     /** The number of parameter arguments including colon separated sub-parameters. */
     private static final int MAX_ESCAPE_PARAMETERS = 32;
@@ -658,6 +662,10 @@ public final class TerminalEmulator {
                     case ESC_CSI:
                         doCsi(b);
                         break;
+                    case ESC_CSI_UNSUPPORTED_PARAMETER_BYTE:
+                    case ESC_CSI_UNSUPPORTED_INTERMEDIATE_BYTE:
+                        doCsiUnsupportedParameterOrIntermediateByte(b);
+                        break;
                     case ESC_CSI_EXCLAMATION:
                         if (b == 'p') { // Soft terminal reset (DECSTR, http://vt100.net/docs/vt510-rm/DECSTR).
                             reset();
@@ -1057,6 +1065,37 @@ public final class TerminalEmulator {
         for (int i = mCursorCol + 1; i < mColumns; i++)
             if (mTabStop[i] && --numTabs == 0) return Math.min(i, mRightMargin);
         return mRightMargin - 1;
+    }
+
+    /**
+     * Process byte while in the {@link #ESC_CSI_UNSUPPORTED_PARAMETER_BYTE} or
+     * {@link #ESC_CSI_UNSUPPORTED_INTERMEDIATE_BYTE} escape state.
+     *
+     * Parse unsupported parameter, intermediate and final bytes but ignore them.
+     *
+     * > For Control Sequence Introducer, ... the ESC [ is followed by
+     * > - any number (including none) of "parameter bytes" in the range 0x30–0x3F (ASCII 0–9:;<=>?),
+     * > - then by any number of "intermediate bytes" in the range 0x20–0x2F (ASCII space and !"#$%&'()*+,-./),
+     * > - then finally by a single "final byte" in the range 0x40–0x7E (ASCII @A–Z[\]^_`a–z{|}~).
+     *
+     * - https://en.wikipedia.org/wiki/ANSI_escape_code#Control_Sequence_Introducer_commands
+     * - https://invisible-island.net/xterm/ecma-48-parameter-format.html#section5.4
+     */
+    private void doCsiUnsupportedParameterOrIntermediateByte(int b) {
+        if (mEscapeState == ESC_CSI_UNSUPPORTED_PARAMETER_BYTE && b >= 0x30 && b <= 0x3F) {
+            // Supported `0–9:;>?` or unsupported `<=` parameter byte after an
+            // initial unsupported parameter byte in `doCsi()`, or a sequential parameter byte.
+            continueSequence(ESC_CSI_UNSUPPORTED_PARAMETER_BYTE);
+        } else if (b >= 0x20 && b <= 0x2F) {
+            // Optional intermediate byte `!"#$%&'()*+,-./` after parameter or intermediate byte.
+            continueSequence(ESC_CSI_UNSUPPORTED_INTERMEDIATE_BYTE);
+        } else if (b >= 0x40 && b <= 0x7E) {
+            // Final byte `@A–Z[\]^_`a–z{|}~` after parameter or intermediate byte.
+            // Calling `unknownSequence()` would log an error with only a final byte, so ignore it for now.
+            finishSequence();
+        } else {
+            unknownSequence(b);
+        }
     }
 
     /** Process byte while in the {@link #ESC_CSI_QUESTIONMARK} escape state. */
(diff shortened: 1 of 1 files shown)


robertkirkman:
I have compiled and tested this PR

I can confirm that this passes the test case given in the issue,
the command `printf '\x1b[=5u'`  no longer prints anything, whereas without this PR it prints `5u`.

Tested device: Samsung Galaxy S III SPH-L710, Android 7.1.2

robertkirkman (approved):

TomJo2000 (approved):

krobelus:
I'm not sure if it's easy for me to test but FWIW the changes look good.
There are some (contrived) cases missing like `CSI ! ! a`; fixing them would require a slightly larger change.

agnostic-apollo:
You can grab a build from https://github.com/termux/termux-app/actions/runs/13851563135?pr=4417 if you are using github releases instead of F-Droid.

There is no need to handle invalid sequences as there would be tonne/or infinite possibilities, the above was handled because it was valid, but unsupported by our terminal currently. The `!` is already captured by `ESC_CSI_EXCLAMATION`, but only `CSI ! p` is valid for a DECSTR reset, other parameters would result in error being logged if logging is enabled.

krobelus:
On Fri, Mar 14, 2025 at 01:55:33AM -0700, agnostic-apollo wrote:
> agnostic-apollo left a comment (termux/termux-app#4417)
> 
> You can grab a build from https://github.com/termux/termux-app/actions/runs/13851563135?pr=4417 if you are using github releases instead of F-Droid.

I don't have Android Studio or Android set up.

> There is no need to handle invalid sequences as there would be
> tonne/or infinite possibilities, the above was handled because
> it was valid, but unsupported by our terminal currently. The `!`
> is already captured by `ESC_CSI_EXCLAMATION`, but only `CSI ! p`
> is valid for a DECSTR reset, other parameters would result in error
> being logged if logging is enabled.

`CSI ! ! a` is a syntactically valid CSI command, we should consume the whole thing.

agnostic-apollo:
>`CSI ! ! a` is a syntactically valid CSI command, we should consume the whole thing.

Under what spec is it valid and what operation is supposed to be performed?

>I don't have Android Studio or Android set up.

Github action already builds the APKs for you, download the attached zip, extract the apk and install.

https://github.com/termux/termux-app#github

robertkirkman:
I have tested the newer version. For me it is still working the same (`printf '\x1b[=5u'` is not printing anything)

krobelus:
On Fri, Mar 14, 2025 at 02:30:51AM -0700, agnostic-apollo wrote:
> agnostic-apollo left a comment (termux/termux-app#4417)
> 
> >`CSI ! ! a` is a syntactically valid CSI command, we should consume the whole thing.
> 
> Under what spec is it valid and what operation is supposed to be performed?

Given
1. ECMA-48 does not limit the number of intermediate bytes
2. the combination of intermediate bytes and final byte identifies the control function
3. Regarding the final byte, it says: `Bit  combinations 07/00 to 07/14 are available as Final Bytes of control sequences for private (or experimental) use.`

it seems to me that in theory someone might legally use `CSI ! ! 0x70` (aka `CSI ! ! p`) for a private-use command
(and private-use ones might become de-facto standard.. I think fixterm's `CSI u` is an example but that might be inaccurate)

agnostic-apollo:
>I have tested the newer version. For me it is still working the same (printf '\x1b[=5u' is not printing anything)

I already tested that part, checking fish shell, or at least no apparent issue with text editors is what would need testing.

robertkirkman:
oh ok, now I have tested fish with https://github.com/termux/termux-packages/pull/23686 reverted, vim, and vim inside fish with https://github.com/termux/termux-packages/pull/23686 reverted, and nothing seems to be wrong. 
I have previously seen the basic symptoms of fish with https://github.com/termux/termux-packages/pull/23686 reverted inside the current official Termux APK releases, so I can tell that is no longer a problem, but I am not a heavy user of fish or highly experienced with it, so if fish relies on subtler behaviors for some operations that aren't easy to encounter during normal use, I might not have been able to check those.

agnostic-apollo:
The only supported byte after a `!` is a `p` as per spec. A double `!!` is not a valid command. Additionally, even `!` is not part of original ECMA-48 standard, as it only allows bit combinations from 03/00 to 03/15, which are `0–9:;<=>?`. Private modes would require starting the sequence with one of the officially supported bit combinations, or publishing a spec.

>CSI ! p   Soft terminal reset (DECSTR), VT220 and up.

- https://invisible-island.net/xterm/ctlseqs/ctlseqs.html
- https://terminalguide.namepad.de/seq/
- https://vt100.net/docs/vt220-rm/chapter2.html#S2.4.1

agnostic-apollo:
Thanks for testing @robertkirkman. That should probably be enough.

krobelus:
I agree that it's enough testing for this fish version. Every new CSI related command is sent at fish startup so errors should be noticeable.

Thanks everyone for your work. We've had a bunch of reports about this since late December (from the fish beta release); I'll try to act faster next time.

---

Now on to getting nerrd-sniped:

I'm not thinking of any commands that are in use today.
I agree that the ! byte is never a parameter byte but it's an intermediate byte (of which there may be multiple according to
[ECMA-48](https://www.ecma-international.org/wp-content/uploads/ECMA-48_2nd_edition_august_1979.pdf)).

I'm also not talking about private mode.  Only about commands that
are possibly allowed by ECMA-48, either by future versions of a
standard or (more likely) by other protocols. Since I don't know how
to read "are available for private or experimental use" in section 5.4 d)
of ECMA-48, I do not know if the standard allows the latter outside
private mode.

To me it looks like a complete CSI parser should parse any (unknown) "private or experimental use" commands.
I read most of section 5.4 but I may easily be wrong in my interpretation.

agnostic-apollo:
> I agree that the ! byte is never a parameter byte but it's an intermediate byte.

But an intermediate byte must always be proceeded by a valid parameter byte, which `!` is not, at least as per ECMA-48. The VT200 series did have `CSI ! p`, but no other command exists for any standard or spec that I have seen where `!` is a parameter byte. If there is one, let me know, and we can support it.

- https://ecma-international.org/publications-and-standards/standards/ecma-48/

> To me it looks like a complete CSI parser should parse any (unknown) "private or experimental use" commands.

This pull already started parsing all private and experimental ranges and ignoring them. Any other character outside the supported ranges should not be used and should be an error, until someone publishes a standard. There is enough range left for private/experimental use left by the standard for someone to have to go outside it.

> The unallocated bit combinations are reserved for future standardization and shall not be used.

> If the first bit combination of the parameter string is in the range 03/12 to 03/15, the parameter string
is available for private (or experimental) use. Its format and meaning are not defined by this
Standard.

> The number of Intermediate Bytes is not limited by this Standard; in practice, one Intermediate Byte will be
sufficient since with sixteen different bit combinations available for the Intermediate Byte over one
thousand control functions may be identified.

> Bit combinations 07/00 to 07/14 are available as Final Bytes of control sequences for private (or
experimental) use.

- https://ecma-international.org/wp-content/uploads/ECMA-48_5th_edition_june_1991.pdf

Check the code table in following link and compare the `<column>/<row>` combinations to get the ranges.

- - https://vt100.net/docs/vt220-rm/chapter2.html#S2.4.1

Additionally, the 5.1.2 section of 3rd and 4.1.2 section of 2nd editors of the standard had an explicit error condition statement for using unsupported combinations, which seems to have gotten removed in later editions for some reason.

> The occurrence of any bit combinations which do not conform to the above format is an error condition for which recovery is not specified by this Standard.

- https://ecma-international.org/wp-content/uploads/ECMA-48_3rd_edition_march_1984.pdf

agnostic-apollo:
You are welcome and thanks again for the pull.

PR #4473, opened 2025-03-31
Fix outdated branch #7

benja2998:
This PR fixes the README in modifiers-pageup-pagedown


TomJo2000:
Development branches are not usually updated manually.
If and when they need to be updated to a newer state of `master` that is usually done via a `git rebase`.

PR #4480, opened 2025-04-06
Usable Termux project for latest Android Studio + Compilable and usable same as old one

Comical78:
Usable Termux project for latest Android Studio + Compilable and usable same as old one, this comits will let the project being able to be updated even later on. This may be uselfull.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -20,6 +20,7 @@
         android:label="@string/permission_run_command_label"
         android:protectionLevel="dangerous" />
 
+    <!-- Expected or not, this pemissions: DUMP, READ_LOGS and WRITE_SECURE_SETTINGS are limited -->
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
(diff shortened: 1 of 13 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -31,7 +32,7 @@
     <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
     <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" />
     <uses-permission android:name="android.permission.READ_LOGS" />
-    <uses-permission android:name="android.permission.DUMP" />
+    <uses-permission android:name="android.permission.DUMP" />
```
benja2998, line 35: Definitely the most important change

Comical78: Definetly is not. But I didn't changed it at all just added a notice about permissions may broke, so yeah.

benja2998 (commented):
Changing dependencies may break the app.

Comical78:
Sir, I have tested the app already on a Xiaomi Android 15 device, it worked smooth and same as old version but that time the application color of the logo worked since it existed configs about it. I tryed to change the andorid sdk versions but it broke so I let them as it is.

benja2998:
> Sir, I have tested the app already on a Xiaomi Android 15 device, it worked smooth and same as old version but that time the application color of the logo worked since it existed configs about it. I tryed to change the andorid sdk versions but it broke so I let them as it is.

Older Android versions may not have the new versions of these dependencies. Use the Android Stuido emulator to test older Androd versions.

termux-shared/build.gradle
```diff
@@ -71,7 +79,8 @@ dependencies {
 
 task sourceJar(type: Jar) {
     from android.sourceSets.main.java.srcDirs
-    classifier "sources"
+    //classifier "sources" //depracated
```
twaik, line 82:
```suggestion
    // classifier parameter is deprecated
``

agnostic-apollo:
This won't be merged so no need to spend time on it. These issues have already been fixed locally and will be available in coming days.

The only thing that's needed for latest android studio version to build termux-app is using jdk 11 or 17 instead of default 21.

PR #4554, opened 2025-05-21
Update HiddenApiBypass to 6.1

aviraxp:
This should fix crash on A16 QPR1.

### DIFF
--- a/termux-shared/build.gradle
+++ b/termux-shared/build.gradle
@@ -14,7 +14,7 @@ android {
         implementation "io.noties.markwon:ext-strikethrough:$markwonVersion"
         implementation "io.noties.markwon:linkify:$markwonVersion"
         implementation "io.noties.markwon:recycler:$markwonVersion"
-        implementation "org.lsposed.hiddenapibypass:hiddenapibypass:5.0"
+        implementation "org.lsposed.hiddenapibypass:hiddenapibypass:6.1"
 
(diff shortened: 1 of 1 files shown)


agnostic-apollo:
What crash? A16 crash was already fixed with https://github.com/termux/termux-app/commit/d2cd6ac2e579869b4fa1b1d2737c434e64c20fb9 unless there is anew one.

agnostic-apollo:
Alas, I guess fixed in https://github.com/LSPosed/AndroidHiddenApiBypass/commit/9efadf062d0feb1dbb8ad91e3d7a8d71a143ec46, can you post the stacktrace for termux. I will also have to review the library changes since last update before I can bump the version here. And then new releases would have to be made too.

agnostic-apollo:
Have you tested that the crash is fixed by bumping the library version? I don't have a QPR device and latest AVD release is older, so I can't test and need a confirmation before I do the release chores of multiple apps.

If not, then can you or someone else test the build from https://github.com/termux/termux-app/actions/runs/15184167555?pr=4554 if using GitHub releases.

---

The new library version has added another mechanism to bypass hidden API restrictions, but we will stick to older way.

- https://github.com/LSPosed/AndroidHiddenApiBypass/commit/0df64f21c9bf728ea811a5bdd9ed634afa42e2c4
- https://github.com/LSPosed/AndroidHiddenApiBypass/commit/d5e236779d234736e5bbe33ee59609ded7cafe49

PR #4557, opened 2025-05-22
Replace text "Latest version" with auto-updating badge in README.md

Dvd-Znf:
Latest version in README.md is outdated again because of the newly released version.  

It might be nicer to have an auto-updating badge instead of manually changing the text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ The core [Termux](https://github.com/termux/termux-app) app comes with the follo
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


agnostic-apollo:
Fix commit message.

agnostic-apollo:
And only bump version. Docs must be text readable.

Dvd-Znf:
> And only bump version. Docs must be text readable.

Fair,
Reopened in #4558 to fix meaningful commit message

PR #4558, opened 2025-05-22
Changed README.md: Set latest version to v0.118.3

Dvd-Znf:
Bumped README.md version to new 0.118.3 release

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ The core [Termux](https://github.com/termux/termux-app) app comes with the follo
 
 ## Installation
 
-Latest version is `v0.118.2`.
+Latest version is `v0.118.3`.
 
 **NOTICE: It is highly recommended that you update to `v0.118.0` or higher ASAP for various bug fixes, including a critical world-readable vulnerability reported [here](https://termux.github.io/general/2022/02/15/termux-apps-vulnerability-disclosures.html). See [below](#google-play-store-experimental-branch) for information regarding Termux on Google Play.**
 


agnostic-apollo:
Set commit message to `Changed(README): Set latest version to 0.118.3`. Don't send another pull.

README.md
```diff
@@ -49,7 +49,7 @@ The core [Termux](https://github.com/termux/termux-app) app comes with the follo
 
 ## Installation
 
-Latest version is `v0.118.2`.
+Latest version is `v0.118.3`.
```
hannesa2, line 52: To be honest, I would delete this line complete, because everyone can read under **[releases](https://github.com/termux/termux-app/releases)**, what's he current version is.
This would simplify a new release.

agnostic-apollo:
Closed by https://github.com/termux/termux-app/commit/c6e2ca7fde421d97e8f2f4990cd443d621007a02

README.md
agnostic-apollo: Beta releases are marked as a pre-release and are not the final release version and they also show up on releases page, so users should know which to install. I have rewritten the README to specify both and the differences, but some of the docs are not complete yet. For the termux site in future, something like https://github.com/termux/termux-core-package/blob/master/site/pages/en/projects/releases/index.md will be done too.

The reason this issue is occurring is because temporarily releases are being made from separate branches instead of `master` branch, where the bump in the `README` would normally get done along with in `build.gradle`, but that will be resolved soon as I am working on finalizing most of my local changes to push them, and next beta should likely be from `master`, unless fixup betas needs to be made before it.

Issue #4569, opened 2025-05-29
[Bug]: build on arm64


hannesa2:
### Problem description

Hi,
I try to build it by my own on my Apple Silicon with `./gradlew build`

<img width="1175" alt="Image" src="https://github.com/user-attachments/assets/2b240999-c77e-4ff5-820f-2e77ad1f1f2c" />

and it ends up with

> [CXX1429] error when building with ndkBuild using /Users/hannes/git/termux-app/terminal-emulator/src/main/jni/Android.mk: ERROR: Unknown host CPU architecture: arm64

I see there is nor [arm64](https://github.com/termux/termux-app/blob/master/app/build.gradle#L215-L227) at all.
Any chance to make this work, or what I've to do ?

### Steps to reproduce the behavior.

I build it by my own on my Apple Silicon with `./gradlew build`

<img width="1175" alt="Image" src="https://github.com/user-attachments/assets/2b240999-c77e-4ff5-820f-2e77ad1f1f2c" />

and it ends up with

> [CXX1429] error when building with ndkBuild using /Users/hannes/git/termux-app/terminal-emulator/src/main/jni/Android.mk: ERROR: Unknown host CPU architecture: arm64

I see there is nor [arm64](https://github.com/termux/termux-app/blob/master/app/build.gradle#L215-L227) at all.
Any chance to make this work, or what I've to do ?

### System information

* Termux application version:
* Android OS version:
* Device model:

hannesa2:
Same here 

<img width="784" alt="Image" src="https://github.com/user-attachments/assets/291ce57e-8df5-4a59-aeca-6b4a2a2a828c" />

![Image](https://github.com/user-attachments/assets/bed59afc-1f68-4ced-ba8b-adc382160898)

agnostic-apollo:
That is still using `21.4*/ndk-build`. Possibly run clean first, and sync gradle files.

hannesa2:
even after
```
git clean -x -d -f
./gradlew clean build
```
error still exists.

Westokid204:
pin yang nmesahau

hannesa2:
Hmmm, on top I see 

<img width="1120" alt="Image" src="https://github.com/user-attachments/assets/ca7eb554-9b61-43aa-9ea5-284322b67914" />

later in build this 

<img width="1047" alt="Image" src="https://github.com/user-attachments/assets/be2a3cd4-bb61-4c4e-9f92-cff55b516bce" />

[...]

PR #4571, opened 2025-05-30
Build on arm64 with github action

hannesa2:
If points out build issues on MacOS arm64  #4569

### DIFF
--- a/.github/workflows/debug_build.yml
+++ b/.github/workflows/debug_build.yml
@@ -76,7 +76,7 @@ jobs:
           echo "APK_VERSION_TAG=$APK_VERSION_TAG" >> $GITHUB_ENV
           echo "APK_BASENAME_PREFIX=$APK_BASENAME_PREFIX" >> $GITHUB_ENV
 
-          echo "Building APKs for 'APK_VERSION_TAG' build"
+          echo "Building APKs for $APK_VERSION_TAG build"
           export TERMUX_APP_VERSION_NAME="${RELEASE_VERSION_NAME/v/}" # Used by app/build.gradle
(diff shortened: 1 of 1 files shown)


agnostic-apollo:
I had already pushed commits to fix the runner environment to build. macos-14 doesn't have sha256sum, that's why I also deleted it after testing build works in it too, no specific need to test it, as not an issue with termux app build config, but runner environment.

Also FYI, this pull won't be merged, at least until I push to master as workflow file doesn't even exist locally and no immediate need to test macos builds. Will have to disable artifact upload for it too if it were added to prevent duplicate artifacts to be uploaded for each os.

hannesa2:
> and no immediate need to test macos builds

There is a need !
Btw, the over engineered and error prone versionName doesn't work on MacOS. 
Why not simply use `git describe --tags` like [this](https://github.com/hannesa2/paho.mqtt.android/blob/master/extendedSample/build.gradle.kts#L19-L20)

agnostic-apollo:
Firstly, commit hashes are only added for github action builds, and not normal local or release builds. And users build locally as well with specific versioning schemes, so need to support that.

Secondly, fdroid requires hardcoded values.

https://github.com/termux/termux-app/commit/0aa5a123b7adfe4fc01d76d60ce47ed20d382658

Thirdly, it does work on macos 14 and 15 as I already tested in the workflow run referred above as per workflow file in https://github.com/termux/termux-app/blob/d435ee2fdc58e1c6554657ead2c72fc315cb9275/.github/workflows/debug_build.yml. You overwrote my commits, I can push them again if that's what you want.

hannesa2:
Don't worry, I just want to see, if d435ee2fdc58e1c6554657ead2c72fc315cb9275 works

PR #4743, opened 2025-09-02
Update TermuxActivityTest.java

veo3-0:


agnostic-apollo:
Those tests are meant to test the `TermuxUrlUtils.extractUrls()` function that's actually used by `TermuxActivity`, and you have removed it, what are you even doing.

PR #4744, opened 2025-09-02
Update UrlUtils.java

veo3-0:


veo3-0:
This PR refactors the existing `UrlUtils` class to improve reliability, consistency, and extend its capabilities. Key enhancements and additions include:

---

###  Changes Made

1. **Switched from `java.net.URL` to `java.net.URI`:**

   * Improves compatibility and better handles edge cases in URL parsing and joining.
   * Replaced all `URL` usage with `URI` to avoid unnecessary exceptions (`MalformedURLException` → `URISyntaxException` is more granular and flexible).

2. **Enhanced `joinUrl(...)` Method:**

   * Uses `URI.resolve(...)` for better accuracy when resolving relative URLs.
   * Improves error handling with cleaner fallback logic.

3. **Added `extractUrls(String text)` Method:**

   * Uses regex to scan and extract URLs from any arbitrary string content.
   * Returns a `LinkedHashSet<String>` to maintain order and uniqueness.

4. **Improved `areUrlsEqual(...)`:**

   * Normalizes both URLs by:

     * Removing protocol and `www.`
     * Stripping trailing slashes
     * Converting to lowercase for case-insensitive comparison

5. **Simplified and Enhanced `removeProtocol(...)`:**

   * Now strips protocol (`http://`, `https://`) and optional `www.` prefix more cleanly.

6. **Improved Documentation:**

   * Javadoc-style comments added for all methods and enum values.
   * Clearer intent and usage for each method.

PR #4806, opened 2025-10-06
Delete .github/FUNDING.yml

andreanadr:


robertkirkman:
This is already done in this commit: https://github.com/termux/termux-app/commit/0b4f4561323616b4b99867e6b0ebcccf94c47f69

however, unfortunately the structure of the branches is a bit confusing, so that commit hasn't been applied to all the branches yet.

PR #4931, opened 2026-01-08
Enable RTL support in AndroidManifest.xml

ziadaburas:


ziadaburas:
add RTL support

mgnegypt:
**I strongly support adding RTL to the project**

sylirre:
`android:supportsRtl="true"` won't enable RTL support in terminal. This will affect only UI elements. However as app doesn't use any translations and doesn't use right-to-left layouts, this change doesn't make sense and won't be merged alone.

Terminal renderer would require a significant overhaul for adding a true RTL support, not a mere flip of `android:supportsRtl`.

PR #4962, opened 2026-02-11
feat: auto hide extra keys & detect keyboard changes

ozwaldorf:
Adds a new toggle to the settings to only show extra keys when no hardware keyboard is present.

Also allows receiving updates from android system when a hardware keyboard is enabled/disabled.
This enables the soft keys settings to apply right away, without having to reload the activity.

IMHO it should be enabled by default, but for now left it as disabled.


ozwaldorf:
rebased onto latest master

Issue #4957, opened 2026-02-07
[Bug]: Termux refuses to open / download files.


TheEpicDoge17:
### Problem description

When trying to open a file (Example: object.json) through Termux, the "Save file to ~/downloads" screen appears as intended, but pressing "Edit" gives the error "The following file does not exist; $HOME/bin/termux-file-editor" even though the directory exists.

https://github.com/user-attachments/assets/a67e1ad0-2a75-43e0-b930-16561ce2972b

### Steps to reproduce the behavior.

Depending on the file editor you use, the steps to replicate this may differ, but should have the same process. For this guide, I used the built-in file manager for Android.

1. If you have not done so already, create the folder "termux-file-editor" within $HOME/bin using any method.
2. Open an external (not Termux) file browser.
3. Select the file you would like to edit. (I used "example.class", but also works for other files too.)
2b. If necessary, select Termux as the program used to open the file.
3. When (or if) the TermuxFileReciever window appears, name the file how you so choose and select "Edit".
4. Termux fails to edit the file under the error "The following file does not exist; $HOME/bin/termux-file-editor"

### What is the expected behavior?

Termux copies the chosen file to $HOME/downloads and opens the file.

### System information

* Termux application version: 0.199.0-beta.3
* Android OS version: 6.6.77-android15-8-31998796-abogkiS938USQS7BYJ9-4k`  
* Device model: SM-S938U

robertkirkman:
Hello, please try these commands in order to resolve your problem:

```bash
rmdir $HOME/bin/termux-file-editor
mkdir -p $HOME/bin
ln -s $PREFIX/bin/editor $HOME/bin/termux-file-editor
```

wqadir594-coder:
Please viral my tiktok

mustafahaidar865725-ux:
rmdir $HOME/bin/termux-file-editor
mkdir -p $HOME/bin
ln -s $PREFIX/bin/editor $HOME/bin/termux-file-editor

PR #4972, opened 2026-02-16
file-receiver: add fallback editor when termux-file-editor is not configured

HemanthArjun09:
This commit addresses issue #4957 where opening shared files in Termux
fails if the user has not manually configured the `termux-file-editor`
script.

Problem:
FileReceiverActivity currently assumes the presence of
$HOME/bin/termux-file-editor when handling shared files. If the script
is missing, the application displays an error dialog and aborts the
operation.

For new or default installations, this results in file sharing not
working out-of-the-box, even when common editors like nano or vim are
already installed. Users are required to manually create a script or
symlink before basic file-opening functionality works, which creates
confusion and unnecessary setup friction.

Solution:
Introduce a fallback editor detection mechanism when
`termux-file-editor` is not found:

- First attempt to execute the existing `termux-file-editor` script
  (current behavior preserved).
- If the script is missing, automatically search for commonly
  available editors (`nano`, `vim`, `vi`) in `$PREFIX/bin`.
- Launch the first executable editor found.
- If no editor is available, display a clear installation hint
  instead of a generic failure message.

Permissions are only adjusted for user-provided scripts to avoid
modifying packaged binaries.

Impact:
- Enables file sharing to work immediately on fresh installations.
- Preserves backward compatibility for users with custom editor setups.
- Improves usability without changing execution flow or APIs.
- Aligns behavior with typical terminal environments where a default
  editor is automatically used if configured scripts are absent.
- Reduces user confusion and setup requirements.

Fixes: #4957

### DIFF
--- a/app/src/main/java/com/termux/app/api/file/FileReceiverActivity.java
+++ b/app/src/main/java/com/termux/app/api/file/FileReceiverActivity.java
@@ -166,25 +166,29 @@ public class FileReceiverActivity extends AppCompatActivity {
                 if (outFile == null) return;
 
                 String editorExecutable = EDITOR_PROGRAM;
-
                 File editorProgramFile = new File(editorExecutable);
 
                 if (!editorProgramFile.isFile()) {
-                    editorExecutable = findDefaultEditor();
-                }
+                    editorExecutable = findDefaultEditor(); // Fallback to common editors if termux-file-editor is not configured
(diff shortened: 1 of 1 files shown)


robertkirkman:
This problem is going to be fixed in a future version using code that is already made but which is not yet public.

PR #4977, opened 2026-02-18
Increase read buffer size from 4KB to 64KB

### LINKED ISSUE #4976: [Bug]: Emulator lags in unusable fashion with large output from terminal multiplexers
### Problem description

Currently when using Termux with large content (e.g. large log tailing, `chafa` output), it's _already_ slow to stream. For example, you'd see `chafa` images being streamed almost row by row. But once streamed, it's usable as rows streamed are buffered. Scrolling works fine.

But when using a terminal multiplexer like `zellij` or `tmux`, content are delivered as screen updates, which involve lots of control sequences. This is the case for both the initial streaming and subsequent scrolling.

The end result is extremely laggy. Attempting to scroll on a high-end phone (small screen at < 100 width + fast chip) gives you something less than 1 FPS. It's painfully slow and essentially unusable on these content.

1. Create a folder with a bunch of decently sized images.
2. Install `chafa` and a multiplexer of your choice (e.g. `zellij`)
3. Enter a multiplexer session and go to the folder.
4. Execute `chafa *`.
5. Try to scroll.
6. The terminal emulator is extremely laggy at probably lower than 1 FPS.

Terminal emulator should be reasonably smooth.

### System information

* Termux application version: e50ac9e019f5dd359b252bc4312f25de7b7dc8ed
* Android OS version: 16
* Device model: Samsung Galaxy S25 Ultra

xJonathanLEI:
Fixes #4976.

I've tested very large content streaming scenarios (`chafa`-ing dozens of large images). `master` lags to a painfully unusable state where the terminal emulator remains unresponsive for up to minutes.

With this patch it's very smooth and scrolling in the multiplexer also works perfectly. 64KB seems a much more sensible size than 4KB it appears.

### DIFF
--- a/terminal-emulator/src/main/java/com/termux/terminal/TerminalSession.java
+++ b/terminal-emulator/src/main/java/com/termux/terminal/TerminalSession.java
@@ -41,7 +41,7 @@ public final class TerminalSession extends TerminalOutput {
      * A queue written to from a separate thread when the process outputs, and read by main thread to process by
      * terminal emulator.
      */
-    final ByteQueue mProcessToTerminalIOQueue = new ByteQueue(4096);
+    final ByteQueue mProcessToTerminalIOQueue = new ByteQueue(64 * 1024);
     /**
      * A queue written to from the main thread due to user interaction, and read by another thread which forwards by
      * writing to the {@link #mTerminalFileDescriptor}.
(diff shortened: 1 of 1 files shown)


fornwall:
Makes sense to increase the buffer size - 4 KB is quite small. Thanks!

PR #4979, opened 2026-02-20
fake root container 

0m364:
fake root container

### DIFF
--- a/.github/workflows/attach_debug_apks_to_release.yml
+++ b/.github/workflows/attach_debug_apks_to_release.yml
@@ -17,7 +17,7 @@ jobs:
 
     steps:
       - name: Clone repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
         with:
           ref: ${{ env.GITHUB_REF }}
 
--- a/.github/workflows/debug_build.yml
+++ b/.github/workflows/debug_build.yml
@@ -19,10 +19,10 @@ jobs:
 
     steps:
       - name: Clone repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
 
       - name: Setup java 17
-        uses: actions/setup-java@v5
+        uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
           java-version: '17'
@@ -85,7 +85,7 @@ jobs:
           fi
 
       - name: Attach universal APK file
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v4
         with:
           name: ${{ env.APK_BASENAME_PREFIX }}_universal
           path: |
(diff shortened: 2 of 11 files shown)


sylirre:
Could you give some practical examples where fake root is useful?

0m364:
installing packages in termux that request root access without need for proot or opening the android system via root ...some tools request this

sylirre:
Give practical examples. What tools? What do you meaning by *opening the android system via root*?

Features added to the app must be useful for most users. It shouldn't be something niche, like bypass root check in specific script from GitHub. It is not hard to type `proot -0 bash` on the command line when user really needs it.

So I would like to know what we miss without Fake Root session.

0m364:
> Give practical examples. What tools? What do you meaning by _opening the android system via root_?
> 
> Features added to the app must be useful for most users. It shouldn't be something niche, like bypass root check in specific script from GitHub. It is not hard to type `proot -0 bash` on the command line when user really needs it.
> 
> So I would like to know what we miss without Fake Root session.

i personally use it for running agent based local llms like opencode directly in termux ... it would also allow users who build lfs to do so in their phone without the lag of nested vm .. would also open up to testing shells ... i meant rooting/jailbreaking the system

sylirre:
> i personally use it for running agent based local llms like opencode

Local llms never required fake or real root permissions.

> it would also allow users who build lfs to do so in their phone without the lag of nested vm

LFS (if you meant Linux-from-scratch) would require a full fledged proot setup. Using `proot -0 <shell>` alone won't help because it only impersonates current user as root.

Furthermore, usage of `proot` would actually make performance worse.

> would also open up to testing shells ... i meant rooting/jailbreaking the system

It doesn't help with rooting or jailbreaking the system because as said previously, `proot` only impersonates current user and that has effect only for proot child processes. The kernel still sees the original user name because proot can't hijack value in its memory space. No process permission attributes being changed.

0m364:
> > i personally use it for running agent based local llms like opencode
> 
> Local llms never required fake or real root permissions.
> 
> > it would also allow users who build lfs to do so in their phone without the lag of nested vm
> 
> LFS (if you meant Linux-from-scratch) would require a full fledged proot setup. Using `proot -0 <shell>` alone won't help because it only impersonates current user as root.
> 
> Furthermore, usage of `proot` would actually make performance worse.
> 
> > would also open up to testing shells ... i meant rooting/jailbreaking the system
> 
> It doesn't help with rooting or jailbreaking the system because as said previously, `proot` only impersonates current user and that has effect only for proot child processes. The kernel still sees the original user name because proot can't hijack value in its memory space. No process permission attributes being changed.

yes.... i know .. this avoids the latency with proot... llm chatbots do not require root ... but agents and sub agents ... unless fine tuned for termux .. will refuse what they see as a sudo command ... the fix for this is to have them create an "askpass tool" on the fake container you make... this allows the agent to control a sub agent on standard termux .. allowing control of termux api controlled actions with natural language

DeezDevelopes:
Yes

Denny Donson

PR #4998, opened 2026-03-04
improve ui

NIK2703:
Screenshot_2026-03-04-13-00-36-948_com termux

- Terminal tabs have been moved from the side panel to the bottom, positioned above the additional keys bar;
- The text input field for swipe typing has been relocated from the additional panel (accessed via left swipe) to the area above the tabs panel (hideable);
- Added background styling for keys on the additional keys panel.

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -548,6 +548,11 @@ public final class TermuxActivity extends AppCompatActivity implements ServiceCo
 
         setTerminalToolbarHeight();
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -4,7 +4,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:8.13.2"
+        classpath "com.android.tools.build:gradle:8.5.2"
     }
 }
```
twaik: No.

twaik:
Seems like the whole PR is LLM product and it was not properly reviewed by human before publishing. It adds and removes shitton of stuff without it being properly discussed or even motivated.

PR #5007, opened 2026-03-10
Fix fullscreen mode on Android 11+ where navigation bar was not hidden

wintrics:
On Android 11 and newer versions, the navigation bar remained visible in fullscreen mode.


mio-19:
How do I enter fullscreen mode?

sylirre:
@mio-19 By editing `~/.termux/termux.properties` file. Uncomment the line with `fullscreen = true`.

PR #5082, opened 2026-04-18
Fix background character overlay bug and add bell-based task completion notifications

yanmengdie:
## Summary

- **Bug fix (P1) — Background character overlay:** Removed the `isVisible()` guard from `onTextChanged()` in `TermuxTerminalSessionActivityClient`. Previously, the terminal view was not updated while the app was backgrounded, causing accumulated buffer changes to visually overlay already-rendered text when the user returned to the foreground. The fix ensures the terminal view state is always kept in sync regardless of visibility.

- **Feature (P2) — Bell-based task completion notifications:** Added two new `bell-character` property values for `~/.termux/termux.properties`:
  - `notification` — fires an Android notification when a bell character (``) is received, works in foreground and background.
  - `vibrate-and-notification` — vibrates (via `BellHandler`) when the app is visible, and sends a notification (with vibration) when backgrounded, avoiding double-vibration.

  This enables SSH-based remote workflow notifications: append `echo -e ''` to a script on the remote server and Termux will notify you when it finishes, even with the screen off.

  Example `termux.properties` config:
  ```
  bell-character = notification
  # or
  bell-character = vibrate-and-notification
  ```

- **Bug fix (P3) — TerminalBuffer resize wide-char indexing:** Fixed a pre-existing bug in the terminal buffer resize/reflow logic where `mStyle` was indexed using character (char array) indices instead of column indices, causing incorrect style application for wide characters (CJK) and surrogate pairs.

## Files changed

- `terminal-emulator/.../TerminalBuffer.java` — fix wide-char column index bug in resize logic
- `termux-shared/.../TermuxPropertyConstants.java` — add `IVALUE_BELL_BEHAVIOUR_NOTIFICATION` (4) and `IVALUE_BELL_BEHAVIOUR_VIBRATE_AND_NOTIFICATION` (5) constants and BiMap entries
- `app/.../TermuxTerminalSessionActivityClient.java` — remove `isVisible()` guard from `onTextChanged()`; add `sendBellNotification()` helper; update `onBell()` to handle new notification modes
- `app/src/main/res/values/strings.xml` — add `notification_bell_title` and `notification_bell_text` string resources

## Test plan

- [ ] Background the app, run a long output command, return to foreground — no character overlay
- [ ] Set `bell-character = notification`, run `echo -e ''` while foregrounded → notification appears
- [ ] Same while app is backgrounded → notification still appears
- [ ] Set `bell-character = vibrate-and-notification`, background app, run `echo -e ''` → single vibration + notification (not double vibration)
- [ ] SSH to a remote server, append `echo -e ''` to a long-running command, confirm notification arrives when command finishes
- [ ] Resize terminal with CJK/wide characters — no crash or style corruption

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -295,6 +295,13 @@ public final class TermuxActivity extends AppCompatActivity implements ServiceCo
         if (mTermuxTerminalViewClient != null)
             mTermuxTerminalViewClient.onStart();
 
+        // BUGFIX P1: Force synchronization of screen when returning from background
+        // This ensures that any background terminal updates are properly displayed
+        if (mTerminalView != null && mTerminalView.mEmulator != null) {
+            mTerminalView.onScreenUpdated();
+            mTerminalView.setTopRow(0); // Ensure scroll position is synchronized
+        }
+
         if (mPreferences.isTerminalMarginAdjustmentEnabled())
             addTermuxActivityRootViewGlobalLayoutListener();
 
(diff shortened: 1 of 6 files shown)


ghost (commented):
I love you

PR #5084, opened 2026-04-19
Add toggle terminal toolbar intent

### LINKED ISSUE #5083: [Feature]: Make terminal toolbar toggle scriptable
### Feature description

Currently, the only ways to toggle the visibility of the terminal toolbar appear to just be:

- Long-pressing the keyboard button
- Volume + `q`
- Volume + `k`

all of which involve physically operating the device. There also seems to be no way to programmatically toggle this setting.

As a long time Termux user I find myself using the app in two modes:

1. Handheld mode: using the toolbar + virtual keyboard
2. Desktop mode: using the device only as a screen; w/ toolbar and virtual keyboard disabled - use external physical keyboard

When switching from mode 1 to mode 2 by turning on an external keyboard, most systems are smart enough to automatically hide the virtual keyboard. But the terminal toolbar itself remains visible. On small devices every pixel counts and the toolbar takes up valuable screen real estate. So I'd end up having to reach for the phone again to turn it off.

A simple fix would be to just allow programmatic access to toggling the terminal toolbar so that it can be set as an alias in the shell.

### Additional information

N/A

xJonathanLEI:
Resolves #5083.

### DIFF
--- a/app/src/main/java/com/termux/app/TermuxActivity.java
+++ b/app/src/main/java/com/termux/app/TermuxActivity.java
@@ -915,10 +915,17 @@ public final class TermuxActivity extends AppCompatActivity implements ServiceCo
         context.sendBroadcast(stylingIntent);
     }
 
+    public static void toggleTermuxTerminalToolbar(Context context) {
+        Intent toggleIntent = new Intent(TERMUX_ACTIVITY.ACTION_TOGGLE_TERMINAL_TOOLBAR);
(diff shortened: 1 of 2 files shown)


xJonathanLEI:
Tested. This command toggles the toolbar:

```
termux-am broadcast -a com.termux.app.toggle_terminal_toolbar -p com.termux
```

PR #5105, opened 2026-05-01
Proposal: Modernize multi-module Build System for Gradle 9+ compatibility

batoripX:
​Hi Termux Team,
​I’ve been trying to contribute a new Kotlin file to the project, but I’ve run into a significant barrier: the build system is heavily reliant on deprecated APIs that prevent building in modern environments (Gradle 9.4.1 / AGP 9.2.0).
​Specifically, I had to manually refactor every module (:app, :terminal-emulator, :terminal-view, :termux-shared) to:
​Remove/comment out the applicationVariants and libraryVariants scripts.
​Update ProGuard configurations to modern R8 syntax.
​Fix hidden syntax errors in build.gradle headers.
​The "Build Tax" for new contributors is currently very high because of this technical debt. I suggest a project-wide modernization so that new developers can focus on features rather than fighting the build system. I'm happy to share the errors I encountered if it helps!


agnostic-apollo:
There are no plans to rewrite termux app or terminal to kotlin, or removing application variants code that would break github actions and release builds. Gradle versions are already near latest and should build the app just fine without fighting.

PR #5213, opened 2026-07-26
feat: hide terminal toolbar when hardware keyboard is connected

### LINKED ISSUE #5199: [Feature] Hide terminal toolbar if there's a hardware keyboard
### Feature description

Add a toggle for hiding the terminal toolbar if there's a hardware keyboard in the Termux settings (OTG connected or Bluetooth keyboard).

It's preferred to keep the current implement as the default option like the "enable soft keyboard if there's no hardware keyboard" toggle.

### Additional information

It's annoying to not have this, especially with those who's using swipe gesture to navigate and had to precisely swipe so that you can open that left sidebar in order to get into the settings menu (that's me)

`termux-info` output: [termux-info_cli.log](https://github.com/user-attachments/files/30218673/termux-info_cli.log)

sujithmolakala:
Adds an optional setting to hide the terminal toolbar when a hardware keyboard is connected. The setting is disabled by default to preserve the existing behavior. When enabled, Termux uses the existing hardware keyboard detection to hide the terminal toolbar and restores its visibility when appropriate.

Closes #5199

Test plan

- `./gradlew assembleDebug`
- Verified the new setting appears under Settings > Termux > Terminal I/O
- Verified the setting is disabled by default
- Verified the terminal toolbar remains visible when the setting is disabled
- Verified the terminal toolbar is hidden when the setting is enabled and a hardware keyboard is detected


132ikl:
would love to see this merged ❤️

PR #5222, opened 2026-07-30
Update and rename README.md to pkg update && pkg upgrade -y

hesamazadbakht007-ux:
pkg install python git -y


hesamazadbakht007-ux:
python --version

PR #5223, opened 2026-07-31
Added|Fixed: Abort bootstrap install if device SDK incompatible with …

HeroRickyGAMES:
…APK bootstrap variant and run bootstrap second stage

- Set `minSdkVersion` and `maxSdkVersion` depending on the APK bootstrap variant and abort bootstrap installation at runtime with a clear error message if the device Android SDK version is not compatible with the APK bootstrap variant. Android does not enforce `maxSdkVersion`, so installing an APK of a variant not supporting the device Android version (like `apt-android-5` on Android >= 7) would crash later on instead of showing a proper error.
- Run `termux-bootstrap-second-stage.sh` after bootstrap extraction since newer bootstrap packages require it to configure the bootstrap packages, otherwise `apt`/`dpkg` may fail during package installation.


robertkirkman:
> installing an APK of a variant not supporting the device Android version (like apt-android-5 on Android >= 7) would crash later on instead of showing a proper error.

The current error message looks pretty normal... is your error message much better than this one?

image

robertkirkman:
> Run termux-bootstrap-second-stage.sh after bootstrap extraction since newer bootstrap packages require it to configure the bootstrap packages, otherwise apt/dpkg may fail during package installation.

For me master branch (built manually) doesn't have that problem... are you sure this PR is fixing something?

Screenshot_20260804-041825_two termux

Issue #5229, opened 2026-08-03
[Bug]: Local TerminalView input becomes severely delayed while SSH remains responsive on Android 16 / ColorOS


zhaodong1982:
# [Bug]: Local TerminalView input becomes severely delayed while SSH remains responsive on Android 16 / ColorOS

### Problem description

After the Termux UI has been running for some time, character input in the local `TerminalView` becomes extremely delayed and can become almost unusable. The shell and PTY are still responsive: connecting to the same Termux environment through SSH on localhost works normally while local on-screen input is affected.

This has reproduced more than once. Reducing the transcript size and enabling character-based input did not prevent recurrence.

### Steps to reproduce

The exact trigger is not yet deterministic:

1. Start Termux and use an interactive terminal session for an extended period, including a program that produces incremental terminal output.
2. Continue using the on-screen keyboard.
3. Eventually, typed characters are severely delayed or appear not to be accepted.
4. Connect to the device's Termux SSH server and interact through SSH.
5. SSH input remains responsive while the local Termux UI remains affected.

### Expected behavior

Local keyboard input should remain responsive for long-running terminal sessions.

### Observed diagnostics during the faulty state

- Termux remained the focused foreground activity.
- Android's served view and served input connection still pointed to `com.termux.view.TerminalView`; the input connection was not marked deactivated.
- `dumpsys gfxinfo com.termux` reported:
  - `Number High input latency: 6341`
  - `Number Slow UI thread: 136`
  - `Janky frames: 144 / 3917 (3.68%)`
- A stack snapshot caught the Termux main thread in `android.os.MessageQueue.nextLegacy`, waiting while acquiring an object monitor.
- `TermSessionInput` itself showed negligible CPU usage.
- The app was not under memory pressure and there was no ANR.
- SSH input continued to work normally throughout, suggesting the shell, PTY, and backend process were healthy and the issue was isolated to the local Android UI/input path.

### Configuration already tested

```properties
terminal-transcript-rows = 1000
enforce-char-based-input = true
```

Both settings were loaded before the issue reproduced.

### System information

- Termux application version: `0.118.3`
- APK release: official GitHub build
- Android version: `16`
- Device: OnePlus PJA110 (ColorOS)
- Architecture: `aarch64`
- Kernel: `5.15.180-android13`
- Termux target SDK reported by the system: `28`

[...]

PR #5230, opened 2026-08-03
Coalesce redundant terminal input messages

zhaodong1982:
## Summary

The PTY reader posts one `MSG_NEW_INPUT` message per chunk it writes to `mProcessToTerminalIOQueue`, but the main-thread handler drains the whole queue on the first message it handles. Under sustained output the surplus messages accumulate in the main Looper queue, and key events — which are enqueued behind them — are delayed.

This coalesces the wakeups so at most one `MSG_NEW_INPUT` is outstanding at a time.

Fixes the input latency / ANR reported in #5229.

## Root cause

The reader reads at most 4 KiB per `read()` and posts a wakeup after each write. The handler's `mReceiveBuffer` is 64 KiB — the same size as `mProcessToTerminalIOQueue` — so a single `handleMessage` call empties the queue completely.

Draining 64 KiB therefore consumes one message but retires roughly sixteen. The remaining fifteen run with an empty queue and do nothing. Because the producer keeps posting at the same rate, the deficit is not bounded: the main thread's message queue grows for as long as output continues, and every input event queued behind that backlog waits for all of it to dispatch.

## Fix

An `AtomicBoolean` records whether a wakeup is already outstanding. The reader posts only when it wins the compare-and-set; the handler clears the flag *before* draining the queue.

Clearing before the read is what makes this safe. If the reader's CAS fails, the flag was still `true`, which means the already-posted message had not yet reached `handleMessage` — so its `read()` is guaranteed to observe the bytes just written. If the CAS succeeds after a concurrent drain, the result is one harmless empty wakeup, which `bytesRead > 0` filters out. No wakeup can be lost.

This also relies on `mReceiveBuffer.length` matching the `mProcessToTerminalIOQueue` capacity, so one read always empties the queue. That invariant is noted in the code comment.

Compared with a `Handler.hasMessages()` / `removeMessages()` approach, this is O(1) and does not scan or mutate the main thread's `MessageQueue`.

### Automated

- `./gradlew :terminal-emulator:test` — 290 tests pass, Debug and Release. `BUILD SUCCESSFUL`.
- Debug and Release Java compilation of the modified module pass.
- `git diff --check` clean.

### Device A/B on this PR's head

Two debug APKs were built from an identical tree; the **only** difference between them is this PR's diff (`TerminalSession.java` reverted to its `3df69d1d` content for the control). Same signing key, same build flags, same native libraries.

| | classes2.dex contains `mNewInputMessagePending` |
| --- | --- |
| control (`3df69d1d`) | no |
| patched (`e2bd0ca2`) | yes |

Both were run on the same device — OnePlus PJA110, Android 16 — under the same 60-second workload:

```sh
timeout 60 seq -f 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-%09.0f' 1 999999999
```

The advancing numeric suffix makes a rendering stall distinguishable from repeated-looking output. Measurement was scripted rather than judged by eye: the screen was captured every ~3.5 s and hashed, the producer's liveness was polled from a root shell, and frame timings were read from `dumpsys gfxinfo` (reset immediately before each run). A stall is a run of identical screen hashes while `seq` is still alive.

| Metric | control `3df69d1d` | patched `e2bd0ca2` |
| --- | ---: | ---: |
| Total frames rendered | 33 | 3179 |
| 50th percentile frame | 19 ms | 12 ms |
| 90th percentile | 30 ms | 17 ms |
| 95th percentile | 3150 ms | 19 ms |
| 99th percentile | 4950 ms | 22 ms |
| Longest frozen stretch while producer alive | 24 s | 0 s |
| Distinct screens across 40 samples | 8 | 23 |

In both runs `seq` ran for the full 60 seconds (start/end marks exactly 60 s apart), so the producer was healthy in both cases and only the UI froze — which is the symptom described in #5229. The control's 99th-percentile frame time of 4950 ms is close to the 5530 ms frame reported in this PR's comments.

One caveat, stated plainly: the janky-frame *percentage* barely moves (45.45% → 37.53%), because the control rendered only 33 frames in total and the denominator collapses. The meaningful numbers are the absolute frame count and the tail latency. All of the patched run's janky frames are within 22 ms — slightly over the 16.7 ms budget, none of them stalls.

Earlier device testing of the superseded `Handler.hasMessages()` / `removeMessages()` implementation showed the same pattern on a Nubia P0110 and a Pixel 10 Pro XL.

## Notes for reviewers

Single file, +11/−1, no API or behaviour change other than the number of main-thread wakeups. Output is still appended in queue order and `notifyScreenUpdate()` still fires whenever bytes were consumed — only the dispatch granularity changes, from per-4-KiB-chunk to per-drain.

## Trying the patch

Anyone hitting #5229 can build this branch directly:

```sh
git clone https://github.com/zhaodong1982/termux-app
cd termux-app
git checkout agent/coalesce-terminal-input-messages
./gradlew assembleDebug
# app/build/outputs/apk/debug/termux-app_apt-android-7-debug_arm64-v8a.apk
```

Installing the debug APK requires uninstalling the existing Termux first unless your current build already uses the repo's `testkey_untrusted.jks`; uninstalling removes `$PREFIX`, so back up anything you need.

Related: #5229


turgs:
I'm so glad I found this PR, I was about to open my own. This bug currently makes Termux unusable for my normal workflow. 

I’m seeing what looks like the same issue on a Pixel 6a (bluejay) running Android 17. My normal workflow is Termux > SSH to a remote machine > remote tmux with multiple windows. Termux eventually becomes completely unresponsive and Android shows the ANR dialog.

Logcat immediately before the ANR shows 330 skipped frames and a 5530 ms frame, followed by 150 skipped frames / 2538 ms and 227 skipped frames / 3826 ms. There’s no Java FATAL EXCEPTION, so this seems consistent with the main-thread message backlog described here.

@zhaodong1982 could you move this PR out of draft and mark it ready for review? That should make it more visible to repo admins. Then GitHub Actions runs will kick off tests and checks automatically.

PR #5231, opened 2026-08-03
Add Chinese translations (zh-rCN + zh-rTW)

1366214813:
Add complete Simplified Chinese (zh-rCN) and Traditional Chinese (zh-rTW) translations.

## Changes
- 6 new strings.xml files across 3 modules
- 163 translated strings total
- Coverage: 100% of user-facing UI

## Translations cover
- Session management (new, rename, kill, max terminals)
- Menu actions (share, URL select, autofill, style, settings)
- All 4 settings categories (debugging, terminal IO, terminal view, API/float/tasker/widget plugins)
- Bootstrap installer and error dialogs
- Notifications and file receive
- About and Donate preferences


twaik:
We are not planning to add any translations.

TomJo2000:
To expand on that.
Android's Bionic libc does not provide the same kind of locale mechanism available in Glibc.
So packages compiled for it can't provide localization support in the same way they would on a Glibc system.
Thus there is no way for us to provide translations for packages available on Termux, thus adding translations to the Termux app itself has little tangible benefit as it would just create a disconnect between the app interface and command output.

ishu89532-netizen:
Don't disturb me

twaik:
Unsubscribe from the repo...

TomJo2000:
> Don't disturb me

You weren't even part of this PR discussion.
Check your notification settings,
you might have selected "Watch all activity" on the repo front page.

Issue #5068, opened 2026-04-11
[Feature]: Support the Kitty image protocol to show images


stevenxxiu:
### Feature description

Support the *Kitty* image protocol.

#2973 implements the *Sixel* image protocol, but this is a rather legacy protocol. Many programs now only support the *Kitty* image protocol to show images. Terminal emulators too, like *Ghostty*.

### Additional information

Per <https://sw.kovidgoyal.net/kitty/graphics-protocol/>, terminal emulators with support include:

- [Ghostty](https://ghostty.org)
- [Konsole](https://invent.kde.org/utilities/konsole/-/merge_requests/594)
- [st (with a patch)](https://st.suckless.org/patches/kitty-graphics-protocol)
- [Warp](https://docs.warp.dev/getting-started/changelog#id-2025.03.26-v0.2025.03.26.08.10)
- [wayst](https://github.com/91861/wayst)
- [WezTerm](https://github.com/wez/wezterm/issues/986)
- [iTerm2](https://github.com/gnachman/iTerm2/commit/4fe5b2173193b6c3e45234b6b2ab7a144a5cfa01)
- [xterm.js](https://github.com/xtermjs/xterm.js/discussions/5683)

robertkirkman:
I agree that since iTerm2, Kitty and Konsole all support the Kitty image protocol, having Kitty image protocol in Termux would be ok, but **if any any image support is merged, I believe that Sixel image protocol should be merged first, with Kitty image protocol as a possibility for the future, and not something that should be merged first.** I will explain my reasoning for holding that opinion.

Kitty is a high performance image protocol optimized for modern terminals. It is performant enough that it supports animated video playback and even real time framebuffer rendering in the terminal at a high framerate, and a nice demo of that is this project, glkitty https://github.com/larkmjc/glkitty which is a port of the glxgears program to the kitty image protocol. 

Kitty has an opinionated lead developer who makes sure to focus on making Kitty not just a good terminal, but a leading terminal that pushes the terminal ecosystem into the future. For that reason, 

- https://github.com/kovidgoyal/kitty/issues/2511

**the lead developer has explicitly chosen to implement only the Kitty image protocol and not Sixel in Kitty specifically because the Kitty image protocol is more performant and better-architected than Sixel. however, despite Kitty being my favorite and most used X11 terminal emulator (including on Termux within X11), I do not believe that Termux should follow Kitty's example of abandoning Sixel and prioritizing the Kitty protocol**, and I will explain why.

[...]

PR #5238, opened 2026-08-05
Implement Kitty graphics protocol as a follow-up to #2973, not a replacement for Sixel

### LINKED ISSUE #2: Add special keys with volume key shortcuts
Here's the list that uses "Terminal emulator" https://github.com/jackpal/Android-Terminal-Emulator
I find those shortcuts super useful, would be nice to implement them to Termux.
photo_2015-10-26_16-51-27
Bajar = Down
Subir = Up
Espacio = Space

Keep up the good work! =D

### LINKED ISSUE #3: [FR] option to read password with termux-api?
I'm new to termux (very nice package and Android interaction concept!) and tried to use termux-api to query a password from a script, but termux-dialog seems to be rather limited so far.

Do you think it makes sense and is possible to add such an option, for example to perform encfs or sshfs mounts controlled by scripts?

kotamat:
## Summary

This PR implements the [Kitty graphics protocol](https://sw.kovidgoyal.net/kitty/graphics-protocol/) (issue #5068) by wiring up the APC (`ESC _ G ... ESC \`) parsing path that currently exists but is inert on `master` and on #2973 — `doApc()` is a no-op and the call from `receiveApcEsc()` is commented out ("Eat APC sequences silently for now").

This is **not a standalone feature**. It is built entirely on top of the image infrastructure introduced in #2973 ("Add graphics in terminal support: - Sixel and iTerm2 protocols", by @MatanZ and @agnostic-apollo). This PR does not modify #2973's Sixel or iTerm2 code paths; it adds a second front-end that shares the same backend (image registry, bitmap storage, rendering). It only makes sense as a PR to be applied after #2973, and I've written it and tested it that way (rebased on top of #2973's branch).

## Relationship to #2973 and to the maintainer's stated direction

In #5068, @robertkirkman noted that Sixel should be prioritized for merge, and that Kitty support is a possible future addition rather than something that should land first, since Termux isn't in the position kitty is in and shouldn't drop Sixel. I agree with that ordering and I'm not asking for this to be merged ahead of, or instead of, #2973.

I'm opening this now, ahead of #2973 actually merging, mainly so the work is visible and reviewable in parallel, and so it doesn't get lost. I'd expect this to sit until #2973 lands, and I'm fine with this being revisited, rebased, or reconsidered at that point — including the possibility that the maintainers decide not to take it. If it's more useful for me to close this and resubmit once #2973 merges, I'm happy to do that instead; just let me know your preference.

## What's implemented

| Feature | Status |
|---|---|
| Action `a=t` (transmit) | Supported |
| Action `a=T` (transmit + display) | Supported |
| Action `a=p` (put / placement) | Supported |
| Action `a=q` (capability query) | Supported |
| Action `a=d` (delete: all, `d=a`/`d=A`, `d=i`/`d=I`) | Supported |
| Transmission medium `t=d` (direct, base64) | Supported |
| Format `f=100` (PNG) | Supported |
| Format `f=24` (RGB raw) | Supported |
| Format `f=32` (RGBA raw) | Supported |
| Chunked transmission `m=1`/`m=0` | Supported (payload decoded incrementally per chunk, not buffered) |
| Image id `i=` (full 32-bit unsigned range) | Supported |
| Placement id `p=` | Supported, keyed by `(image_id, placement_id)` |
| Cell sizing `c=`/`r=` | Supported |
| Pixel sizing `s=`/`v=` | Supported |
| Source rectangle `x=`/`y=`/`w=`/`h=` | Supported (crop, then scale) |
| Response control `q=` (0/1/2) | Supported |
| Success/error responses (`OK`, `<CODE>:<msg>`) | Supported |
| `C=1` (don't move cursor after placement) | Supported |
| Transmission media `t=f`/`t=t`/`t=s` (file, temp file, shared memory) | Not supported — returns `ENOTSUP` |
| Compression `o=z` (zlib) | Not supported — returns `ENOTSUP` |
| Animation actions `a=f`/`a=a`/`a=c` | Not supported — returns `ENOTSUP` |
| Unicode placeholder placement (`U=1`) | **Not supported.** The key is parsed but ignored — no placeholder-based placement is implemented |
| `z=` (z-index) | Parsed, ignored |
| `X=`/`Y=` (sub-cell pixel offset) | Parsed, ignored — can cause up to a 1-cell visual offset compared to a fully compliant renderer |

On `U=1`: some clients (e.g. yazi) can use the Unicode placeholder path, but since Termux advertises Sixel support in DA1, yazi selects Sixel instead of Kitty's placeholder mechanism in that case, so this gap does not affect yazi in practice. It would matter for a client that requires Unicode placeholders specifically.

## Reuse of #2973's infrastructure

This PR does not introduce a parallel image system. It reuses, unmodified in behavior:

- The 64-bit packing of image coordinates into `TextStyle`
- `TerminalBuffer`'s image registry and scroll-driven garbage collection
- `TerminalBitmap`'s cell-splitting and size-limiting logic
- `TerminalRenderer`'s existing bitmap draw branch

The Sixel and iTerm2 code paths (`build(byte[]...)`, `build(TerminalSixel...)`, `addTerminalBitmapForImage()`) are structurally unchanged by this PR — Kitty parsing produces the same intermediate bitmap representation that Sixel/iTerm2 already produce, and goes through the same storage/rendering code from there.

## Tests

- Baseline: 145 existing terminal-emulator tests, all passing on both `master` and on #2973's head, unmodified.
- Added: 74 new tests for the Kitty parsing/action/response path, for a total of 219, with zero regressions in the existing baseline. `./gradlew :terminal-emulator:test` reports 219 tests / 0 failures / 0 errors for both `testDebugUnitTest` and `testReleaseUnitTest`.
- #2973 adds roughly 2,500 lines to the image pipeline (registry, bitmap handling, Sixel/iTerm2 parsing) without adding terminal-emulator tests for that pipeline. The tests added here exercise the shared image infrastructure (registry, bitmap storage, GC) as a side effect of testing Kitty, so this is incidentally the first test coverage that path has had.
- Error-path coverage includes: unterminated sequences, malformed base64 (invalid characters, `len % 4 == 1`, excess padding, data after padding, padding in a non-final chunk), unknown actions/keys, interruption mid-chunk-transfer, sequences left open with `m=1` and never closed, oversized dimensions, deletion of nonexistent ids, out-of-scope keys for a given action, very large `i=` values, and response suppression behavior for each `q=` level.
- Chunk-boundary base64 decoding is cross-checked against `java.util.Base64` as an independent reference implementation, across a range of chunk split points, to confirm the incremental decoder produces byte-identical output to decoding the same payload in one piece.
- Two end-to-end tests reproduce specific byte sequences captured from disassembling real client output, rather than only hand-constructed protocol messages.

## Verification on emulator

All of the following was verified by running the build on an Android emulator (API 35, AOSP system image, arm64-v8a, headless, software rendering) and inspecting the result — not just by reading the code. Verification consisted of pixel measurements on captured screenshots, `dumpsys meminfo` readings, and `adb logcat -b crash` checks, unless noted otherwise:

- End-to-end draw pipeline for raw RGB (`f=24`), raw RGBA (`f=32`), and PNG (`f=100`), at sizes from 1×1 up to 64×64.
- Repeated placement via `a=p`: one transmit followed by five separate put commands, all five renders confirmed present.
- Channel order is R,G,B[,A], non-premultiplied: confirmed by sending pure red, pure green, pure blue, and white, and confirmed that alpha = 128 produces the same on-screen result for both raw and PNG input.
- Source rectangle selects the correct quadrant: tested against a four-quadrant test image for all four quadrants, for both raw and PNG input, confirming the crop-before-scale ordering.
- `c=`/`r=` cell-based scaling is pixel-accurate against a 19×38px cell size: both dimensions specified matches exactly, one dimension specified preserves aspect ratio, neither specified renders at native size.
- `C=1` leaves the cursor position unchanged, confirmed via DSR cursor-position queries before and after placement.
- Deletion actually clears the affected pixels, with no leftover image data on screen afterward.
- 20 malformed/edge-case sequences produce no crash and no visible display corruption; subsequent text and SGR colors continue to render correctly afterward.
- A full-screen 1080×2400 RGBA frame (10.4MB payload, sent across 3375 chunks) renders without an out-of-memory error.
- **Sixel and iTerm2 output is pixel-identical to their behavior before this PR (no regression).** Sixel output was checked row-by-row against expected RGB values.
- The crash log buffer remained empty and the process PID stayed constant across the full verification run (roughly 30 minutes).

## Verification on physical device

The above was also verified on a physical device: Samsung SM-F966Q (Galaxy Z Fold7-generation), Android 16 / SDK 36, arm64-v8a, 10.85GB RAM. Measured terminal calibration: origin at (7,119), cell size 19×38px, 56 columns × 56 rows. A 1-cell image renders as exactly 19×38 = 722px of solid color with zero bleed into neighboring cells; this was reproduced twice.

Verified items (each backed by a pixel measurement or a DSR-based numeric check):

- **Sixel/iTerm2 regression check.** An iTerm2 4×2 quad image measures 76×76px, with each of the four quadrants covering exactly 1,369px. A sixel image measures exactly 120px (spanning two cells: 114 + 6), pixel-identical to its pre-existing output.
- **Text on the same row as an image (the effect of bug fix #2, below).** Verified across seven mixed-content cases, including: two images on one row (`L2_A` / image / `L2_B` / image / `L2_C`), three adjacent segments (text in columns 0–3, a 6-cell-wide bitmap in columns 4–9, text in columns 10–13), and a case where **styled text spans a renderer flush boundary** — `L7_RED_` rendered in (205,0,0) before an image, `L7_GRN` rendered in (0,205,0) after it — confirming that the flush correctly carries `lastRunStyle` across the boundary.
- **Alpha handling.** α=255 → (255,0,0); α=128 → exactly (128,0,0); α=0 → fully invisible (0 of 1,444px show any trace, with no bleed of the green background); α=128 on white → (128,128,128). Raw RGBA and PNG produce identical results in every case, which rules out premultiplied-alpha handling differences on the real GPU.
- **Formats.** 1×1 / 2×2 / 19×38 images sent as `f=24`, `f=32`, and `f=100`, with histograms matching exactly across formats (no channel-order drift).
- **Error paths.** The existing 20 malformed-input cases, plus 5 additional boundary cases, plus a single oversized command sent as one unchunked payload (262,144 characters). No crash, no display corruption, no corruption of subsequent text.
- **Memory** — see below.
- **Crash monitoring.** PID 23641 remained constant over a 27-minute run, the crash log buffer stayed empty, and no `OutOfMemoryError` or ANR occurred.
- **Reflow via fold/unfold.** `cmd device_state state 3` was used to force a column-count change (56×56 → 102×48), exercising the terminal's reflow path. No crash. **iTerm2, sixel, and Kitty images all produce identical reflow artifacts** — see the corresponding entry in Known Limitations below.

### Memory under sustained placement load

Measured under a streaming workload (one `a=t` transmit followed by 200 `a=p` re-placements of the same image), sampled 1,738 times:

| | Native heap PSS | Total PSS |
|---|---|---|
| Baseline | 15.5 MB | 90.2 MB |
| Peak (t=8s) | 90.7 MB | 170.8 MB |
| Shortly after (t=18–22s) | 18.9 MB | 104.3 MB |
| After 120s idle | 18.9 MB | 103.4 MB |

Memory returns to baseline+13MB within roughly 10 seconds and stays flat from there; the residual is consistent with the single displayed 1064×760×4 = 3.2MB bitmap. This was checked against a false-pass: the on-screen marker `=== HEAVY DONE 200 frames ===` confirmed the full run completed, the rendered image area measured 1064×760 with each of the four quadrants at exactly 200,870px, and there were no `KittyImage`/`TerminalBitmap` error log lines during the run.

Why this matters: `a=p` re-places an already-cached image with a command of roughly 30 bytes, so placements can happen at a much higher rate than Sixel/iTerm2 allow, since those protocols require re-sending the full image on every frame and are bandwidth-bound as a result. A naive implementation could accumulate orphaned bitmaps under that kind of load. This implementation performs targeted collection at placement time: it collects the `bitmapNum`s referenced by the cells being overwritten, scans only the rows with `mHasTerminalBitmap` set, and exits early once a reference to a given bitmap is found elsewhere. It does not call `recycle()` — references are simply dropped and left to the GC, since the renderer may hold a reference from a different thread. The existing throttled `doTerminalBitmapsGC` from #2973 is unmodified.

## Included bug fixes (independent of Kitty, cherry-pickable to #2973)

While building on #2973's image pipeline, I found and fixed three pre-existing issues unrelated to the Kitty protocol itself. Each is isolated into its own commit so it can be reviewed, applied, or cherry-picked to #2973 independently of the rest of this PR. All three commits apply cleanly against a pristine #2973 tree — I confirmed `git apply --check` succeeds for each one individually and for all three together, without any of the Kitty commits.

1. **Infinite loop in `TerminalBitmap.build()`.** When `BitmapFactory` fails to decode bounds for the input, `outWidth`/`outHeight` come back as 0, and the aspect-ratio-preserving `scaleFactor` loop condition `0 >= 0` never terminates. In testing, `scaleFactor` overflows and pins at 0 after 201 iterations, so the loop spins indefinitely rather than overflowing into a crash. On current #2973, this is reachable from `OSC 1337;File=inline=1;width=10:<garbage>`, i.e. a malformed iTerm2 image sequence can hang the JVM.

2. **Text preceding an image on the same row is dropped.** In `TerminalRenderer.java` (lines 106–121), the bitmap-drawing branch discards a pending text run instead of calling `drawTextRun` for it, and separately has an off-by-one in `lastRunStartColumn` (`column + 1`). This reproduces identically with #2973's existing iTerm2 (`OSC 1337`) images, so it isn't Kitty-specific — it's a pre-existing rendering issue in the shared renderer path that this PR's tests happened to surface.

3. **Cursor position when an image ends exactly in the last column.** The bounds check used `col < mColumns - 1`, which treats an image that ends exactly at the last column as "didn't fit," consuming one more row than the image actually covers and leaving the cursor at the start of the following row. `col < mColumns` is the correct condition.

   The evidence for this one comes specifically from the existing iTerm2 (`OSC 1337`) path, not from a Kitty-vs-Kitty comparison — on a pristine #2973 tree, Kitty's APC sequences are still parsed and discarded, so no Kitty image is ever displayed and there is nothing to compare against there. To isolate the bug from anything Kitty-related, I built a pristine #2973 tree in a separate git worktree and compared it directly against this PR's tree using an iTerm2 image, both built and run the same way:

   - Pristine #2973, a 2×1 iTerm2 image placed in columns 54–55 (1-based) of a 56-column screen: cursor-position DSR reports `[7;1` (wrapped to the next row, column 1).
   - This PR (same image, same placement): DSR reports `[6;56` (the image's last row, column 56 — the next empty cell after the image, so subsequent characters do not overlap it).

   Because this changes existing Sixel/iTerm2 cursor behavior, not just Kitty, it's kept as its own commit so it can be evaluated, merged, or dropped independently of the rest of this PR. If upstream considers the current behavior intentional, this commit can be dropped without affecting the Kitty support — the only consequence would be that a Kitty image ending exactly in the last column would keep consuming one extra row, same as iTerm2/Sixel currently do. Three unit tests are included (image ending exactly in the last column, image that genuinely doesn't fit, and the normal case).

## Known limitations / honest caveats

- For PNG payloads, the client-declared `s=`/`v=` values are used as the primary input to source-rectangle validation. The decoded bitmap dimensions remain the ground truth for rendering, so this is safe in the sense that it can't produce out-of-bounds reads, but a client that declares dimensions inconsistent with the actual PNG could get a legitimate crop request rejected with `EINVAL`.
- `d=a`/`d=A` deletes all Kitty placements regardless of current on-screen visibility; the spec describes this action as scoped to placements "visible on screen."
- `TextStyle` encodes `bitmapNum` in 16 bits while `TERMINAL_BITMAP__NUM_END` is defined as `Integer.MAX_VALUE`. This mismatch predates this PR (introduced in #2973) and is not touched here.
- `TerminalBuffer.resize()` does not remap bitmap-bearing cells. This is existing behavior from #2973, not changed by this PR.
- **Multi-row images split apart on reflow.** When the column count changes (e.g. folding/unfolding a foldable device), an image spanning multiple rows can end up separated into two fragments with a gap row between them, because the wrap-continuation rows move independently of the image. I verified on-device that this happens identically for iTerm2, Sixel, and Kitty images — it's pre-existing behavior from #2973's resize/reflow handling, not something this PR introduces or changes. Noting it here explicitly so it isn't mistaken for a regression caused by this PR.
- **Minor bilinear seam at scaled quadrant boundaries.** When scaling, there's a roughly 0.6% strip of blended pixels at internal quadrant boundaries. This showed up on the physical device's GPU-accelerated rendering path but not on the emulator's software renderer; it's expected behavior of hardware bilinear filtering, not a bug in the scaling logic.

## Emulator vs. physical device: notable differences

- **The transient memory peak is roughly 4x larger on the physical device**: 73.6MB above baseline on the emulator vs. 170.8MB above baseline on the device (baseline+80.6MB is roughly consistent with 25 still-live 3.2MB bitmaps awaiting collection). The device's GC appears to reclaim memory later than the emulator's. This is harmless at 10.85GB of RAM, but worth knowing about for lower-RAM or larger-screen devices, since the transient peak could matter there even though the steady state is fine.
- **Fold/unfold-driven reflow could only be exercised on the physical device** — the emulator has no equivalent state transition to reproduce it.

## What I'd like reviewers to look at

- Whether the deferral posture above (stack on #2973, don't merge ahead of it) matches what maintainers want, or whether this should instead be closed until #2973 lands and reopened later.
- The APC parsing and base64/chunking logic in isolation, since that's the only genuinely new parsing surface this PR adds.
- Whether the three included bug fixes should be split out as their own PR(s) against #2973's branch instead of riding along here — in particular bug fix #3, since it changes existing Sixel/iTerm2 cursor behavior and reviewers may want to weigh in on whether the current behavior was intentional.
- The `U=1` / Unicode-placeholder gap — whether that's an acceptable scope cut for a first pass, or a blocker.
- Error code/message choices for the `ENOTSUP` paths (file/shared-memory transmission, zlib, animation), in case there's a preferred convention already used elsewhere in the codebase.

---

**Note on where to review this:** the same four commits are also open against #2973's own branch as [MatanZ/termux-app#2](https://github.com/MatanZ/termux-app/pull/2), where the diff is limited to this PR's own changes (9 files) instead of also including #2973's 15 commits. Either one is fine to review — whichever is more convenient. I'll keep them in sync.


robertkirkman:
> Whether the deferral posture above (stack on https://github.com/termux/termux-app/pull/2973, don't merge ahead of it) matches what maintainers want, or whether this should instead be closed until https://github.com/termux/termux-app/pull/2973 lands and reopened later.

Thank you, the bug fixes for the PR #2973 that you have made might be helpful to apply on top of it before it is merged. It is fine to leave this PR open so that people are aware of it and can review it also.

alps2006:
The PR #2973 have been merged, I expect to update the work based on it.

PR #5248, opened 2026-08-08
Fixed: Fix locale sensitive case conversion, byte truncation, selection offset, main thread transcript building and add phantom processes docs

BuildWithAni:
Fixes several bugs found while reviewing the codebase, across `app`, `terminal-view` and `termux-shared` modules.

1. **Locale-sensitive `toLowerCase()`/`toUpperCase()`** (8 files, `termux-shared` + `app`). Calls without an explicit locale used the device locale, so on Turkish (tr-TR) devices `"CTRL+I"` lowercased to `"ctrı+ı"` and stopped matching `"ctrl"`, breaking session shortcut parsing, property map lookups, filename sanitizing and MIME type detection. All calls now use `Locale.ROOT`.

2. **`getTruncatedCommandOutput()` truncates by characters, not bytes** (`DataUtils.java`). The limit constants (`TRANSACTION_SIZE_LIMIT_IN_BYTES` = 100KB, `LOGGER_ENTRY_MAX_SAFE_PAYLOAD`, etc.) are byte limits, but truncation counted characters. With multibyte UTF-8 output (CJK, emoji, up to 4 bytes/char) the result could still exceed the binder transaction limit by 4x and cause `TransactionTooLargeException` when sending command results via pending intents. Truncation is now byte-based, never splitting a multi-byte character.

3. **Hardcoded `40` px offset in `TerminalView.getCursorY()`** (`terminal-view`). The `40` was a leftover finger-offset from the old `toggleSelectingText()` code, but `getCursorY()` is now used unconditionally by `TextSelectionCursorController.updatePosition()` when dragging selection handles — and it ignored the font size entirely, so handles tracked the wrong row at any non-default font size. It now uses `mRenderer.mFontLineSpacingAndAscent`, matching `getColumnAndRow()`.

4. **Transcript building on the main thread** (`TermuxTerminalViewClient.java`). `shareSessionTranscript()`, `showUrlSelection()` and `reportIssueFromTranscript()` built the full terminal scrollback string (and ran URL extraction regexes) on the UI thread — an ANR risk with large scrollback. Transcript building now runs on a background thread with the UI work posted back via `runOnUiThread()`.

5. **Phantom processes docs page** (`docs/` + `README.md`). The README notice promised "A proper docs page will be added later". Added `docs/en/apps/termux/phantom-processes.md` covering how to detect kills via logcat, how to check the `max_phantom_processes` limit, and how to disable phantom/excessive-CPU process trimming per Android version (12, 12L/13, 14+), and linked it from the README notice.

All modules compile successfully:

```
BUILD SUCCESSFUL
```

Unit tests for `terminal-emulator` and `app` pass:

```
BUILD SUCCESSFUL
```


robertkirkman:
Thank you for the contributions,

many of these problems may be fixed in an upcoming version of Termux and the code edited may be replaced with other code, so you should be prepared to wait and eventually rebase the PR when the update comes.

PR #5264, opened 2026-08-16
Offer Rename / Kill when long pressing a session in the drawer

kedimuzafer:
Long pressing a session in the drawer goes straight to renaming it, so there is no way to close a
session from the drawer at all. Killing one means switching to it first and using the terminal's own
context menu, or typing `exit`.

This makes long press open a two item dialog offering **Rename** and **Kill**. Rename keeps the
existing behaviour; Kill reuses the same confirmation prompt (`title_confirm_kill_session`) before
calling `finishIfRunning()` and removing the finished session, matching what the terminal context
menu already does.

Two strings are added for the menu items and one for the confirmation.

Tested on an Android 16 device and an API 31 emulator: renaming still works, killing removes the
session from the drawer list, and cancelling leaves it running.

### DIFF
--- a/app/src/main/java/com/termux/app/terminal/TermuxSessionsListViewController.java
+++ b/app/src/main/java/com/termux/app/terminal/TermuxSessionsListViewController.java
@@ -1,6 +1,7 @@
 package com.termux.app.terminal;
 
 import android.annotation.SuppressLint;
+import android.app.AlertDialog;
 import android.graphics.Color;
 import android.graphics.Paint;
 import android.graphics.Typeface;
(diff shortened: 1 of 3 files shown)


kedimuzafer:
Pushed a fix for two problems in the original commit, found while using it.

**Killing a running session did not remove it.** `finishIfRunning()` only sends SIGKILL; the process exits asynchronously, and `TermuxSession.finish()` returns early for a session that is still running, so calling `removeFinishedSession()` in the same breath silently did nothing. An already exited session was removed fine, which hid the bug.

The removal is now deferred to `onSessionFinished()` through a new `killAndRemoveSession()`. Sessions that exit on their own take the existing path unchanged, and the auto-close rule for exit codes 0 and 130 is untouched.

**The confirmation dialog was unreadable.** It was built with `androidx.appcompat.app.AlertDialog`, which picks up the activity theme; `Theme.TermuxActivity.DayNight.NoActionBar` sets a black window background, so the buttons ended up dark on dark. Switched to `android.app.AlertDialog`, which is what `showKillSessionDialog()` and `TextInputDialogUtils` already use.

PR #5265, opened 2026-08-16
Show the toolbar text input and the extra keys at the same time

kedimuzafer:
The terminal toolbar is a two page `ViewPager`: page 0 holds the extra keys, page 1 the text input.
Switching pages means swiping horizontally — but the text input consumes horizontal touches to move
the caret, so the pager only sees the swipe once the caret has reached the start of the text. With a
long line in the field, getting back to the extra keys means first scrolling the caret all the way
back. Paste a paragraph in there and it is a real chore.

This stacks both views in a vertical `LinearLayout` so they are visible together, with the text input
directly above the extra keys.

toolbar

**`TEXTBAR` special key.** Toggles the input row and moves focus with it, alongside the existing
`KEYBOARD`, `DRAWER`, `PASTE` and `SCROLL` keys, so its position is up to the user's `extra-keys`
property. It renders as `▤`.

**Text input.** Multi line and monospace, with hairline separators instead of the `EditText`
underline, which does not sit well in a terminal. Enter sends, so no send button competes for
horizontal space.

**Colour scheme.** The toolbar takes background and foreground from the terminal's own scheme. Key
labels were a fixed white before, so they were invisible against a light scheme — that is fixed as a
side effect.

**Label sizing.** Labels are sized from the row height, so lowering `terminal-toolbar-height` shrinks
the text with the row instead of leaving oversized labels crammed into a thin row.

**Key grid.** A hairline grid is drawn between keys in the same scheme derived colour.

### Notes for review

I am aware this removes an existing interaction, so I understand if you would rather have it behind a
property — say the word and I will put the old paged behaviour back as the default with this as an
opt-in, or split the parts up. The colour scheme and label sizing fixes stand on their own if you
only want those.

`isTerminalToolbarTextInputViewSelected()` and `isTerminalViewSelected()` keep their meaning: the
former now reports whether the input row is visible rather than which page is current.

Tested on an Android 16 device (arm64) and an API 31 emulator (x86_64): toggling, sending, focus
handling, light and dark schemes, and `terminal-toolbar-height` between 0.4 and 3.


fitchmultz:
Tested on a Samsung foldable (`SM-F976U1`, Android 17) with Termux `0.119.0-beta.3` and Samsung Keyboard. This PR addresses a real workflow problem for me.

I enabled `enforce-char-based-input = true` to fix Samsung's issue where composed/swiped words remain invisible until Space. That makes direct terminal input immediate, but Samsung disables normal automatic spacing in that mode. The toolbar Text Input View restores normal swipe typing and auto-spacing, so it is useful for longer AI-agent prompts.

I can reproduce the conflict described here: returning from the Text Input View to extra keys is unreliable while the `EditText` can consume the horizontal gesture. Sending/clearing the field so the caret is at the start makes the right-swipe more reliable, but it is still a finicky mode switch.

Showing both views together and exposing `TEXTBAR` would remove that gesture conflict. An opt-in property would be welcome because cover-screen vertical space is limited, while the unfolded display has plenty of room. Thanks for working on this.

PR #5273, opened 2026-08-20
Open Termux wiki in an external browser

pelya:
Because Anubis prevents wiki page from loading in regular WebView on Android 12

### DIFF
--- a/app/src/main/java/com/termux/app/activities/HelpActivity.java
+++ b/app/src/main/java/com/termux/app/activities/HelpActivity.java
@@ -63,6 +63,9 @@ public final class HelpActivity extends AppCompatActivity {
             }
         });
         mWebView.loadUrl(TermuxConstants.TERMUX_WIKI_URL);
+        // Anubis prevents Termux wiki from loading in regular WebView, we need to launch an external browser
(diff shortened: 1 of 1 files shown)


robertkirkman:
There is a plan for a Termux wiki offline static release to eventually be generated and stored inside the Termux app and then rendered using a lightweight embedded Markdown renderer

pelya:
That's fine, but having wiki in a separate window is still better, because
you can copy command examples to Termux window without closing the wiki.

app/src/main/java/com/termux/app/activities/HelpActivity.java
```diff
@@ -63,6 +63,9 @@ public void onPageFinished(WebView view, String url) {
             }
         });
         mWebView.loadUrl(TermuxConstants.TERMUX_WIKI_URL);
+        // Anubis prevents Termux wiki from loading in regular WebView, we need to launch an external browser
+        Intent browserIntent = new Intent(Intent.ACTION_VIEW, Uri.parse(TermuxConstants.TERMUX_WIKI_URL));
+        startActivity(browserIntent);
```
rbratul216-star: Bine info

PR #5275, opened 2026-08-24
Add Always Keep Alive preference resources and a small foreground ser…

vallind:
…vice to keep app alive (keepalive/always-on)


vallind:
commit

PR #5277, opened 2026-08-24
build: update actions/setup-java to v6

brunoborges:
## Summary

Update all active workflow references to `actions/setup-java` from v5 to v6.

## Validation

- Verified all active `.github/workflows` references now use `@v6`
- Ran `git diff --check`


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

Issue #5269, opened 2026-08-18
[Feature]: Support DECSET 1003 (any-event mouse tracking) for mouse hover


lcamus:
### Feature description

I would like Termux to support DECSET 1003 (xterm any-event mouse tracking), so terminal applications can receive mouse movement events even when no mouse button is pressed.

Termux already supports mouse interaction for clicks, scrolling, and movement while a mouse button is pressed. However, a TUI application cannot currently detect simple pointer movement/hover.

A practical use case is a mouse-enabled TUI where hovering over a control can highlight it or open a contextual menu, while clicking remains available as a fallback.

Ideally, applications explicitly enabling DECSET 1003 would receive mouse movement events without a button being pressed. Existing mouse behavior should remain unchanged for applications that do not request this mode.

### Additional information

I tested this with a physical mouse connected to an Android device running Termux.

A small prompt_toolkit test application with mouse support enabled receives mouse clicks and other supported mouse events, but receives no mouse movement event when the pointer is moved without pressing a button.

Android physical mice can generate hover motion events (such as ACTION_HOVER_MOVE), so one possible implementation may be to handle these events when DECSET 1003 is enabled and translate them into terminal mouse reports, using SGR 1006 encoding when requested by the terminal application.

This is only an implementation suggestion; the main request is support for standard xterm any-event mouse tracking (DECSET 1003).

PR #5281, opened 2026-08-26
Add: Support DECSET 1003 (any-event mouse tracking)

### LINKED ISSUE #5280: Support DECSET 1003 (any-event mouse tracking); currently ignored, breaks Claude Code fullscreen, Vim `balloonevalterm`, Zellij/Bubble Tea TUIs over mosh
## Summary

Termux's `TerminalEmulator` implements DECSET `1000` (press/release), `1002` (button-event) and `1006` (SGR encoding), but `1003` (any-event tracking) is silently ignored: it is not mapped in `mapDecSetBitToInternalBit()`, and `case 1003:` in `doDecSetOrReset()` falls into the no-op `break` group.

The practical consequence is not "hover is missing" — it is that **mouse tracking is off entirely** for any application that requests only `1003`. `isMouseTrackingActive()` returns `false`, so `TerminalView` never forwards touch events to the child process; swipes scroll Termux's own scrollback and taps do nothing in the TUI.

## Reproduction

Termux 0.118.x (F-Droid), Android, touch input only.

```sh
printf '\e[?1003h\e[?1006h'; cat -v
# swipe up/down, then tap
```

**Observed:** no output at all. The swipe scrolls Termux's scrollback instead of reaching `cat`.

**Expected** (and what happens today if you substitute `1002` for `1003`):

```sh
printf '\e[?1002h\e[?1006h'; cat -v
# swipe up   -> ^[[<64;12;7M
# swipe down -> ^[[<65;12;9M
# tap        -> ^[[<0;12;8M  ^[[<0;12;8m
```

Reset with `printf '\e[?1003l\e[?1006l'`.

## Why applications hit this

Several TUIs request `1003` rather than `1002` because they want motion-without-button (hover) on desktop terminals. `1003` is a strict superset of `1002` on every other terminal, so requesting it is the correct thing to do — the app has no reason to expect it to be *worse* than `1002`.

The failure is amplified by multiplexers and remote shells. Claude Code's fullscreen mode requests only `?1003h` (for hover highlighting). Through tmux and mosh this collapses to "1003 only" by the time it reaches Termux, so Termux ends up with no tracking at all and swipes fall back to arrow keys — which the TUI then interprets as cursor movement rather than scrolling.

## Cross-references

- vim/vim#7422 — `balloonevalterm` sends `1003`; taps stop working in Termux.
- mobile-shell/mosh#1364 — states Termux lacks `1003`.
- zellij-org/zellij#4543 — Zellij mouse issues on Termux.
- Gitlawb/zero#505 and NousResearch/hermes-agent#30084 — both carry "use 1002 when `$TERMUX_VERSION` is set" workarounds.

Each of these is a per-application workaround for the same one-line gap in the emulator. No existing Termux issue or PR covers `1003` (checked Aug 2026); termux-app#4083 is the only nearby mouse PR and is about a touch-drag → mouse-move *setting*, not about mode `1003`.

## Proposed fix

Treat `1003` as an alias of `1002`. Android touch input has no hover state, so there are no motion-without-button events to report; synthesizing them would be wrong. Making `1003` behave exactly like `1002` is both the correct semantics for a touch device and enough to unbreak every application above.

Concretely: add a `DECSET_BIT_MOUSE_TRACKING_ANY_EVENT` internal bit, map `1003` to it, include it in the mutual-exclusion group with `1000`/`1002`, make `isMouseTrackingActive()` account for it, and send `MOUSE_LEFT_BUTTON_MOVED` when either the button-event or the any-event bit is set. DECRQM, `CSI ? s` / `CSI ? r` save/restore and `reset()` all route through the existing generic paths and need no extra work.

Happy to open a PR — patch is written and under local testing on a device.

samuellvicente:
## Description

`TerminalEmulator` implements DECSET `1000`, `1002` and `1006`, but `1003` is silently ignored — it is not mapped in `mapDecSetBitToInternalBit()`, and `case 1003:` in `doDecSetOrReset()` falls into the no-op `break` group.

Because nothing sets a tracking bit, `isMouseTrackingActive()` returns `false` for an application that requests only `?1003h`. `TerminalView` gates *all* touch forwarding on that method, so the result is not "hover is missing" but "mouse tracking is off entirely": swipes scroll Termux's own scrollback and taps never reach the child process.

This change adds a `DECSET_BIT_MOUSE_TRACKING_ANY_EVENT` bit for `1003` and treats it as an alias of `1002`.

Fixes #5280.

## Rationale for aliasing 1003 to 1002

`1003` differs from `1002` only in reporting motion *without* a button held — hover. Android touch input has no hover state: there is no pointer position between an `ACTION_DOWN` and the matching `ACTION_UP`. There are therefore no additional events to report, and synthesizing them would produce motion reports no real pointer generated.

So `1003` is implemented with exactly `1002` semantics. That is the correct behaviour for a touch device, and it is sufficient to unbreak the applications listed below, all of which request `1003` because on desktop terminals it is a strict superset of `1002`.

Note this also fixes the Termux side of mobile-shell/mosh#1364: mosh (and tmux) pass `?1003h` through unchanged, so an application that requests `1003` over `mosh` + `tmux` currently arrives at Termux as "1003 only" and loses tracking altogether. Affected upstreams include vim/vim#7422 (`balloonevalterm`), zellij-org/zellij#4543, and Claude Code's fullscreen mode; Gitlawb/zero#505 and NousResearch/hermes-agent#30084 both carry "use 1002 on Termux" workarounds that this makes unnecessary.

## Changes

`terminal-emulator/src/main/java/com/termux/terminal/TerminalEmulator.java`:

- Add `DECSET_BIT_MOUSE_TRACKING_ANY_EVENT` (`1 << 13`, the next free bit).
- Map `1003` → that bit in `mapDecSetBitToInternalBit()`.
- Generalise the mouse-mode mutual exclusion in `setDecsetinternalBit()` from a pairwise `if`/`else if` to a mask over all three modes, so setting any one clears the other two.
- `sendMouseEvent()`: send `MOUSE_LEFT_BUTTON_MOVED` when either the button-event or the any-event bit is set.
- `isMouseTrackingActive()`: account for the new bit.

`terminal-emulator/src/test/java/com/termux/terminal/TerminalTest.java`:

- Add `testMouseAnyEventTracking()` alongside the existing `testMouseClick()`.

No changes outside the emulator layer. `TerminalView` already gates every touch path on `isMouseTrackingActive()`, so it picks this up unchanged.

Paths that needed no change, for reviewers:

- `case 1003:` was already present in the no-op `break` group of `doDecSetOrReset()`, and the internal bit is applied before that switch — so `h`/`l` work via the existing generic path.
- DECRQM and `CSI ? s` / `CSI ? r` (XTSAVE/XTRESTORE) both go through `mapDecSetBitToInternalBit()`.
- `reset()` clears `mCurrentDecSetFlags` wholesale.

## Test plan

Unit tests:

```sh
./gradlew :terminal-emulator:test
```

`testMouseAnyEventTracking()` covers: `?1003h` activates tracking; SGR click, drag (`MOUSE_LEFT_BUTTON_MOVED`) and wheel reports are emitted; switching to `?1000h` clears the any-event bit so drags stop being reported; `?1003l` deactivates tracking.

Manual, on a debug APK on a touch device:

```sh
printf '\e[?1003h\e[?1006h'; cat -v
# swipe up   -> ^[[<64;C;RM
# swipe down -> ^[[<65;C;RM
# tap        -> ^[[<0;C;RM  ^[[<0;C;Rm
printf '\e[?1003l\e[?1006l'
```

Verified on device (arm64, `apt-android-7` debug build): `?1003h` + `?1006h` produces the wheel and click reports above on swipe and tap, where stock 0.118.x produces nothing.

Regression checks, all run on the same build over `mosh` + `tmux` and all behaving as they did before the change:

- Claude Code fullscreen — the case this change is for. It requests only `?1003h`, which previously collapsed to no tracking at all through mosh and tmux; swipe now scrolls the conversation and taps land where expected.
- `btop` and `htop` — taps hit the right buttons and columns.
- `nvim` with `:set mouse=a` — tap positions the cursor, drag selects, swipe scrolls the buffer.
- `tmux` with `set -g mouse on` driving its own panes, not merely passing events through.
- Plain `1000`-only and `1002`-only applications, checked explicitly because the mutual-exclusion rewrite in `setDecsetinternalBit()` is the one place shared behaviour changed.

The `1000`-only case is worth spelling out, since it is the regression a reviewer should care about. Against `?1000h` + `?1006h`:

```
tap   -> ^[[<0;41;39M ^[[<0;41;39m      (press/release pair, correct)
drag  -> no ^[[<32;... reports at all   (correct: 1000 must not report motion)
swipe -> ^[[<64;... / ^[[<65;...        (wheel; Termux maps swipe to wheel
                                         buttons, and wheel is reported under
                                         every tracking mode - unchanged by
                                         this patch, same as on master)
```

So neither the button-event nor the any-event bit leaks into `1000`.


lcamus:
Thanks for working on DECSET 1003. This should fix an important issue for applications that currently lose mouse tracking entirely when requesting 1003.
I just wanted to point out a related but distinct use case from #5269: Termux running on Android with a physical mouse (Bluetooth/USB), rather than touch input.
Android can provide hover motion from physical pointing devices (ACTION_HOVER_MOVE). In that case, implementing 1003 strictly as an alias of 1002 would still leave true no-button mouse motion unavailable to terminal applications.
So #5281 looks useful independently, but I don't think it completely addresses #5269. It would be great if physical-mouse hover forwarding could remain a separate follow-up, potentially in TerminalView.
Thanks!

jakekausler:
Just installed a build from this and it successfully solved my scroll and click issues I was having with Claude Code. I built on commit b6a2615 with `gradlew assembleDebug`, installed the arm64 APK on my device, and it resolves the issue completely.

I came across this PR while trying to solve the problem on my end. The following can reproduce the issue
```shell
# Broken on master: dragging emits arrow keys
printf '\033[?1000h\033[?1002h\033[?1003h\033[?1006h'; cat -v
# => ^[[A ^[[A ^[[B ^[[B ...

# Works on master: dragging emits SGR mouse events
printf '\033[?1000h\033[?1002h\033[?1006h'; cat -v
# => ^[[<64;77;37M ^[[<65;77;37M ...
```

Two things that may not be captured yet:
1. There's a second-order symptom that the patch should also be checked against. On master, sending `?1003l` after `?1000h ?1002h ?1003h` turns tracking fully off rather than falling back to the still-set 1002. So the state is wrong in both directions, not just on set. Worth a test case if the mutual-exclusion rework doesn't already cover it.
2. It breaks more than scrolling. Under tmux with `mouse on`, the dead tracking state also kills status-bar window clicks. Mouse events never reach tmux at all. This gets fixed by this patch too.

My case was with Claude Code. On startup the fullscreen TUI emits `\033[?1000h\033[?1002h\033[?1003h\033[?1006h` at startup. On the unpatched Termux, dragging sent arrow keys rather than wheel events. The same session over Windows Terminal was fine.

NightMachinery:
It'd be great if this could be merged and released. Using Claude Code under mosh is very difficult with this bug.

NightMachinery:
I coded a workaround with Claude (following is Claude):

Independent confirmation of the failure mode, plus a userspace workaround for anyone who lands here before this ships.

The part that took me a while to see: **plain ssh + tmux does not reproduce it, mosh does.** When a pane asks for any-event mode, tmux emits the whole ladder — `CSI ?1006h`, then `CSI ?1000h` `CSI ?1002h` `CSI ?1003h` — and Termux keeps 1002 and drops 1003, so the mouse still works. mosh keeps a *single* mouse reporting mode, last one wins, and re-emits only that one, so what actually reaches Termux is `CSI ?1003h` on its own. Since touch forwarding is gated on mouse tracking being active at all, the result is no mouse rather than a coarser mouse, exactly as this PR describes. With `set -g mouse on` it is worse than it sounds: one pane in any-event mode raises the whole tmux client to that mode, so it takes the mouse away from every other pane in the window too.

Measured with Termux 0.118.1, mosh 1.4.0 on both ends, tmux 3.7b.

On the Claude Code claim: traced at the pty, its fullscreen mode sets 1000, 1002, 1003, 1006, in that order. So 1003 is its last word and the aliasing in this PR is exactly what it needs. Antigravity's CLI does the same thing, ending on 1003.

For anyone who cannot run a patched Termux yet, the same aliasing can be done one hop earlier, on the machine the application runs on: https://github.com/NightMachinery/decset-rewrite

```
decset-rewrite -termux -- <app>
```

It is a pty proxy that rewrites `CSI ? 1003 h/l` to 1002 in the application's output and leaves DECRQM replies honest. A stopgap, not an alternative — the fix belongs in the emulator, which is what this PR does. It also has a `-trace` flag that logs just the private mode numbers an application requests, which is how the Claude Code ordering above was measured, so it may be useful for triaging other reports of this.

PR #5287, opened 2026-08-29
Fix Samsung Keyboard right arrow suppression

### LINKED ISSUE #3944: Arrow keys not working on samsung keyboard[Bug]:
### Problem description

Hi
Im using a samsung tab s8 ultra with termux, when i try and use samsungs on screen keyboard the arrow keys do not work
Using standard samsung keyboard
Left arrow key does work, right key does not
Using Samsung keyboard with keys cafe - using this I can add up and down arrow keys as well, but again these do not work
Also the escape key minimises termux rather than the correct function

I have already tried setting enforce-char-based-input to True but this has not had any effect

Is there anything else I can do, or is there a way to fix this?

I have used other keyboards with termux and they work perfectly, hackers keyboard and SwiftKey (even the arrow keys work!) But I find samsungs keyboard much more comfortable and usable for me

Any help is appreciated and thank you for the amazing app!

Use a samsung device with a samsung keyboard and samsung keyboard with keys cafe installed and activated (from samsung galaxy store)

Arrow keys to work as expected

### System information

* Termux application version: 0.118.0
* Android OS version: 14
* Device model: SM-X900

Sparito181:
This pull request addresses the issue with Samsung Keyboard + Keys Cafe right-arrow cursor control.

Fixes #3944.

Samsung Honeyboard performs its own cursor-boundary checks before emitting DPAD navigation key events. Since TerminalView uses a BaseInputConnection without a conventional editable buffer, Honeyboard can consider the cursor to be at the right boundary and suppress KEYCODE_DPAD_RIGHT before Termux receives it.

This change exposes a Honeyboard-only virtual surrounding-text context with the cursor centered between private sentinels and keeps Honeyboard's selection and extracted-text state synchronized.

The virtual sentinel text exists only as IME metadata and is never written to the terminal or PTY. Other input methods retain the existing Termux InputConnection behavior.

Tested with Samsung Keyboard and Keys Cafe. LEFT and RIGHT arrow keys now both work normally in Termux.


arey9898:
**THIS IS A BUG TRACKER OF THE TERMUX APP. IF YOU HAVE ISSUES WITH A PACKAGE INSIDE THE APP, THEN PLEASE OPEN AN ISSUE AT [TERMUX/TERMUX-PACKAGES](https://github.com/termux/termux-packages) INSTEAD.**

PR #5295, opened 2026-09-03
Fixed(terminal): Pass unhandled Ctrl+Alt shortcuts to the terminal

### LINKED ISSUE #5240: [Bug]: C-M keybinds don't really work in Emacs
### Problem description

In emacs, shortcuts using C-M do not work. My guess is that all keybinds that start with Ctrl + Alt are instantly handled by Termux rather than the running program.
Solution: Pass Ctrl + Alt shortcuts to the running program if Termux cannot handle them
User workaround: Set disable-hardware-keyboard-shortcuts to false in termux.properties

Use any shortcut starting with C-M in Emacs (or use any Ctrl + Alt keybind in any other program)

These should be passed through to the running program if Termux cannot handle these keybinds.

### System information

* Termux application version: 0.118.3
* Android OS version: 14
* Device model: HONOR Pad 9

EchoEe247:
### Summary

Fixes #5240.

Termux currently consumes every hardware Ctrl+Alt key event whenever hardware shortcuts are enabled, even when the chord does not match any Termux shortcut. Returning `false` for only the unmatched branch lets `TerminalView` perform its existing Ctrl/Alt terminal encoding while preserving all recognized Termux shortcuts.

### Change

- Keep every existing recognized Ctrl+Alt shortcut on the existing consumed path.
- Return `false` only when no Termux shortcut matches, allowing the original `KeyEvent` to fall through to normal terminal handling.

### Validation

Validated against `termux/termux-app@3b66f8799635a4dba4a206563048ff0e6792c487`.

- `git apply --check`: PASS
- `git diff --check`: PASS
- Static routing review: PASS
- Upstream-equivalent Unit tests: PASS on Temurin Java 17 (`./gradlew test`, run 33729058553)
- Upstream-equivalent Build matrix `apt-android-7`: PASS (run 33729058528)
- Upstream-equivalent Build matrix `apt-android-5`: PASS (run 33729058528)
- Functional device test: not performed because safely replacing the production Termux installation/signature set was not appropriate

No production Termux installation or state was modified during validation.


belalalqaisi55-afk:
[]()

Issue #5309, opened 2026-09-09
[Bug]: Toolbar text input Enter (send) uses session.write() instead of emulator.paste() bypassing DECSET 2004


krecod:
### Problem description

I'm not sure whether the use of session.write() here is intentional or a bug.

When entering text through the toolbar text input and send ( Enter), the text is sent to the terminal using `session.write()`

The toolbar text input is intended as an alternative input method for Android's IME composed text. For this composed text, especially long text, it seems more appropriate to use paste semantics rather than treating the entire contents as raw keyboard input.

Using `session.write()` bypasses `TerminalEmulator.paste()`, and therefore does not honor Bracketed Paste Mode (DECSET 2004).

Other Termux paste entry points, like long press paste, route the text through `TerminalEmulator.paste()`

Toolbar text input  text is delivered as raw input without the bracketed paste sequences:

`\033[200~ ...text... \033[201~`

Applications that process input and redraw on each character can therefore receive and process the text "character by character". 

There is input delay due to streaming of letters, particularly in some reactive or component based TUIs.

Lighter TUI applications like lazygit may show the same behavior without an obvious performance impact but heavier ones like Antigravity CLI have noticable delay/wait time.

I don't see an obvious benefit (if any, then please let me know) to treating the entire contents of the toolbar text box as raw keyboard input instead of pasted text. The empty input case can still be handled separately so that pressing send with no text retains normal "Enter" semantics.

Proposed Fix: 
Use `session.getEmulator().paste()` for non empty toolbar text while retaining `session.write("\r")` for an empty submission:

In `app/src/main/java/com/termux/app/terminal/io/TerminalToolbarViewPager.java`

```
if (textToSend.length() == 0) {
     session.write("\r");
 } else {
     session.getEmulator().paste(textToSend);
}
```

### Steps to reproduce the behavior.

1. Run a terminal application that enables Bracketed Paste Mode (DECSET 2004).

2. Open the Termux  toolbar text input by swiping left on the bottom toolbar.

3. Enter a long text.

4. Send (Enter)

5. Compare the result with entering the same text using long press paste. 

### What is the expected behavior?

Non empty text submitted from the toolbar text input field should have the same semantics as other paste operations:

`session.getEmulator().paste(textToSend)`

When DECSET 2004 is enabled, the text should be delivered as a bracketed paste, rather than as individual keyboard input.

When bracketed paste mode is disabled, behavior  remain unchanged.

[...]

PR #5310, opened 2026-09-09
Fixed Bug #5309 - Use emulator.paste() for toolbar text input in TerminalToolbarViewPager.java

krecod:
This PR fixed bug #5309 and updates the toolbar text input field  so that submitting non empty text routes through `session.getEmulator().paste(textToSend)` instead of directly calling `session.write(textToSend)`

If `textToSend` is empty (length 0), we keep `session.write("\r")`, it preserves standard "Enter" keypress semantics without wrapping an empty send in bracketed paste escape codes.

Applications that enable Bracketed Paste Mode (DECSET 2004) receive the composed text as a single paste event `\033[200~` ...  `\033[201~` rather than raw individual keystrokes. 

This eliminates per character redraw cycles and input latency in some modern reactive or component based TUIs.


agnostic-apollo:
Responded in https://github.com/termux/termux-app/issues/5309#issuecomment-5641493673

PR #5312, opened 2026-09-10
Make terminal output auto-scrolling configurable

rebroad:
## Why this is needed

Some terminal applications, including Codex, continuously produce animated output. Under the previous behavior, every update forced the terminal viewport back to the bottom, making it effectively impossible to scroll upward and inspect earlier output. This was extremely frustrating in normal use and affects anyone using these applications in Termux.

## What changed

This PR preserves the user's scroll position when they have moved away from the bottom. When new output arrives, an optional down-arrow indicator appears and gives a single visible brightness flash. It remains available until the user clicks it or scrolls back to the bottom. Clicking it jumps to the latest output. The indicator fades while the user scrolls and returns when scrolling stops.

The indicator is controlled by Settings -> Termux -> Terminal View -> Show new output indicator and is enabled by default.

Terminal resizing, including keyboard dismissal and returning from Settings, preserves the current scroll position and clears stale output indicators when the bottom is visible.

## Validation

- `:terminal-view:test`
- `:app:assembleRelease`
- Manual verification on a Samsung Galaxy Z Flip7

The implementation is split into focused commits covering scroll preservation, indicator wiring, visual behavior, and resize handling.

### DIFF
--- a/app/src/main/java/com/termux/app/fragments/settings/termux/TerminalViewPreferencesFragment.java
+++ b/app/src/main/java/com/termux/app/fragments/settings/termux/TerminalViewPreferencesFragment.java
@@ -57,6 +57,9 @@ class TerminalViewPreferencesDataStore extends PreferenceDataStore {
             case "terminal_margin_adjustment":
                     mPreferences.setTerminalMarginAdjustment(value);
                 break;
+            case "terminal_new_output_indicator":
+                mPreferences.setTerminalNewOutputIndicator(value);
+                break;
             default:
                 break;
         }
@@ -69,6 +72,8 @@ class TerminalViewPreferencesDataStore extends PreferenceDataStore {
         switch (key) {
             case "terminal_margin_adjustment":
                 return mPreferences.isTerminalMarginAdjustmentEnabled();
+            case "terminal_new_output_indicator":
+                return mPreferences.isTerminalNewOutputIndicatorEnabled();
             default:
                 return false;
         }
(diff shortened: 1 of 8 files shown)


agnostic-apollo:
This is already solved by 5fc2b4cd4ae59a7cbfccea60c76019e96efb847c

krecod:
Does this also fix the issue where, if a user puts the Termux app in the background and then comes back or when user switch from one session to another, or when user open and close the IME keyboard , it starts scrolling from top to bottom(basically flickering)?

agnostic-apollo:
If you toggle the scroll extra key in `0.119.0-beta*` or master branch, auto scroll is disabled for all sessions until you toggle the key again.

agnostic-apollo:
@krecod Thanks for the report, not sure about flickering, but scroll row position should be restored after those changes with 74ab5126cb884199d499cc847a2e26f3e358de80. The scroll extra key currently doesn't go in active state (red color) like CTRL, etc, but I have fixed that with the updated extra keys implementation, will look into pushing that.

krecod:
Yeah, with the changes in 74ab512, it should fix the scroll row position issue (also flickering was wrong term, my bad). Thanks @agnostic-apollo for that!

I also looked at the 5fc2b4c commit from 2022 and was honestly surprised that it still hasn’t made it into stable version. It reflects how much work it must be to keep things moving alone.

I guess my PR for #5309 is going to be a bit of a long journey too!

I should probably switch from stable to beta.

agnostic-apollo:
@krecod Welcome. I am busy writing Android/JDK libraries and Termux rewrite for last few years, which indeed has been lot of work. Not sure about if/when 0.119.0 will be released as a stable release, but any Termux beta releases I make are stable code wise, they are just not stable for features, which may change, so feel free to switch. SCROLL key can probably be ported back to 0.118.0 stable releases though, will look into it.

PR #5322, opened 2026-09-18
fix: replace printStackTrace with Log in Logger.java

wized2:
Prefer Android Log over printStackTrace for production diagnostics.


krishnenduchakraborty:
Logs are useful, rather than stack-trace.

